PR #17, opened 2021-09-30
Introduce a check to simplify time expressions in annotations

nathankooij:
I wish to expand the test suite further, but it should already give a good indication of what is possible with this check.

A few things:
- I am not entirely certain if what I'm doing in `getTimeUnit` and `getDefaultTimeUnit` is always safe.
- We can extract some of the things to utilities which will a) make this code cleaner b) allow us to re-use it in other checks, some things:
      - Returning a map of attributes -> expression (or assignment trees if that's always the tree) by the matcher;
      - Extracting enum value from an expression;
      - Extracting default attributes on annotations;
      - An "annotation builder" to avoid special casing the implict `value` attribute.

I know @Stephan202 mentioned he has some of these on a branch already, so curious to see what we can already do out of the box. :)

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SimplifyTimeAnnotationCheck.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SimplifyTimeAnnotationCheck.java
@@ -23,6 +23,7 @@ import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.AnnotationTree;
 import com.sun.source.tree.AssignmentTree;
 import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.Tree;
 import com.sun.tools.javac.code.Scope;
 import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Symbol.VarSymbol;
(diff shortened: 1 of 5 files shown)


rickie (commented):
This is a really nice BugPattern 🚀

Stephan202 (commented):
Added a small commit and did a partial review. More later/another time :)

Stephan202 (commented):
Rebased, resolved conflicts and added a commit. (Won't do a full review now, but saw that I had the branch checked out, and was wondering about its state.)

rickie:
I encountered the `CanonicalDuration` [BugPattern](https://errorprone.info/bugpattern/CanonicalDuration) today. I made me think of this check, maybe there is some logic there that we can (re-)use?

Stephan202:
Yep, likely! Too bad it's [all `private`](https://github.com/google/error-prone/blob/9f8c332d9ea67dfd31d8a91a023392f61097ae8a/core/src/main/java/com/google/errorprone/bugpatterns/CanonicalDuration.java#L60), so that would entail some copying. Alternatively we could try to contribute improvements upstream, possibly by extracting shared logic to a separate class. (For this annotation case I'm not sure that a [banlist](https://github.com/google/error-prone/blob/9f8c332d9ea67dfd31d8a91a023392f61097ae8a/core/src/main/java/com/google/errorprone/bugpatterns/CanonicalDuration.java#L112-L118) applies.) Would then first open an issue to discuss that possibility, though.

> Thx for the feedback guys! I'll try to circle back to this PR somewhere this week. :)

I suppose that @nathankooij currently resides in the vicinity of a black hole ;)

PR #28, opened 2022-01-06
Improve naming of assorted Refaster templates and associated tests

rickie:
This PR contains code of #25 . We want to prepare the Refaster templates such that we can easily add the annotations `@Template{,Collection}` in the appropriate places. 
In the afternoon we talked about the reordering of some Refaster templates. I considered doing that in a second commit, but checked the files again, and saw that there is no alphabetical ordering currently, so I left it out.
Also, in the future we most likely want to automate the ordering with another BugPattern 😉, so did not bother to do it right now.

Note: for some Refaster templates our "current" naming scheme does not suffice.


Stephan202 (commented):
Rebased and added a commit. Seems to require a bit more thinking :thinking:

PR #46, opened 2022-03-31
Update comments referencing Refaster rule limitation

rickie:
While applying `error-prone-support` on repositories within Picnic we found that for some templates we lose generic type information. Therefore, we opened a PR upstream https://github.com/google/error-prone/pull/2706 to fix this. The problem is that for method invocations the type arguments were simply ignored by Refaster in the `@AfterTemplate`s. 

Our fork cherry-picked the commit from the upstream repository. In addition, we also use the fork for Picnic internally. 
However, currently we are not on the latest version due to [some issues](https://github.com/PicnicSupermarket/picnic-java-support-modules/pull/1718).

This should prepare the code for when we upgrade EPS and PSM to `v2.11.0-picnic-1` or `2.11.1` (which doesn't exist yet).


rickie:
I looked into this again. Rebased it and tried some things. However, I think we need to have a `HasTypeArgumentsMatcher` or something like that. Now it unconditionally introduces the type arguments, that is not the goal I would say. So we need to have two Refaster rules per case and use the `@{Not,}Matches` to make sure we don't introduce it everytime.

WDYT @Stephan202?

rickie:
Well, you are right (discussed [here](https://github.com/PicnicSupermarket/error-prone-support/pull/644#pullrequestreview-1444480015)) and indeed we cannot use a `Matcher` for this. In that case I'm not sure how we can best approach this. I think we need a `BugChecker` for this that only introduces the type arguments if we need it. If we would proceed with this PR we would introduce this while it's not necessary. Other option would be to create a check that detects we don't need the type arguments....

Stephan202:
> Other option would be to create a check that detects we don't need the type arguments....

That sounds more complex, but on the plus side would enable also other cleanup. I haven't looked into it deeply, but it would amount to identifying method invocations that (a) specify explicit type parameters and (b) invoke a specific method overload that, absent such type parameters, would also be selected by the algorithm defined by [JLS 18.5.2](https://docs.oracle.com/javase/specs/jls/se17/html/jls-18.html#jls-18.5.2). For (b) we'd of course rely on a few heuristics; just enough to avoid false positives while still covering the cases discussed here.

Stephan202 (approved):
I synced this branch and added two commits:
- One to drop `HasTypeArgumentsMatcher`, as we don't have uses for it.
- One to revert other changes, and to instead update comments that referenced google/error-prone#2706.

Rationale is that google/error-prone#2706 is no longer a blocker, and that instead we should make up our mind about whether to propagate the type parameters.

I suggest merging this PR with the following commit message:
```
Update comments referencing Refaster rule limitation (#46)

Issue google/error-prone#2706 has been resolved; we have yet to decide
how to proceed.
```

mohamedsamehsalah (approved):
✏️

rickie:
Thanks a lot @Stephan202 for picking up this old one. I think this is the right approach for now. It's an interesting case.

I encountered this case a lot when trying to implement the Guava integration tests.

I'd say lets merge this one :), updated the branch (might be offline in a bit when its ready to merge)

PR #142, opened 2022-06-22
Introduce `{Mono,Flux}Map{,NotNull}` Refaster rules

cernat-catalin:
Based on a suggestion from [here](https://github.com/PicnicSupermarket/picnic-platform/pull/8878#discussion_r851248541).

Adds a Refaster template to prefer `mono.map(x -> valueTransformation(x))` over the `mono.flatMap(x -> Mono.just(valueTransformation(x)))`

This avoids an unnecessary inner inner subscription.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -379,13 +379,21 @@ final class ReactorRules {
    * Prefer {@link Mono#map(Function)} over alternatives that unnecessarily require an inner
    * subscription.
    */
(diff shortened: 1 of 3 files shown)


Stephan202 (approved):
Rebased and added a commit to cover the `Mono.fromSupplier` case. We should certainly start looking into a "smarter" approach, but this is a good step forward, so let's go! :rocket:

Suggested commit message:
```
Introduce `{Mono,Flux}Map{,NotNull}` Refaster rules (#142)
```

cernat-catalin:
Thanks @Stephan202  for the changes!

rickie (approved):
Really curious to see how this will change things downstream. 

Thanks for another Reactor PR @cernat-catalin 🚀 !

PR #181, opened 2022-08-06
Migrate from JSR 305 to JSpecify

Stephan202:
:exclamation: This PR is currently on top of #182 :exclamation:

Suggested commit message:
```
Migrate from JSR 305 to JSpecify (#181)

JSpecify's annotations have more well-defined semantics. Its `@Nullable`
annotation is also a type-use annotation recognized by Google Java
Format, so the formatter places it after any field or method modifiers.

See https://jspecify.dev
```

~This is a draft PR. The build currently fails because `LexicographicalAnnotationListing` now incorrectly suggests swapping `@Nullable` and `@Override` annotations.~ #182 resolves this.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/package-info.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/package-info.java
@@ -1,4 +1,4 @@
 /** Picnic Error Prone Contrib checks. */
-@org.jspecify.nullness.NullMarked
 @com.google.errorprone.annotations.CheckReturnValue
+@org.jspecify.nullness.NullMarked
 package tech.picnic.errorprone.bugpatterns;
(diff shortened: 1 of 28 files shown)


Stephan202:

Stephan202:
Rebased and resolved conflicts; might be nice to get this PR moving :upside_down_face:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/package-info.java
```diff
@@ -1,4 +1,4 @@
 /** Picnic Error Prone Contrib checks. */
+@org.jspecify.nullness.NullMarked
```
rickie, line 2: Swapping order of the annotations for sorting and consistency with the other file.

rickie (approved):
Added a tiny commit. 

Cool that we'll use JSpecify as default 😄.

oxkitsune (approved):

rickie:
Rebased and resolved conflicts. Will merge once 🍏.

PR #182, opened 2022-08-06
Have `LexicographicalAnnotationListing` sort `TYPE_USE` annotations last

Stephan202:
~:exclamation: This PR is currently on top of #179 :exclamation:~

Suggested commit message:
```
Have `LexicographicalAnnotationListing` sort `TYPE_USE` annotations last (#182)

This ensures compatibility with Error Prone's `AnnotationPosition`
check.
```


Stephan202:

oxkitsune (approved):

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationListingTest.java
```diff
@@ -106,6 +119,13 @@ void identification() {
             "  @Bar",
             "  @Foo(ints = {1, 2})",
             "  A unsortedRepeatableAnnotation();",
+            "",
+            "  // BUG: Diagnostic matches: X",
+            "  @Baz",
```
rickie, line 120: We could add an example that shows _only_ flagging `@{Foo,Bar}TypeUse` annotations. 
I know it is implied but might be nice to have a specific use case for it :).
```
" default @FooTypeUse @BarTypeUse A unsortedTypeAnnotations() {",
"    return null;",
"  }",
```

rickie (approved):
Added a commit.

Got one small thing to consider, but already looks good to me!

(Sorry for the late review 😬 😅 )

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationListingTest.java
```diff
@@ -106,6 +119,13 @@ void identification() {
             "  @Bar",
             "  @Foo(ints = {1, 2})",
             "  A unsortedRepeatableAnnotation();",
+            "",
+            "  // BUG: Diagnostic matches: X",
```
rickie: Forgot to comment the unrelated improvement, but we can just use `Diagnostic contains:` here instead of the `X`.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationListingTest.java
Stephan202: Sure! Rebased and added a commit.

rickie:
Nice! Rebased and will merge once 🟢 !

PR #188, opened 2022-08-11
Introduce `JUnitValueSource` check

oxkitsune:
Implement BugPattern to migrate parameterized JUnit tests with a single argument that's known at compile time to the `@ValueSource` annotation.

The `@ValueSource` annotation supports:
- all primitives and their `java.lang` wrapper
- `java.lang.String`
- `java.lang.Class`

Example:
```java
- @ParameterizedTest
- @MethodSource("fooProvider")
- void foo (int number) {
-   assertThat(number).isNotEqualTo(0);
- }
- 
- private static Stream<Arguments> fooProvider () {
-   return Stream.of(
-     arguments(1),
-     arguments(2)
-   );
- }

+ @ParameterizedTest
+ @ValueSource(ints = {1, 2})
+ void foo (int number) {
+   assertThat(number).isNotEqualTo(0);
+ }
```

Suggested commit message:
```
Introduce `JUnitValueSource` check (#188)
```


rickie (approved):
Added a few commits. PTAL @oxkitsune 😄. 

There are two `XXX`'s which I consider to be out of scope for now but important to write down.

The PR is now ready for your 👀 @Stephan202.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitValueSource.java
```diff
@@ -0,0 +1,181 @@
...
+
+    return valueSourceAttributeValue.map(
+        attributeValue ->
+            SuggestedFix.builder()
+                .addImport("org.junit.jupiter.params.provider.ValueSource")
+                .replace(
+                    methodSourceAnnotation,
+                    String.format(
+                        "@ValueSource(%s = {%s})",
+                        toValueSourceAttributeName(parameterType.toString()), attributeValue))
```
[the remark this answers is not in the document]
Stephan202: For generic types such as `Class<?>` this will also create an invalid `@ValueSource` argument.

Stephan202 (commented):
Rebased and added a commit. Cool feature!

You will notice that I refactored quite a bit. While I do agree that we should be reasonable when defining the scope of a test, we should:
1. Err on the side of correctness, but to avoid false-positives and to ensure a good first impression of users who try out the library.
2. (As a result) aim for very high test coverage.
3. (Where reasonable) try to support code bases which aren't as uniform as Picnic's.

Still TODO:
- Extend test coverage, primarily "should not match" cases. (See the `XXX` comments I left.)
- ~One more full pass over the code.

rickie:
Oops, only ment to re-request review from myself 😬.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitValueSourceTest.java
```diff
@@ -0,0 +1,414 @@
...
+            "  @MethodSource(\"immutableSetOfStringArguments\")",
+            "  void string(String s) {}",
+            "",
+            "  private static Stream<Class<?>> streamOfClassesAndClassArguments() {",
+            "    return Stream.of(Stream.class, java.util.Map.class);",
+            "  }",
+            "",
+            "  @ParameterizedTest",
+            "  @MethodSource(\"streamOfClassesAndClassArguments\")",
+            "  void string(Class<?> c) {}",
```
Stephan202, line 310:
```suggestion
            "  void clazz(Class<?> c) {}",
```
(Don't have the code checked out right now; something to fix later.)

Stephan202 (commented):
@oxkitsune have you started working on the tests? Asking to prevent double work :).

oxkitsune:
> @oxkitsune have you started working on the tests? Asking to prevent double work :).

I have!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitValueSource.java
```diff
@@ -0,0 +1,316 @@
...
+      return ImmutableSet.of(ASTHelpers.constValue(value, String.class));
+    }
+
+    NewArrayTree arrayTree = (NewArrayTree) value;
+    return arrayTree.getInitializers().isEmpty()
+        ? ImmutableSet.of(method.getName().toString())
+        : arrayTree.getInitializers().stream()
+            .map(name -> ASTHelpers.constValue(name, String.class))
+            .collect(toImmutableSet());
+  }
```
Stephan202, line 207: This implementation proves incorrect; pushed a fixed version, including tests, to #335.

Stephan202:
@oxkitsune I see that the commits were squashed, which makes it significantly harder to review the latest changes. Next time please use a merge commit if the conflicts are too complex to resolve otherwise. (Luckily I still had a local copy; will take it from there.)

oxkitsune:
> @oxkitsune I see that the commits were squashed, which makes it significantly harder to review the latest changes. Next time please use a merge commit if the conflicts are too complex to resolve otherwise. (Luckily I still had a local copy; will take it from there.)

There were no conflicts, I squashed because of the large amounts of commits. I now realise that this isn't beneficial as we'll squash when merging into master. My apologies!

rickie:
Rebased.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitValueSource.java
```diff
@@ -0,0 +1,294 @@
...
+ * org.junit.jupiter.params.provider.MethodSource} annotation that can be replaced with an
+ * equivalent {@link org.junit.jupiter.params.provider.ValueSource} annotation.
+ */
+// XXX: Where applicable, also flag `@MethodSource` annotations that reference multiple value
+// factory methods.
+// XXX: Support value factory return expressions of the form `Stream.of(a, b,
+// c).map(Arguments::argument)`.
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Prefer `@ValueSource` over a `@MethodSource` where possible and reasonable",
```
rickie, line 69: I'd say it's nicer to use one word for "possible and reasonable". Maybe "feasible"?

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitValueSourceTest.java
```diff
@@ -0,0 +1,433 @@
...
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+final class JUnitValueSourceTest {
+  private final CompilationTestHelper compilationTestHelper =
+      CompilationTestHelper.newInstance(JUnitValueSource.class, getClass());
+  private final BugCheckerRefactoringTestHelper refactoringTestHelper =
+      BugCheckerRefactoringTestHelper.newInstance(JUnitValueSource.class, getClass());
+
+  // XXX: Combine `identification` tests, focussing on what should _not_ be flagged.
```
rickie, line 14: Rebased, so it appears this check might need some extra love based on the results 👀.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitValueSource.java
```diff
@@ -0,0 +1,294 @@
...
+
+      NewArrayTree newArray = (NewArrayTree) tree;
+      return newArray.getDimensions().isEmpty()
+          && newArray.getInitializers().stream()
+              .allMatch(element -> elementMatcher.matches(element, state));
+    };
+  }
+
+  private static Matcher<ExpressionTree> isArrayArgumentValueCandidate() {
+    return anyOf(classLiteral((t, s) -> true), (t, s) -> ASTHelpers.constValue(t) != null);
```
rickie: Writing out `t, s` is a bit more readable IMO.

rickie (commented):
Added a tiny commit and saw that the rebase unveiled some mutation testing fix-work 👀.

oxkitsune:
For the remaining mutants;
* The mutant on line 145 (`RemoveFilterMutator`) can be ignored, as we want to ignore any value factories declared in other classes, and on line 127, we try to find the value factory in the current class using `hasSiblingReferencingValueFactory` and since the value factory doesn't exist in this class the factory cannot be found. This results in the `matchMethod` returning `Description.NO_MATCH`
* The mutant on line 219 (`NullReturnValsMutator`) can be ignored because we don't use the return value of the `TreeScanner`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitValueSource.java
```diff
@@ -0,0 +1,181 @@
...
+  private static String toValueSourceAttributeName(String type) {
+    switch (type) {
+      case "Class":
+        return "classes";
+      case "Character":
+        return "chars";
+      case "Integer":
+        return "ints";
+      default:
+        return type.toLowerCase() + "s";
```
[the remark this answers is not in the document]
rickie: Fixed :).

rickie (commented):
Added a commit and have a few minor open suggestions.

Man, this check has become a high quality beast! 

Fun one to review! Let's finish the open points and then I'm ready to approve!

rickie:
Apparently didn't rebase 🤔. Anyway did it now and fixed JSpecify rename.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitValueSource.java
```diff
@@ -0,0 +1,296 @@
...
+          anyOf(
+              isPrimitiveOrBoxedPrimitiveType(),
+              isSameType(String.class),
+              isSameType(state -> state.getSymtab().classType)));
+
+  /** Instantiates a new {@link JUnitValueSource} instance. */
+  public JUnitValueSource() {}
+
+  @Override
+  public Description matchMethod(MethodTree tree, VisitorState state) {
```
Stephan202, line 116: To be checked: the [JUnit 5.9.2 release notes](https://junit.org/junit5/docs/5.9.2/release-notes/) mention:
> New `@MethodSource` syntax for explicitly selecting an overloaded local factory method without specifying its fully qualified name.

We should check whether that means another case for this PR to cover.

Stephan202 (commented):
Mostly leaving this as a note to self. Hope that one of these weekends I find the time to finalize my review of this magnificent beast.

Stephan202 (commented):
Rebased and added a commit. ~one more open point for which I added an `XXX`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreJUnitMatchers.java
```diff
@@ -55,25 +55,54 @@ private MoreJUnitMatchers() {}
...
+   * MethodTree)} in that it drops any parenthesized method parameter type enumerations. That is,
+   * method descriptors such as {@code factoryMethod()} and {@code factoryMethod(java.lang.String)}
+   * are both simplified to just {@code factoryMethod}.
+   *
    * @param methodSourceAnnotation The annotation from which to extract value factory method names.
+   * @param method The method on which the annotation is located.
    * @return One or more value factory names.
+   * @see #getMethodSourceFactoryDescriptors(AnnotationTree, MethodTree)
+   */
+  // XXX: Test this method.
```
Stephan202, line 68: I realized after committing that by using this method we:
1. May attempt to replace `foo()` even though the test method references `foo(someParam)`. (Due to loss of details.)
1. May attempt to replace `foo()` even though the test method references both `foo()` and `foo(someParam)`. (Due to loss of details i.c.w. set deduplication.)

Stephan202 (commented):
So a bit more work is needed. Can have a closer look later.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitValueSource.java
```diff
@@ -0,0 +1,302 @@
...
+      anyOf(
+          isArrayArgumentValueCandidate(),
+          toType(
+              MethodInvocationTree.class,
+              allOf(
+                  staticMethod()
+                      .onClass("org.junit.jupiter.params.provider.Arguments")
+                      .namedAnyOf("arguments", "of"),
+                  argumentCount(1),
+                  argument(0, isArrayArgumentValueCandidate()))));
```
Stephan202, line 87: This constraint assumes that the method doesn't have any [additional dependency-injected arguments](https://junit.org/junit5/docs/current/user-guide/#writing-tests-dependency-injection). That's a fair limitation, but let's document it.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitValueSource.java
```diff
@@ -0,0 +1,302 @@
...
+  private static boolean hasSiblingReferencingValueFactory(
+      MethodTree tree, String valueFactory, VisitorState state) {
+    return findMatchingSibling(tree, m -> hasMethodSource(m, valueFactory, state), state)
+        .isPresent();
+  }
+
+  private static Optional<MethodTree> findNullaryFactorySibling(
+      MethodTree tree, String methodName, VisitorState state) {
+    return findMatchingSibling(
+        tree, m -> m.getParameters().isEmpty() && m.getName().contentEquals(methodName), state);
```
Stephan202, line 163: The factory method doesn't actually need to be nullary; see the [documented](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-MethodSource) `@RegisterExtension` example.

Stephan202 (commented):
Documenting some additional caveats; will try to cover these in one go.

Stephan202 (approved):
Rebased and added one more commit. Documented as many limitations as I could identify; those are for another day.

Suggested commit message:
```
Introduce `JUnitValueSource` check (#188)

This new check replaces JUnit `@MethodSource` usages with an equivalent
`@ValueSource` annotation where possible.
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitValueSource.java
```diff
@@ -0,0 +1,307 @@
...
+      return newArray.getDimensions().isEmpty()
+          && !newArray.getInitializers().isEmpty()
+          && newArray.getInitializers().stream()
+              .allMatch(element -> elementMatcher.matches(element, state));
+    };
+  }
+
+  private static Matcher<ExpressionTree> isArrayArgumentValueCandidate() {
+    return anyOf(
+        classLiteral((tree, state) -> true), (tree, state) -> ASTHelpers.constValue(tree) != null);
```
Stephan202: Just realized we can write this in a nicer way.

Stephan202 (commented):
Added one more commit.

rickie (approved):
Yoooooo the quality and epicness of this check 🤯🤯🤯. 

Great improvements @Stephan202 , let's get this beast merged and see the impact 😄.

Added a commit to drop a space and will rebase. After that when its 🍏 ==> merge!

PR #194, opened 2022-08-13
Introduce `StringJoin` check

Stephan202:
Suggested commit message:
```
Introduce `StringJoin` check (#194)

This new `BugChecker` flags `String#format` invocations which can be
replaced with a `String#join` or even `String#valueOf` invocation.
```


Stephan202:
Rebased and resolved conflict. @rickie who shall we ask for a second review? :smile:

EnricSala (approved):
Very cool :) Nothing from my side.

PR #198, opened 2022-08-14
Update all tests to use text blocks

Stephan202:
~**Edit:** PR #603 was spun off from this PR. That change should be merged and finalized first.~

Suggested commit message:
```
Update all tests to use text blocks (#198)

All `CompilationTestHelper`, `BugCheckerRefactoringTestHelper` and
`Compilation` test code is now specified using a single text block.
These changes were automated thanks to the new text block support added
to the `TestHelperSourceFormat` check (renamed from
`ErrorProneTestHelperSourceFormat`).
```


error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/AutowiredConstructorTest.java
```diff
@@ -16,94 +16,103 @@ void identification() {
...
+              class H {
+                @SafeVarargs
+                H(List<String>... lists) {}
+              }
+            }
+            """)
         .doTest();
   }
 
+  @SuppressWarnings("RegexpMultiline" /* Check may introduce empy line at start of code block. */)
```
rickie, line 74: A while back you mentioned this could be a possible Checkstyle contribution, so I checked whether that would be the case. Found some related issues, but the gist is that we can work around it.

So for when we get back to this PR in the future, I'll link the issue(s):
https://github.com/checkstyle/checkstyle/issues/11742

With example suppression: https://github.com/checkstyle/checkstyle/issues/11653#issuecomment-1180373074.

Stephan202: Check! I think the proposed regexes for `SuppressWithPlainTextCommentFilter` would need some tweaking for our case, but indeed that could work. That said, for this specific case it would be nicer if we emit a better suggested fix; for this I filed #347.

Stephan202:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/SourceCode.java
```diff
@@ -1,16 +1,53 @@
...
+   *
+   * @param state A {@link VisitorState} from which the targeted Java language level can be derived.
+   * @return {@code true} iff text block expressions are supported.
+   */
+  // XXX: Add tests!
+  public static boolean isTextBlockSupported(VisitorState state) {
+    // XXX: String comparison is for JDK 11 compatibility. Is there a better way?
+    return Target.instance(state.context).toString().compareTo("JDK1_15") >= 0;
+    // return Target.instance(state.context).compareTo(Target.JDK1_15) >= 0;
+  }
```
Stephan202, line 30: Note to future self: there's a better way of doing this; see #3646. We can then also revert all references to `com.sun.tools.javac.jvm` in this PR.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/AmbiguousJsonCreatorTest.java
```diff
@@ -20,99 +20,100 @@ void identification() {
...
-            "      this.s = s;",
-            "    }",
-            "",
-            "    @JsonCreator",
-            "    public static F of(String s) {",
-            "      return new F(s);",
-            "    }",
-            "  }",
-            "}")
+            """
```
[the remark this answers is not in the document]
Stephan202: Alright, after trying several approaches I landed on a solution involving [Jabel](https://github.com/bsideup/jabel) and [setup-java](https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#modifying-maven-toolchains)'s support for [Maven Toolchains](https://maven.apache.org/guides/mini/guide-using-toolchains.html), using which we can configure Surefire to run tests using JDK 11, even though the rest of the build uses JDK 17. This does rely on [SUREFIRE-1234](https://issues.apache.org/jira/browse/SUREFIRE-1234), forcing the upgrade to a Surefire pre-release version. It all feels a bit wonky, but (a) I didn't find a nicer alternative and (b) we can drop this once we drop support for JDK 11 (see also google/error-prone#3803).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/AmbiguousJsonCreatorTest.java
```diff
@@ -20,99 +20,100 @@ void identification() {
...
+                }
+
+                @JsonCreator(mode = JsonCreator.Mode.DELEGATING)
+                public static B of(int i) {
+                  return FOO;
+                }
+              }
+
+              enum C {
+                FOO(1, \"s\");
```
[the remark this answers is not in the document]
Stephan202: This has been resolved since.

Stephan202:
The current state of the PR contains a Surefire upgrade, as it relied on functionality only available in a milestone release. This upgrade is now part of #532, to be merged first.

Stephan202:
I filed #603 for the Jabel build changes; once merged the scope of this PR will be reduced to just the `CompilationTestHelper`/`BugCheckerRefactoringTestHelper` text block formatting support.

Stephan202:
Rebased, resolved conflicts, and made the changes more compatible with Jabel.

Stephan202:
Rebased and resolved conflicts and other issues. The PR now targets `master`. Next up would be to get the code in a reviewable state, primarily by extending test coverage. One other key decision to make, is whether to keep support for the "don't use text blocks" mode. During development this has been very helpful, so perhaps we should keep it for now, but over time I expect that this functionality will become disused. (Though I suppose it'll simplify contributing checks upstream.)

NB: I added a commit with a proposal to rename the check. I've updated the suggested commit message accordingly.

PR #214, opened 2022-08-24
Introduce `JUnitClassModifiers` check

oxkitsune:
Implement BugPattern to ensure JUnit test classes are declared as package private final.

Example:
```java
-public class A {
-  @Test
-  void foo () {}
-}

+final class A {
+  @Test
+  void foo () {}
+}
```

Suggested commit message:
```
Introduce `JUnitClassModifiers` check (#214)
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassDeclaration.java
```diff
@@ -0,0 +1,73 @@
...
+              hasModifier(Modifier.PUBLIC)));
+
+  @Override
+  public Description matchClass(ClassTree tree, VisitorState state) {
+    if (!NOT_FINAL_TEST_CLASS.matches(tree, state)) {
+      return Description.NO_MATCH;
+    }
+
+    SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
+    SuggestedFixes.addModifiers(tree, state, FINAL).ifPresent(fixBuilder::merge);
```
rickie, line 67:
```suggestion
    SuggestedFixes.addModifiers(tree, state, Modifier.FINAL).ifPresent(fixBuilder::merge);
```
Hmm, I'm a little on the fence about using `Modifier.` here. It is kind of clear from the `addModifiers` context. OTOH, `FINAL` is not something we use a lot, so to make clear where it comes from we could add the `Modifiers.`. WDYT?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreMatchers.java
```diff
@@ -0,0 +1,35 @@
...
+import com.sun.source.tree.Tree;
+import com.sun.tools.javac.code.Symbol;
+
+/**
+ * A collection of extra static factory methods to make the {@link Matcher} DSL read more fluently.
+ */
+public final class MoreMatchers {
+  private MoreMatchers() {}
+
+  private static TypePredicate hasAnnotation(String annotationClassName) {
```
rickie, line 15: This method is used in the other method, therefore it should go below that one.

rickie (changes requested):
Added a commit. 

We need to think about nested classes. What if that occurs, what are the rules w.r.t. modifiers. Additionally, we need a test case for that 😉. 

Maybe we could to a test run on an internal codebase to see how it performs 😄.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitMethodDeclaration.java
```diff
@@ -185,18 +184,4 @@ private static Optional<String> tryCanonicalizeMethodName(MethodTree tree) {
...
-  private static Matcher<AnnotationTree> hasMetaAnnotation(String annotationClassName) {
-    TypePredicate typePredicate = hasAnnotation(annotationClassName);
-    return (tree, state) -> {
-      Symbol sym = ASTHelpers.getSymbol(tree);
-      return sym != null && typePredicate.apply(sym.type, state);
-    };
-  }
-
-  // XXX: Move to a `MoreTypePredicates` utility class.
-  private static TypePredicate hasAnnotation(String annotationClassName) {
```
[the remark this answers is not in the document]
rickie: Hmm, I think we can defer this for now indeed 🤔.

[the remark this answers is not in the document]
rickie: Maybe it'd be better to move this XXX with the method and add the words `Consider to move .....`. WDYT?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassDeclaration.java
oxkitsune: Yea I was thinking the same thing, but ultimately opted to statically import as I figured it would be clear from the context. I'd say we just add it back.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitMethodDeclaration.java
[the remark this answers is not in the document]
oxkitsune: Let's move the XXX to the method, and then hold off making the `MoreTypePredicates` class until there's more `TypePredicate` util methods.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitClassDeclarationTest.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.errorprone.BugCheckerRefactoringTestHelper.TestMode;
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+final class JUnitClassDeclarationTest {
+  private final CompilationTestHelper compilationTestHelper =
+      CompilationTestHelper.newInstance(JUnitClassDeclaration.class, getClass());
+  private final BugCheckerRefactoringTestHelper refactoringTestHelper =
+      BugCheckerRefactoringTestHelper.newInstance(JUnitClassDeclaration.class, getClass());
+
```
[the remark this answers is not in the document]
oxkitsune: I've added a test case for a nested class. 
However, right now the check also flags nested classes with `@Test` annotated methods WITHOUT the required `@Nested` annotation. Should we ignore nested classes if it doesn't have the `@Nested` annotation?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassDeclaration.java
```diff
@@ -0,0 +1,76 @@
...
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "JUnit test classes should be declared as package private final",
+    linkType = CUSTOM,
+    link = BUG_PATTERNS_BASE_URL + "JUnitClassDeclaration",
+    severity = WARNING,
+    tags = FRAGILE_CODE)
+public final class JUnitClassDeclaration extends BugChecker implements ClassTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final ImmutableSet<Modifier> ILLEGAL_MODIFIERS =
```
rickie, line 45: Based on [`RefasterRuleModifiers`](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/RefasterRuleModifiers.java) we can inline the `ILLEGAL_MODIFIERS` field.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassDeclaration.java
```diff
@@ -0,0 +1,76 @@
...
+import com.sun.source.tree.AnnotationTree;
+import com.sun.source.tree.ClassTree;
+import com.sun.source.tree.MethodTree;
+import javax.lang.model.element.Modifier;
+
+/**
+ * A {@link BugChecker} which flags non-final or non package-private JUnit test class declarations
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
```
rickie, line 37: We can rename the class to `JUnitClassModifiers` instead of `Declaration`. WDYT?

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitClassDeclarationTest.java
```diff
@@ -0,0 +1,99 @@
...
+            "  @Test",
+            "  void foo() {}",
+            "}")
+        .addSourceLines(
+            "B.java",
+            "import org.junit.jupiter.params.ParameterizedTest;",
+            "",
+            "// BUG: Diagnostic contains:",
+            "class B {",
+            "  @ParameterizedTest",
```
rickie, line 32: Is there a specific reason to use `@ParameterizedTest`? To also account for that case?

rickie (commented):
Added a commit.

Also asked a few questions :).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitClassDeclarationTest.java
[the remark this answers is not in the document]
rickie: I think it is nice to match these instances TBH. The `@Nested` annotation is not **required**, right? 
https://junit.org/junit5/docs/5.4.1/api/org/junit/jupiter/api/Nested.html

It's in some cases nice to use but not a required thing. I think it should be possible to do this, as it is valid code. Therefore we should flag this case as well :).

Maybe we should add a testcase for this as well though 😉.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitClassDeclarationTest.java
oxkitsune: Yes, we check whether the class contains methods annotated with `@Test` or with _meta_ annotation `org.junit.jupiter.api.TestTemplate`. `@ParameterizedTest` is annotated with `@TestTemplate` and this test ensures we also match tests like this.

rickie (changes requested):
Added a commit. 

I think this is a really nice improvement. Would love to see this applied on our internal codebase :).

I found some edge cases that we should consider:
- If something is abstract, it cannot be final. So we need to exclude that _and_ add a test case for that :).
- In our shared modules we have some `public abstract` classes with tests that can be used by downstream projects. These classes are **not** located in `/src/test/` and marked `public` on purpose. I think this use case is rather specific to our shared modules repository, but IMO we should do something with it. On the other hand, I'm not sure if looking at the file's location is something we should add as heuristic for flagging these things.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
```diff
@@ -0,0 +1,80 @@
...
+  private static final long serialVersionUID = 1L;
+  private static final MultiMatcher<MethodTree, AnnotationTree> TEST_METHOD =
+      annotations(
+          AT_LEAST_ONE,
+          anyOf(
+              isType("org.junit.jupiter.api.Test"),
+              hasMetaAnnotation("org.junit.jupiter.api.TestTemplate")));
+  private static final Matcher<ClassTree> NON_FINAL_TEST_CLASS =
+      allOf(
+          not(hasMetaAnnotation("org.springframework.context.annotation.Configuration")),
```
rickie, line 53: Just like we did for the `TestTemplate`, by using the `ParameterizedTest` in the tests, we should also verify the `@Configuration` meta annotation.

rickie (commented):
Another thought :).

oxkitsune:
Our `MoreMatchers#hasMetaAnnotation(String)` method cannot match Spring's `@Configuration` as meta annotation. 
I suspect this has something to do with the fact that `@Configuration` is not explicitly `@Inherited`. 
I've implemented a quick work around that flags some often used `@Configuration` annotations for now.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
```diff
@@ -0,0 +1,85 @@
...
+          not(
+              hasMetaAnnotation(
```
rickie: We can combine these three `not(hasMetaAnnotation(` by wrapping them in a `not(anyOf(`.

rickie (approved):
Reverted the introduction of `MoreMatchers` as this will be part of #335. 

This means we'll temporarily introduce two duplicate methods `hasMetaAnnotation` and `hasAnnotation`. However, they will soon be de-duplicated by #335. 
I seems better to focus this PR solely on introducing the `JUnitClassModifiers` check. 

Suggested commit message (repeating PR title):
```
Introduce `JUnitClassModifiers` check (#214)
```

Really cool check @oxkitsune , this will have **many** impact on other teams 🚀 !

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * A {@link BugChecker} that flags non-final and non package-private JUnit test class declarations.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "JUnit test classes should be declared as package private final",
+    linkType = CUSTOM,
+    link = BUG_PATTERNS_BASE_URL + "JUnitClassDeclaration",
+    severity = WARNING,
+    tags = FRAGILE_CODE)
```
rickie, line 45:
```suggestion
    tags = SIMPLIFICATION)
```
Maybe this would fit better? Or `STYLE`?

rickie (commented):
One open question from my side left.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
rickie: Pushing something as it certainly does not match the description of `FRAGILE_CODE`.

rickie (approved):
Re-approving the latest changes. 

Have one thing to ponder, in `RefasterRuleModifiersTest` we give more descriptive names to the classes. This makes it easier to see what exact case is being tested. We _could_ decide to do the same here. Already liking the existing setup, so approving regardless.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiersTest.java
```diff
@@ -0,0 +1,133 @@
...
+            "  }",
+            "}")
+        .addSourceLines(
+            "E.java",
+            "import org.junit.jupiter.api.Test;",
+            "",
+            "final class E {",
+            "  @Test",
+            "  void foo() {}",
+            "}")
```
Stephan202: I'd place this one first; we usually list the non-violating test cases first.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiersTest.java
```diff
@@ -0,0 +1,133 @@
...
+            "class B {",
+            "  @ParameterizedTest",
+            "  void foo() {}",
+            "}")
+        .addSourceLines(
+            "C.java",
+            "import org.junit.jupiter.api.Test;",
+            "",
+            "// BUG: Diagnostic contains:",
+            "public class C {",
```
Stephan202: This one is flagged because it's non-final; to test the `public` trigger the class should be `final`.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiersTest.java
```diff
@@ -0,0 +1,133 @@
...
+            "import org.junit.jupiter.api.Nested;",
+            "import org.junit.jupiter.api.Test;",
+            "",
+            "class D {",
+            "  @Nested",
+            "  // BUG: Diagnostic contains:",
+            "  class Nested1 {",
+            "    @Test",
+            "    void foo() {}",
+            "  }",
```
Stephan202: The `@Nested` doesn't influence the checker one way or another :eyes:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.lang.model.element.Modifier;
+
+/**
+ * A {@link BugChecker} that flags non-final and non package-private JUnit test class declarations.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "JUnit test classes should be declared as package private final",
+    linkType = CUSTOM,
+    link = BUG_PATTERNS_BASE_URL + "JUnitClassDeclaration",
```
Stephan202:
```suggestion
    link = BUG_PATTERNS_BASE_URL + "JUnitClassModifiers",
```

Stephan202 (commented):
Rebased and added a commit. Only open point from my side is the `@Configuration` case. Feel free to send me example classes over Slack.

Updated the suggested commit message.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
```diff
@@ -0,0 +1,79 @@
...
+  private static final long serialVersionUID = 1L;
+
+  private static final Matcher<ClassTree> NON_FINAL_TEST_CLASS =
+      allOf(
+          not(
+              annotations(
+                  AT_LEAST_ONE,
+                  anyOf(
+                      isType("org.springframework.context.annotation.Configuration"),
+                      hasMetaAnnotation("org.springframework.context.annotation.Configuration")))),
```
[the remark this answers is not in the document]
oxkitsune: any springboot tests using an annotation marked with `@Configuration` cannot be marked final. Usually these classes can be package private, so I'm in favour of doing that.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
```diff
@@ -0,0 +1,91 @@
...
+  /** Instantiates a new {@link JUnitClassModifiers} instance. */
+  public JUnitClassModifiers() {}
+
+  @Override
+  public Description matchClass(ClassTree tree, VisitorState state) {
+    if (!TEST_CLASS_WITH_INCORRECT_MODIFIERS.matches(tree, state)) {
+      return Description.NO_MATCH;
+    }
+
+    boolean hasSpringConfiguration = TEST_CLASS_WITH_SPRING_CONFIGURATION.matches(tree, state);
```
rickie, line 73: I really like the code. Got one minor point to consider, do we want to name it `TEST_CLASS_WITH_SPRING_CONFIGURATION_ANNOTATION`? It's a bit long I'd say but `WITH_SPRING_CONFIGURATION` out of context is not that clear.

rickie (commented):
Let's get this one out 😄.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
```diff
@@ -0,0 +1,91 @@
...
+    linkType = CUSTOM,
+    link = BUG_PATTERNS_BASE_URL + "JUnitClassModifiers",
+    severity = SUGGESTION,
+    tags = STYLE)
+public final class JUnitClassModifiers extends BugChecker implements ClassTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<ClassTree> TEST_CLASS_WITH_INCORRECT_MODIFIERS =
+      allOf(
+          hasMethod(TEST_METHOD),
+          not(hasModifier(Modifier.ABSTRACT)),
```
Stephan202, line 53: Let's move this case down so that the modifier ordering here matches the resolution order below.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
```diff
@@ -0,0 +1,91 @@
...
+    SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
+    SuggestedFixes.removeModifiers(
+            tree.getModifiers(),
+            state,
+            ImmutableSet.of(Modifier.PRIVATE, Modifier.PROTECTED, Modifier.PUBLIC))
+        .ifPresent(fixBuilder::merge);
+
+    if (!hasSpringConfiguration) {
+      SuggestedFixes.addModifiers(tree, state, Modifier.FINAL).ifPresent(fixBuilder::merge);
+    }
```
Stephan202: Pitest doesn't flag it because we're not using the [`STRONGER`](https://pitest.org/quickstart/mutators/) mutators group (I'll fix that), but right now the tests also pass if this suggested fix is applied unconditionally. We need to extend the `replacement` test as well.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
Stephan202: ^ I'll shuffle the test code accordingly.

Stephan202 (approved):
Rebased and added one more commit. Repeating the suggested commit message:
```
Introduce `JUnitClassModifiers` check (#214)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiers.java
Stephan202: > I'll fix that

I filed #383.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiersTest.java
```diff
@@ -0,0 +1,129 @@
...
+            "  class NonFinal {",
+            "    @Test",
+            "    void foo() {}",
+            "  }",
+            "",
+            "  // BUG: Diagnostic contains:",
+            "  class NonFinalWithCustomTestMethod {",
+            "    @ParameterizedTest",
+            "    void foo() {}",
+            "  }",
```
werli, line 75: I feel like these two cases are misleading: IIUC, the latter doesn't fail due to the "custom test method" but solely because it's non-final. Did I get this right?

Shouldn't we add a _positive_ test case for test methods annotated with `@ParameterizedTest`? Though one could argue this is already covered by the test cases for `MoreJUnitMatchers#TEST_METHOD` in `MoreJUnitMatchersTest`.

werli (approved):
The rule itself LGTM, but I have one more question on one more of the test cases.

I'm optimistically approving however.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitClassModifiersTest.java
Stephan202: > Did I get this right?

Yes :)

> Shouldn't we add a _positive_ test case for test methods annotated with `@ParameterizedTest`?

Can do :)

Stephan202 (commented):
Tnx for the review @werli! Added one more commit.

rickie:
Nice reviews and commits 🚀 !

PR #251, opened 2022-09-20
Update `BugPattern#{link,linkType.CUSTOM}` to link to our website

rickie:


Stephan202 (approved):
Rebased and added a commit to update a recently introduced bug checker. 

> Rebased and added a commit. Not yet approving because I want to asses the effort required to automate/enforce this (started playing with something).

I'll defer completion of the enforcement check to a separate PR. (It is pretty clear that this will be necessary, though. For example, right now the `Refaster` check doesn't define a URL, and its module (`refaster-runner`) does not have access to `Documentation.BUG_PATTERNS_BASE_URL`.)

PR #255, opened 2022-09-21
Emit website link along with Refaster refactor suggestions

Stephan202:
❗ ~~This PR is on top of #286~~ ❗ 

Work in progress; see [this thread](https://github.com/PicnicSupermarket/error-prone-support/pull/251#discussion_r976713663) for context.


rickie (commented):
Not done reviewing, flushing what I have so far.

refaster-runner/src/test/java/tech/picnic/errorprone/refaster/runner/RefasterTest.java
```diff
@@ -0,0 +1,243 @@
...
+            "  }",
+            "}")
+        .doTest();
+  }
+
+  private static Stream<Arguments> reportedSeverityTestCases() {
+    /* { arguments, expectedSeverities } */
+
+    Stream<Arguments> forkTestCases =
+        isBuiltWithErrorProneFork()
```
Stephan202: "Built" is past tense, while this is about the current classpath :)

refaster-runner/src/test/java/tech/picnic/errorprone/refaster/runner/RefasterTest.java
```diff
@@ -0,0 +1,243 @@
...
+              "com.google.errorprone.ErrorProneOptions",
+              /* initialize= */ false,
+              Thread.currentThread().getContextClassLoader());
+    } catch (ClassNotFoundException e) {
+      throw new IllegalStateException("Can't load `ErrorProneOptions` class", e);
+    }
+    return Arrays.stream(clazz.getDeclaredMethods())
+        .filter(m -> Modifier.isPublic(m.getModifiers()))
+        .anyMatch(m -> m.getName().equals("isSuggestionsAsWarnings"));
+  }
```
Stephan202: Let's reuse the other code; will push a proposal.

refaster-support/src/main/java/tech/picnic/errorprone/refaster/annotation/package-info.java
```diff
@@ -0,0 +1,8 @@
+/**
+ * A collection of annotations that may be placed on Refaster template classes and Refaster template
```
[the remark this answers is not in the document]
Stephan202: "Meh" :stuck_out_tongue:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/OptionalTemplates.java
```diff
@@ -17,8 +17,11 @@
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
+import tech.picnic.errorprone.refaster.annotation.Description;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
 
 /** Refaster templates related to expressions dealing with {@link Optional}s. */
+@OnlineDocumentation
```
Stephan202, line 24: Here and below: don't forget to revert the changes in this file, assuming we introduce the annotations separately.

Stephan202 (commented):
Rebased and added a commit.

Stephan202 (commented):
Added a commit.

refaster-runner/src/test/java/tech/picnic/errorprone/refaster/runner/RefasterTest.java
```diff
@@ -0,0 +1,243 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+final class RefasterTest {
+  private final CompilationTestHelper compilationHelper =
+      CompilationTestHelper.newInstance(Refaster.class, getClass())
+          .matchAllDiagnostics()
+          .expectErrorMessage(
+              "StringOfSizeZeroTemplate",
+              containsPattern(
+                  "\\[Refaster Rule\\] FooTemplates\\.StringOfSizeZeroTemplate: Refactoring opportunity\\s+.+\\s+null"))
```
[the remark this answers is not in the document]
Badbond: According to the `Description.Builder`, `link` is `Nullable`. We could instead of an empty string pass in `null` at `AnnotatedCompositeCodeTransformer#augmentDescription` if we don't have a `url`, causing this to not appear at all (verified this). This might be nice to do in this transition period until all checks and templates have a link. Got a commit ready if we wish to do this.

Badbond (commented):
Flushing first part of review and added a commit.

Stephan202 (commented):
Tnx for the review so far @Badbond! Added a small commit :)

refaster-runner/src/test/java/tech/picnic/errorprone/refaster/runner/RefasterTest.java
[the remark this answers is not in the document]
rickie: > Wanna open a PR upstream? :)

Yes that sounds good. Would be a nice improvement.
Will write it down and do after the EPS OSS-ing journey 😛.

refaster-runner/src/main/java/tech/picnic/errorprone/refaster/runner/Refaster.java
```diff
@@ -118,12 +123,52 @@ private static void applyMatches(
...
+   * <p>The original check name (i.e. the Refaster template name) is prepended to the {@link
+   * Description}'s message. The replacement check name ("Refaster Template") is chosen such that it
+   * is guaranteed not to match any canonical bug checker name (as that could cause {@link
+   * VisitorState#reportMatch(Description)}} to override the reported severity).
+   */
+  private static Description augmentDescription(
+      Description description, Optional<SeverityLevel> severityOverride) {
+    return Description.builder(
+            description.position,
+            "Refaster Rule",
```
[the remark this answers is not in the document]
rickie: We use both `Refaster Rule` and `Refaster Template` a lot. Not really sure if we need to settle on using one of the two. 
Now it depends a bit on the situation, I feel.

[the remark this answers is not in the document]
rickie: We use both `Refaster Rule` and `Refaster Template` a lot. Not really sure if we need to settle on using one of the two. 
Now it depends a bit on the situation, I feel.

To be honest, I try to only use the word "template" to avoid confusing and make people use the same words. I'd be okay with switching everything to template. WDYT?

rickie (commented):
Sooo, the open points for now (that are a MUST before merging) are:

- Settle on a name for the `AnnotatedCompositeCodeTransformer`.
- Revert the changes in `OptionalTemplates`. 
- Fixing the `runtime` scope declaration of `refaster-compiler`.

For me a review of the latest commits and `RefasterRuleCompilerTaskListener` left.

refaster-runner/src/main/java/tech/picnic/errorprone/refaster/runner/Refaster.java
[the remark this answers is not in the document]
Badbond: Can't find a clear distinction between a rule and a template in the [original documentation](https://errorprone.info/docs/refaster) either. I'm happy with both (slightly prefer template, given the annotations), as long as we are consistent :+1:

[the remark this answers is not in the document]
Stephan202: Yeah, this has bugged me for a while as well. I _suspect_ it's more correct to call "the whole thing" a Refaster _rule_, since (a) it's represented by the `RefasterRule` type and (b) the individual methods are called before/after _templates_.

So that would make "Refaster Rule" correct for this code.

That still leaves many incorrect usages elsewhere in the code base... maybe something for a separate PR?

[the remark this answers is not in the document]
rickie: Yea let's defer to another PR. 

I'm also in camp let's be consistent as @Badbond is haha. I bet you are as well. 
It makes slightly more sense to say "lets add a rule for that" instead of "lets add a template for that".

refaster-support/src/main/java/tech/picnic/errorprone/refaster/AnnotatedCompositeCodeTransformer.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
+import tech.picnic.errorprone.refaster.annotation.Severity;
+
+/**
+ * A {@link CompositeCodeTransformer} that can be augmented with annotations.
+ *
+ * <p>The content of the annotations can be used for enriched compilation warnings or errors, and
+ * documentation purposes.
```
Badbond, line 36: Thanks for the Javadocs @rickie! I jotted something down in a private note when I had a sudden brainpulse a while ago:

> A composite code transformer that enhances the description of matches from Refaster rules based on optionally available annotations.

Placing it here to see if (part of it) might be nice to still adopt. Specifically, ^ is more focused around _current_ functionality and Refaster rules specifically. Feel free to consider `s/enhances/augments/` and `s/rules/templates/`.

Badbond (commented):
Another intermittent review for posting the following suggested commit message. Feel free to tweak/suggest an alternative. Slightly different than from https://github.com/PicnicSupermarket/error-prone-support/commit/0561c371de3ca6bf428d118207654e27009a1b5a given that we have changed more here.

Suggested commit message:

```
Augment `Description`s of Refaster rule matches (#255)

By emitting a website link, if available, and reporting the matching Refaster
rule in the description's message rather than its check name. Additionally, it
is now possible to assign a custom severity to Refaster rules and to override 
the severity of matches for all Refaster rules by passing
`-Xep:Refaster:[SEVERITY]`.
```

rickie (approved):
Added a commit.

W.r.t the suggested commit message; made some tweaks, not sure if all of them are good. The capitalized `SEVERITY` is to show that a severity should be passed in all caps. Besides that I'm not sure if "than its check name" is completely correct/clear. 
Lastly, I feel that the Refaster Severity overrides is part of the main goal of the PR not really a "while there". Although the "while there" is not wrong 😬, feel free to revert the changes. Felt like the changes needed some explanations, as in general I like the suggested commit message.

The PR itself is really cool, took some time to wrap my had around this, but man this is some nice stuff!  

Added another commit to revert the changes in `OptionalTemplates`. 

Now the only only open point is: 
- Settle on a name for the `AnnotatedCompositeCodeTransformer`.

Nonetheless, I'm approving as I think the state of the PR is good!
(I'm already okay with `AnnotatedCompositeCodeTransformer` for now 😄)

refaster-support/src/main/java/tech/picnic/errorprone/refaster/AnnotatedCompositeCodeTransformer.java
Badbond: While debugging, I got a better understanding of how the annotations are retrieved and passed down. As such, I understood that the annotations available inside _this_ `AnnotatedCompositeCodeTransformer` are (per annotation type) of the first available parent class (i.e. of one of the encompassing parent template collections) of the `RefasterRule` `CodeTransformer`s herein. Here, we augment descriptions on matches per respective `CodeTransformer` based on annotations available on said `CodeTransformer` and otherwise use the encompassing annotations mentioned above (not that this class knows, it just has been given a list of fallback annotations to rely upon).

As such, I'm wondering whether we should clarify this Javadoc accordingly as 'the annotations' is currently a bit ambigious over which annotations in the context of (`Composite`) `CodeTransformer`s we are talking about. Committed a suggestion, PTAL. Not entirely happy with 'as a fallback' but was less happy with words such as 'otherwise' or 'default'.

Also not sure if we should mention the annotations in the summary given the class' name, but left it for now. I did change the summary line as semantically it was saying that the annotations could augment this `CompositeCodeTransformer`. Also removed the XXX for lack of a better alternative on my side too.

Feel free to revert or change if I did not get the design of this class correctly :sweat_smile:

Badbond (approved):
Thanks for explaining offline how to debug `RefasterRuleCompilerTaskListener` @rickie :+1:. That really helped in reviewing this PR. Good job @Stephan202 and @rickie :+1: Very fun one to review.

Pushed one proposal. All LGTM. Would be happy to discuss an alternative name for the `AnnotatedCompositeCodeTransformer` but current name is good enough for me. Let's see how this class evolves and whether that would cause a name change. Hard to predict now.

Stephan202:
Thanks for the reviews @Badbond and @rickie! I've rebased then branch and added a small commit. Will have a closer look at the suggested changes later; this PR is the next one on my list :smile:

refaster-support/src/main/java/tech/picnic/errorprone/refaster/annotation/OnlineDocumentation.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * The URL or URL pattern of the website at which the annotated Refaster template(s) are
+   * documented.
+   *
+   * @return A non-{@code null} string, optionally containing the {@value
+   *     TOP_LEVEL_CLASS_URL_PLACEHOLDER} and {@value NESTED_CLASS_URL_PLACEHOLDER} placeholders.
+   */
+  String value() default
+      "https://error-prone.picnic.tech/refastertemplates/"
```
Stephan202, line 45: If we agree that they're _rules_, then we should update the website URLs before widely publicizing it...

refaster-support/src/main/java/tech/picnic/errorprone/refaster/ErrorProneFork.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Optional;
+
+/**
+ * Utility class that enables the runtime to determine whether Picnic's fork of Error Prone is on
+ * the classpath.
+ *
+ * @see <a href="https://github.com/PicnicSupermarket/error-prone">Picnic's Error Prone fork</a>
+ */
+// XXX: Add tests for this class. We can at least test `#isErrorProneForkAvailable()` by having
+// Maven inject a corresponding system property.
```
Stephan202: We should indeed add tests. Will propose something.

Stephan202 (commented):
Rebased and added a commit. Also updated the suggested commit message; PTAL!

refaster-support/src/main/java/tech/picnic/errorprone/refaster/annotation/OnlineDocumentation.java
Badbond: Having an agreement is indeed a blocker for the PR to not have to change this URL. And I agree with the idea to call the whole thing a rule and the part of it the templates (i.e. template of a piece of code that the rule matches/writes to). However, I haven't (yet) fully read the original author's [paper](https://research.google.com/pubs/archive/41876.pdf). So ultimately I will agree with what you, @Stephan202 and @rickie, agree on given your greater experience in using and developing these.

refaster-support/src/main/java/tech/picnic/errorprone/refaster/AnnotatedCompositeCodeTransformer.java
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * A {@link CompositeCodeTransformer} that augments the {@link Description} of Refaster rule
+ * matches.
+ *
+ * <p>The content is augmented based on custom {@link tech.picnic.errorprone.refaster.annotation
+ * annotations} available on the matching {@link CodeTransformer} or on this {@link
+ * CompositeCodeTransformer} as a fallback, if any.
+ */
+// XXX: Test this class directly. (Right now it's only indirectly tested through `RefasterTest`.)
```
Badbond, line 40: Is this an XXX we wish to resolve as part of this PR? Haven't gave it much thought, but I feel the approach will be quite the same (have some dummy Refaster rules with these annotations).  I'm okay with leaving it like this (we clearly state the connection), or, if it makes sense, to move some of the introduced testcases from `RefasterTest` to a new test class.

refaster-support/src/test/java/tech/picnic/errorprone/refaster/ErrorProneForkTest.java
```diff
@@ -0,0 +1,65 @@
...
+        .addSourceLines(
+            "A.java",
+            "// BUG: Diagnostic contains: Suggestions as warnings enabled: false",
+            "class A {}")
+        .doTest();
+  }
+
+  @Test
+  void isSuggestionsAsWarningsEnabledWithFlag() {
+    assumeTrue(ErrorProneFork.isErrorProneForkAvailable());
```
Badbond, line 36: TIL, very nice i.c.w. the `error-prone-fork` profile configuration. This is one thing I want to run in an IDE though to see 😄 Will try to do so tomorrow.

Badbond (commented):
Thanks for your commit @Stephan202, very nice. 👍

refaster-support/src/main/java/tech/picnic/errorprone/refaster/AnnotatedCompositeCodeTransformer.java
Stephan202: My thinking is that we can test this class using a "lower-level" integration, possibly mocks. I'm okay with deferring that work to another PR, though, so merging isn't blocked on this. (But who knows, perhaps I'll have a look at this later today.)

refaster-support/src/main/java/tech/picnic/errorprone/refaster/annotation/OnlineDocumentation.java
Stephan202: > Edit: and I'd scope the name consistency change only for the URL for now and touched LoC, to proceed with this PR.

Indeed. In fact, I'd even be okay with deferring all renaming to a separate PR, since the current use of "template", even if we wish to change it, is quite consistent with the rest of the code base.

Stephan202 (commented):
Tnx for the late-night review @Badbond :smile:

refaster-support/src/main/java/tech/picnic/errorprone/refaster/annotation/OnlineDocumentation.java
rickie: Maybe that would even be the best option, deferring to a separate PR. Anyway, yes it should be changed 💪🏻.

refaster-support/src/main/java/tech/picnic/errorprone/refaster/AnnotatedCompositeCodeTransformer.java
```diff
@@ -0,0 +1,153 @@
...
+@AutoValue
+public abstract class AnnotatedCompositeCodeTransformer implements CodeTransformer, Serializable {
```
Stephan202, line 42: Heads up; I'm currently thinking about how the changes of this PR should be incorporated in #261. Some approaches are more elegant than others. I _might_ make this a non-composite code transformer, since we can instead rely on the existing `CompositeCodeTransformer`.

rickie:
Current changes LGTM. The test setup is nice 🚀! 

Thanks for the heads up, let's see 👀 😄.

refaster-support/src/main/java/tech/picnic/errorprone/refaster/AnnotatedCompositeCodeTransformer.java
Stephan202: I added a commit for this. A "smaller" `XXX` comment remains, but that's for another day.

Badbond (approved):
PR Looking very good. Added two tiny commits and got one observation I'm happy to continue without.

refaster-support/src/test/java/tech/picnic/errorprone/refaster/ErrorProneForkTest.java
```diff
@@ -0,0 +1,66 @@
...
+            "A.java",
+            "// BUG: Diagnostic contains: Suggestions as warnings enabled: false",
+            "class A {}")
+        .doTest();
+  }
+
+  @Test
+  void isSuggestionsAsWarningsEnabledWithFlag() {
+    assumeTrue(
+        ErrorProneFork.isErrorProneForkAvailable(), "The Error Prone fork is not on the classpath");
```
Stephan202: Good one @Badbond! I'll add a small commit to emphasize it's _Picnic's_ fork.

refaster-support/src/test/java/tech/picnic/errorprone/refaster/AnnotatedCompositeCodeTransformerTest.java
```diff
@@ -0,0 +1,203 @@
...
+   * Returns a {@link Description} with some default values as produced by {@link
+   * com.google.errorprone.refaster.RefasterScanner}.
+   */
+  private static Description refasterDescription(String name) {
+    return description(name, Optional.of(""), WARNING, "");
+  }
+
+  private static Description description(
+      String name, Optional<String> link, SeverityLevel severityLevel, String message) {
+    return Description.builder(DUMMY_POSITION, name, link.orElse(null), severityLevel, message)
```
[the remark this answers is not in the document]
Stephan202: I wondered the same, but the current approach is consistent with the rest of the code base: the `null` is there only due to an API we don't control :)

Stephan202 (commented):
Tnx for the review @Badbond!

refaster-support/src/test/java/tech/picnic/errorprone/refaster/AnnotatedCompositeCodeTransformerTest.java
```diff
@@ -0,0 +1,203 @@
...
+  private static final TreePath DUMMY_PATH = mock(TreePath.class);
+  private static final String DEFAULT_PACKAGE = "";
+  private static final String CUSTOM_PACKAGE = "com.example";
+  private static final String SIMPLE_CLASS_NAME = "MyRefasterRule";
+
+  private static Stream<Arguments> applyTestCases() {
+    /* { context, packageName, ruleName compositeAnnotations, delegateAnnotations, expectedDescription } */
+    return Stream.of(
+        arguments(
+            context(),
```
rickie, line 54: We pass `context()` for every test case, would it make sense to move this to the test method and instantiate it there?

rickie (approved):
Test setup looks really nice, I'd say lets get this one merged 🎉 !

Got one minor comment, approving regardless.

refaster-support/src/test/java/tech/picnic/errorprone/refaster/AnnotatedCompositeCodeTransformerTest.java
Stephan202: As-is: yes. But when we pick up the `XXX` this parameter is important. Let me move the comment there.

Stephan202: Done.

rickie:
Have meeting now, did already change one file.

refaster-support/src/main/java/tech/picnic/errorprone/refaster/annotation/OnlineDocumentation.java
```diff
@@ -0,0 +1,49 @@
+package tech.picnic.errorprone.refaster.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Signals that a Refaster rule or group of Refaster rule comes with online documentation.
```
Stephan202, line 9:
```suggestion
 * Signals that a Refaster rule or group of Refaster rules comes with online documentation.
```

Stephan202: That said, there are more references to be updated in this PR (and even in this file :eyes:).

rickie:
Changed the base to `rossendrijver/refaster-templates-rules`. 

Added a commit to add the `@OnlineDocumentation` annotations to the Refaster rule collections.

Rebased and resolved conflicts 😄.

Stephan202:
Tweaked suggested commit message (CC @Badbond):
```
Augment `Description`s of Refaster rule matches (#255)

By emitting a website link, if available, and reporting the matching Refaster
rule in the description's message rather than the check name. Additionally, it
is now possible to provide a custom explanatory text, and to assign a custom 
severity to Refaster rules and to override the severity of matches for all
Refaster rules by passing `-Xep:Refaster:[SEVERITY]`.

Violations of Refaster rules defined in this repository now include a link to
the rule's documentation, hosted on https://error-prone.picnic.tech.
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TestNGToAssertJRules.java
```diff
@@ -72,6 +73,7 @@
 // XXX: As-is these rules do not result in a complete migration:
 // - Expressions containing comments are skipped due to a limitation of Refaster.
 // - Assertions inside lambda expressions are also skipped. Unclear why.
+@OnlineDocumentation
```
Badbond, line 76: Assuming these are migration rules, do we wish to keep these enabled by default in the future? Otherwise it might be nice not to add `@OnlineDocumentation` now where we can filter this one out when we start extracting these classes in docgen.

refaster-test-support/src/test/java/tech/picnic/errorprone/refaster/test/MatchInWrongMethodRules.java
```diff
@@ -2,8 +2,10 @@
 
 import com.google.errorprone.refaster.annotation.AfterTemplate;
 import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
 
 /** Refaster rule collection to validate reporting of a match occurring in an unexpected place. */
+@OnlineDocumentation
```
Badbond, line 8: Are we expecting this to be documented on the website? Otherwise we should remove this annotation, I think.

Badbond (commented):
Thanks @rickie for the rebasing and renaming efforts! :+1: 
I am considering not applying `@OnlineDocumentation` on two Refaster rule collections. WDYT?

Badbond:
Slightly tweaked suggested commit message.

refaster-test-support/src/test/java/tech/picnic/errorprone/refaster/test/MatchInWrongMethodRules.java
Stephan202: Nice catch! Will drop.

Stephan202: That said, the script should anyway restrict itself to `*/main/*`, since `refaster-runner/src/test/java/tech/picnic/errorprone/refaster/runner/FooRules.java` does need the documentation, but should be ignored regardless :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TestNGToAssertJRules.java
Stephan202: Hmm, this one is likely to trigger if not explicitly disabled, so perhaps having the link would be nice? (I'm not too worried about breaking the URL later on, TBH.)

Stephan202 (commented):
Added a commit. Commit tweak LGTM; reflowed it.

refaster-runner/src/test/java/tech/picnic/errorprone/refaster/runner/CodeTransformersTest.java
```diff
@@ -12,6 +12,11 @@ final class CodeTransformersTest {
   @Test
```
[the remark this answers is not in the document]
Stephan202: Odd; this comment should have been flushed during the previous review round.

Badbond (approved):
Nothing more from my side! Let's go! :rocket:

refaster-runner/src/test/java/tech/picnic/errorprone/refaster/runner/CodeTransformersTest.java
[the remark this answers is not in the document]
rickie: Ah yeah there was something there. I think I got it wrong during rebasing haha.

rickie:
Due to connection in the metro I think my previous comment didn't get through;

Latest changes LGTM!

Tweaked the suggested commit message as there was some duplication in there 👀. WDYT?

Stephan202:
> I've got one more thing to consider. I think it could be smart to drop the `@OnlineDocumentation` from the `AssertJRules` rule collection. We don't have a test file there, so there is nothing to show, other than the name of the templates.

Works for me :+1:

> Due to connection in the metro I think my previous comment didn't get through;

Ah, so it was not just me. Connectivity was worse today than generally.

> Tweaked the suggested commit message as there was some duplication in there

Tnx; looks like that was my fault.

Badbond:
> I've got one more thing to consider. I think it could be smart to drop the `@OnlineDocumentation` from the `AssertJRules` rule collection. We don't have a test file there, so there is nothing to show, other than the name of the templates.

Added a commit to drop it.

PR #273, opened 2022-10-03
Introduce `{Mono,Flux}OnErrorComplete` Refaster rules

chamil-prabodha:
Add a new refaster template rule that rewrites `.onErrorResume(e -> Mono.empty())` to `.onErrorComplete()`. This is applied to `.onErrorResume(e -> Flux.empty())` as well.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/ReactorTemplates.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/ReactorTemplates.java
@@ -271,20 +271,7 @@ final class ReactorTemplates {
     }
   }
 
-  /** Prefer {@link Flux#onErrorComplete()} over more contrived alternatives */
-  static final class FluxOnErrorComplete<T> {
-    @BeforeTemplate
-    Flux<T> before(Flux<T> flux) {
-      return flux.onErrorResume(e -> Flux.empty());
-    }
-
-    @AfterTemplate
-    Flux<T> after(Flux<T> flux) {
-      return flux.onErrorComplete();
-    }
-  }
-
-  /** Prefer {@link Mono#onErrorComplete()} over more contrived alternatives */
+  /** Prefer {@link Mono#onErrorComplete()} over more contrived alternatives. */
   static final class MonoOnErrorComplete<T> {
     @BeforeTemplate
     Mono<T> before(Mono<T> mono) {
@@ -294,6 +281,19 @@ final class ReactorTemplates {
     @AfterTemplate
     Mono<T> after(Mono<T> mono) {
       return mono.onErrorComplete();
+    }
+  }
+
+  /** Prefer {@link Flux#onErrorComplete()} over more contrived alternatives. */
+  static final class FluxOnErrorComplete<T> {
+    @BeforeTemplate
+    Flux<T> before(Flux<T> flux) {
+      return flux.onErrorResume(e -> Flux.empty());
+    }
+
+    @AfterTemplate
+    Flux<T> after(Flux<T> flux) {
+      return flux.onErrorComplete();
     }
   }
 
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/ReactorTemplates.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/ReactorTemplates.java
@@ -271,6 +271,32 @@ final class ReactorTemplates {
     }
   }
 
+  /** Prefer {@link Flux#onErrorComplete()} over more contrived alternatives */
+  static final class FluxOnErrorComplete<T> {
+    @BeforeTemplate
+    Flux<T> before(Flux<T> flux) {
+      return flux.onErrorResume(e -> Flux.empty());
+    }
+
+    @AfterTemplate
+    Flux<T> after(Flux<T> flux) {
+      return flux.onErrorComplete();
+    }
+  }
+
+  /** Prefer {@link Mono#onErrorComplete()} over more contrived alternatives */
+  static final class MonoOnErrorComplete<T> {
+    @BeforeTemplate
+    Mono<T> before(Mono<T> mono) {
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/ReactorTemplates.java
```diff
@@ -271,6 +271,32 @@ Flux<S> after(Flux<T> flux) {
     }
   }
 
+  /** Prefer {@link Flux#onErrorComplete()} over more contrived alternatives */
+  static final class FluxOnErrorComplete<T> {
```
rickie: Although the added templates are sorted lexicographical, I see that within this class we usually list the `Mono` before the `Flux`, so I'm swapping them around.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/ReactorTemplates.java
```diff
@@ -271,6 +271,32 @@ Flux<S> after(Flux<T> flux) {
     }
   }
 
+  /** Prefer {@link Flux#onErrorComplete()} over more contrived alternatives */
```
rickie:
```suggestion
  /** Prefer {@link Flux#onErrorComplete()} over more contrived alternatives. */
```
Javadoc should end with a dot 😄.,

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refastertemplates/ReactorTemplatesTestInput.java
```diff
@@ -95,6 +95,14 @@ Flux<Number> testFluxCast() {
     return Flux.just(1).map(Number.class::cast);
   }
 
+  Flux<Object> testFluxOnErrorComplete() {
+    return Flux.error(new IllegalStateException()).onErrorResume(e -> Flux.empty());
```
rickie: We can simplify the example and as a result we don't need to use `Flux<Object>`, we can go with `Flux<Integer>`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/ReactorTemplates.java
```diff
@@ -271,6 +271,32 @@ Flux<S> after(Flux<T> flux) {
...
+      return flux.onErrorResume(e -> Flux.empty());
+    }
+
+    @AfterTemplate
+    Flux<T> after(Flux<T> flux) {
+      return flux.onErrorComplete();
+    }
+  }
+
+  /** Prefer {@link Mono#onErrorComplete()} over more contrived alternatives */
```
rickie: I was doubting to make the description more specific. However, it's now the same as the Javadoc of surrounding templates so decided to leave it as is.

rickie (approved):
Congrats on the first EPS PR @chamil-prabodha ! 🚀 

Added a commit and left some comments.

Suggested commit message: 
```
Prefer `{Mono,Flux}#onErrorComplete` over `{Mono,Flux}#onErrorResume(Function)` (#273)
```

Alternative:
```
Prefer `{Mono,Flux}#onErrorComplete` over `onErrorResume` with `{Mono,Flux#empty}` (#273)
```

One thing:
- Usually we sort lexicographical but I feel Mono and Flux are a bit like min and max, which means we order like proposed.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/ReactorTemplates.java
```diff
@@ -271,6 +271,32 @@ Flux<S> after(Flux<T> flux) {
...
+    @BeforeTemplate
+    Flux<T> before(Flux<T> flux) {
+      return flux.onErrorResume(e -> Flux.empty());
+    }
+
+    @AfterTemplate
+    Flux<T> after(Flux<T> flux) {
+      return flux.onErrorComplete();
+    }
+  }
```
Stephan202: There are also templates (such as `EqualityTemplates.Negation`) in which we combine `@BeforeTemplate`s with incompatible types. We _could_ do the same here. It feels kinda hacky, so I think it's good to keep this code as-is, but that does raise questions about the other templates. (Something to revisit once we add a check that validates the "sanity" of a Refaster rule.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/ReactorTemplates.java
```diff
@@ -271,6 +271,32 @@ Flux<S> after(Flux<T> flux) {
...
+    Mono<T> after(Mono<T> mono) {
+      return mono.onErrorComplete();
+    }
+  }
+
+  /** Prefer {@link Flux#onErrorComplete()} over more contrived alternatives. */
+  static final class FluxOnErrorComplete<T> {
+    @BeforeTemplate
+    Flux<T> before(Flux<T> flux) {
+      return flux.onErrorResume(e -> Flux.empty());
```
Stephan202: We can also add a case for `Mono.empty()` here.

Stephan202 (approved):
Tnx for the PR! Rebased and added a commit.

Suggested commit message:
```
Introduce `{Mono,Flux}OnErrorComplete` Refaster rules (#273)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refastertemplates/ReactorTemplates.java
```diff
@@ -301,7 +301,7 @@ Flux<T> after(Flux<T> flux) {
   static final class PublisherProbeEmpty<T> {
     @BeforeTemplate
     PublisherProbe<T> before() {
-      return Refaster.anyOf(PublisherProbe.of(Mono.empty()), PublisherProbe.of(Flux.empty()));
```
rickie, line 278: Aiii, not sure how I misses this 😓.

PR #274, opened 2022-10-03
Unify local and GitHub Actions website generation flow

japborst:
A follow-up from https://github.com/PicnicSupermarket/error-prone-support/pull/253#discussion_r982709523:

A big downside of the standard GitHub actions set up, is that one can not locally run this. Additionally, setting versions in the `gemfile` can quickly go out of date or misaligned compared to the GHA configuration.

Here, we're moving away from the default GH pages configuration and run the jekyll commands ourselves s.t. we can specify all versions in our own `gemfile`.


Stephan202:
Rebased; PR now targets `master`.

website/README.md
```diff
@@ -24,6 +24,26 @@ If you are not familiar with Jekyll, be sure to check out its
 [documentation][jekyll-docs]. It is recommended to follow the provided
 step-by-step tutorial.
 
+**Switch Ruby versions**
```
Stephan202: We can make these "real" headers using `######`.

Stephan202 (approved):
Rebased and added a commit. Very nice quality improvement!

Suggested commit message:
```
Unify local and GitHub Actions website generation flow (#274)
```

website/README.md
```diff
@@ -24,6 +24,26 @@ If you are not familiar with Jekyll, be sure to check out its
...
+```
+
+In that case, run
+
+```sh
+bundle config build.eventmachine --with-cppflags="-I$(brew --prefix openssl)/include"
+bundle install
+```
+
 # Deployment
```
rickie, line 47: Based on the other headers, shouldn't this be `deploying`?

website/README.md
```diff
@@ -24,6 +24,26 @@ If you are not familiar with Jekyll, be sure to check out its
...
+
+###### Resolving Bundler issues
+
+On Macos you may get an error such as the following when `bundle install`:
+
+```sh
+fatal error: 'openssl/ssl.h' file not found
+```
+
+In that case, run
```
rickie, line 40:
```suggestion
In that case, run:
```

website/README.md
```diff
@@ -24,6 +24,26 @@ If you are not familiar with Jekyll, be sure to check out its
...
 step-by-step tutorial.
 
+###### Switching Ruby versions
+
+The required Ruby version is set in `.ruby-version`. To switch, you can use
+[rvm][rvm] to manage your Ruby version.
+
+###### Resolving Bundler issues
+
+On Macos you may get an error such as the following when `bundle install`:
```
rickie, line 34: macOS 😉

rickie (approved):
Added a commit. 

Nice improvements 🎉.

website/README.md
```diff
@@ -24,6 +24,27 @@ If you are not familiar with Jekyll, be sure to check out its
 [documentation][jekyll-docs]. It is recommended to follow the provided
 step-by-step tutorial.
 
+###### Switch Ruby versions
+
+The required Ruby version is set in `.ruby-version`. To switch, you can use
+[rvm][rvm] to manage your Ruby version.
+
+###### Resolve Bundler issues
```
Stephan202, line 32: @japborst I changed the grammar of these headers on purpose: the topic is "switch&ZeroWidthSpace;_ing_ Ruby versions" and "resolve&ZeroWidthSpace;_ing_ Bundler issues". This also matches the format used for the top-level README. Any particular reason this was reverted?

website/README.md
Stephan202: Good one; I think so.

website/README.md
japborst: Reverted to be in line with the rest.

Stephan202: The other headers do not use the imperative form either?

PR #278, opened 2022-10-05
Fix default branch reference in GitHub Actions definition

japborst:
In #253 the GHA workflow was edited to use `$default-branch`. Whilst this works in workflow templates, this actually doesn't in a workflow.

Here, we revert to referencing `master` directly.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,7 +2,8 @@ name: Build and verify
 on:
   pull_request:
   push:
-    branches: [$default-branch]
+    branches:
+      - master
 permissions:
   contents: read
 jobs:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,8 +2,7 @@ name: Build and verify
 on:
   pull_request:
   push:
-    branches:
-      - master
+    branches: [master]
 permissions:
   contents: read
 jobs:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,7 +2,7 @@ name: Build and verify
 on:
   pull_request:
   push:
-    branches: [master]
+    branches: [ master ]
 permissions:
   contents: read
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -2,7 +2,8 @@ name: Build and verify
 on:
   pull_request:
   push:
-    branches: [$default-branch]
+    branches:
+      - master
```
Stephan202, line 6: We could still do:
```suggestion
    branches: [ master ]
```

(Assuming the earlier intention was to match `jdk: [ 11.0.16, 17.0.4 ]` below. Or we could change both to multi-line lists. Maybe that's nicer?)

Stephan202 (approved):
Nice catch! For others: see also [this](https://stackoverflow.com/a/65723433/74939).

Suggested commit message:
```
Fix default branch reference in GitHub Actions workflow definitions (#278)
```

.github/workflows/build.yaml
rickie: IMO the `jdk` field is a bit different as for `matrix` that is the way to define the values, see [here](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs). 

AFAICS that is the only way there. For branches, especially if it is only one, I prefer the `- master` way. 

WDYT?

rickie (approved):
One answer/question.

Already approving as it is not a blocker.

.github/workflows/build.yaml
japborst: The docs also use both approaches, not very consistent. `[something]` is the same as `- something`, so in the end shouldn't matter. It can be a shorthand, which are especially helpful for matrices, whereas branches are longer to begin with.

```
jdk: [ 11.0.16, 17.0.4 ]
```

instead of 

```
jdk:
 - 11.0.16
 - 17.0.4
```

Nevertheless, I don't see us expanding this list to a wide range, so using the shorthand is pretty clean.

rickie:
Okay clear, let's merge when the build is green 😄.

PR #281, opened 2022-10-06
Fix `LexicographicalAnnotationAttributeListing` string sorting

Ptijohn:
Suggested commit message:
```
Fix sorting of annotation nodes (#281)

By ignoring leading and trailing escaped double quotes.
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationAttributeListing.java
```diff
@@ -203,8 +203,10 @@ private ImmutableList<String> tokenize(Tree node) {
         /*
          * Tokens are split on `=` so that e.g. inline Spring property declarations are properly
          * sorted by key, then value.
+         * Leading and trailing double quotes are ignored for sorting purposes.
          */
-        return ImmutableList.copyOf(SourceCode.treeToString(node, state).split("=", -1));
+        return ImmutableList.copyOf(
+            SourceCode.treeToString(node, state).replaceAll("^\"|\"$", "").split("=", -1));
```
Ptijohn, line 209: Not sure if the right fix tbh.
But after loads of testing, I finally got that the escaped quotes were considered in the sorting, which we actually don't want, and we want to sort what's inside the string. 
Thus filtering out the leading and trailing quotes (if exists) from there.

Stephan202: Thanks for tracking down the source of the odd sorting! It looks like we should use `ASTHelpers#constValue` here; I'll push a proposal :)

Stephan202 (approved):
Nice!

```
Fix `LexicographicalAnnotationAttributeListing` string sorting (#281)

In particular, the empty string is now ordered first.
```

Stephan202:
Had a shower thought; added one more commit.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationAttributeListing.java
Ptijohn: Oh, didn't know about this one :D 
Thanks @Stephan202 for pushing those extras!

rickie (approved):
Thanks for picking this up @Ptijohn ! It's nice to have this fixed after quite a while 😄.

PR #284, opened 2022-10-08
Improve build and deployment concurrency handling

Stephan202:
Suggested commit message:
```
Improve build and deployment concurrency handling (#284)

Builds for the same branch are now serialized. Among other things this prevents
concurrent deployments.

While there, reduce the permission assigned to each of the two jobs.
```

### DIFF
--- a/.github/workflows/deploy-website.yaml
+++ b/.github/workflows/deploy-website.yaml
@@ -38,7 +38,6 @@ jobs:
       group: pages
       cancel-in-progress: true
     permissions:
-      contents: read
       id-token: write
       pages: write
     runs-on: ubuntu-22.04
--- a/.github/workflows/deploy-website.yaml
+++ b/.github/workflows/deploy-website.yaml
@@ -3,6 +3,9 @@ on:
   pull_request:
   push:
     branches: [ master, website ]
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.head_ref != null }}
 jobs:
   build:
     permissions:
@@ -34,9 +37,6 @@ jobs:
   deploy:
     if: github.ref == 'refs/heads/website'
     needs: build
-    concurrency:
-      group: pages
-      cancel-in-progress: true
     permissions:
       id-token: write
       pages: write
--- a/.github/workflows/deploy-website.yaml
+++ b/.github/workflows/deploy-website.yaml
@@ -3,10 +3,12 @@ on:
   pull_request:
   push:
     branches: [ master, website ]
+permissions:
+  contents: read
+  id-token: write
+  pages: write
 jobs:
   build:
-    permissions:
-      contents: read
     runs-on: ubuntu-22.04
     steps:
       - name: Check out code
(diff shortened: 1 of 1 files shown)


Stephan202:
The commit history shows my tests to reduce the scope of the `permissions` blocks. The initial attempts failed because I hadn't configured this branch as being allowed to deploy GitHub Pages.

.github/workflows/deploy-website.yaml
```diff
@@ -35,6 +30,12 @@ jobs:
   deploy:
     if: github.ref == 'refs/heads/master'
     needs: build
+    concurrency:
+      group: pages
+      cancel-in-progress: true
```
japborst, line 35: Did you also consider removing this setting completely? What's the downside of deploying twice? 

If I understand correctly, what's happening is:
- PR-1 merged => master is built & deployed
- PR-2 is merged => master is built & deployed

Where PR-2 is somehow faster and thus first PR-2 is deployed, after which PR-1 is deployed and thus leading to an outdated website. Additionally, the current setup is cancelling the _build_ across multiple branches/PRs, which then requires one to manually re-run to fix the build.

Now, we're still grouping globally (`group: pages`), and this works because the `if` is on a single branch, but could still lead to issues across multiple branches I think.

An alternative would be moving this up again, but grouping differently. Arguably, if we merge the two PRs (from the above example), there's no need in building twice. Additionally, if you push 2 commits on a branch, we should be able to cancel them as - again - there's no use in building twice.

```suggestion
    concurrency:
      group: ${{ github.ref }}
      cancel-in-progress: true
```

japborst (commented):
Indeed, an oversight. Thanks for looking into this.
Left some comments.

.github/workflows/deploy-website.yaml
rickie: Agree with this change, as this is undesirable:

> Additionally, the current setup is cancelling the build across multiple branches/PRs, which then requires one to manually re-run to fix the build.

Based on the linked docs, I'd say that this would be even safer (and be prepared on possible future usages of `concurrency` as well):  `group: ${{ github.workflow }}-${{ github.ref }}`.

rickie: I'd say that dropping `cancel-in-progress` would be fine as well.

> Did you also consider removing this setting completely?

However,  I'd say that it is nice if you do multiple commits that `pending` builds will get cancelled:

> When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be pending. Any previously pending job or workflow in the concurrency group will be canceled.

rickie (commented):
Suggested some possible tweaks / asked some questions :).

Stephan202:

Stephan202 (commented):
Updated the PR; sorry for the delay. @japborst @rickie WDYT?

.github/workflows/deploy-website.yaml
Stephan202: Hmm, [this](https://github.com/PicnicSupermarket/error-prone-support/pull/284#discussion_r990669460) was an old comment that I hadn't flushed yet, hence the funny ordering. Anyway, the question is whether we're okay with the PR as-is, or want to add `cancel-in-progress: true` again.

rickie (approved):
Nice discussions! This looks good to me!

> it's a nice feature that each merged commit has a "complete" build status

Completely agree.

rickie:
Rebased. Anything else from your side @japborst  😄?

japborst (approved):
Went through the comments. Makes sense 👍 Thanks!

PR #285, opened 2022-10-08
Have website use theme variables instead of custom SASS overrides

japborst:
Now that https://github.com/just-the-docs/just-the-docs/issues/982 is resolved, we can use the recommended approach instead.

Following https://just-the-docs.github.io/just-the-docs/docs/customization/#custom-schemes this means:
- Create a custom scheme (we're making 2, one for `light` and one for `dark`)
- Set the appropriate variables
- Generate css file for new scheme s.t. we can switch between light and dark

wrt the latter, the file name needs to start with `just-the-docs-`, as this prefix is mandatory.

Whilst at it, move the custom styles to the colour scheme as well, to make sure we have a single place/folder where we're managing styles. Using imports, we make sure we don't duplicate anything.

I've separated the common styles and variables, to mimic the original [_variables.scss](https://github.com/just-the-docs/just-the-docs/blob/main/_sass/support/_variables.scss) file structure.


Stephan202 (approved):
Tested on top of #274; LGTM.

That said: we're still using `v0.4.0.rc2`, while 5he fix is of more recent vintage. So I feel I'm missing something. :thinking:

Suggested commit message:
```
Have website use theme variables instead of custom SASS overrides (#285) 
```

japborst:
Two things:
1. The usage is now better documented, and we're now following the recommended approach.
2. The fix ensures it's extending `light` theme by default. This means if you don't import anything, it'll use the `light` theme. As we need support for both `light` and `dark` theme, we're importing anyway, and thus don't need to rely on a new release.

rickie (approved):
Changes LGTM! Nice that we can clean up like this :). 

Suggested commit message 👍🏻!

PR #286, opened 2022-10-10
Make terminology of "Refaster rule" consistent

rickie:
By renaming many occurrences, files, methods and variables containing "Refaster template".

Decided to leave out the rename of the `*Templates.java`. Thought it would be better to do that in a separate PR.


rickie:
Failure is to be expected: the URL with `refasterrules` doesn't exist _yet_.

rickie:
That can of course be fixed as well. Pushed some more tweaks. Will work on the follow up PR that renames the `*Templates` classes.

japborst:
Note:
There's a catch-22 here. The documentation urls (in the warning/error output) will point to the new "rules" path, but the website - in turn - points towards GitHub (as you see from the error above), which leads to a non-existent url as the change hasn't been comitted.

We should investigate using the branch name in the urls instead of hardcoding `master`, which should work both when deploying but also when building a PR.

rickie:
Pushed a tweak that should fix the build. **Please note**: after merging, this will fail the build. We need to update the URLs directly after. We can do that in a separate commit or in #288.

japborst:
>We should investigate using the branch name in the urls instead of hardcoding master, which should work both when deploying but also when building a PR.

For future reference, this could be a way to solve it:
https://github.com/atomist/docs-sdm/issues/15#issuecomment-485886333

rickie:
I'll port over the changes from #288 as well. The README is not updated accordingly, because of what I explained in my [previous comment](https://github.com/PicnicSupermarket/error-prone-support/pull/286#issuecomment-1274412775).

website/refasterrules.md
```diff
@@ -1,8 +1,8 @@
 ---
 layout: default
-title: Refaster templates
+title: Refaster Rules
```
nathankooij, line 3:
```suggestion
title: Refaster rules
```

If we want to match the previous casing. Also below, if so.

nathankooij (approved):
Originally reviewed the other PR, but with my changes I think we got them all covered. 🔍

website/refasterrules.md
rickie: Hmm, we also have Bug Patterns, so Refaster Rules would be inline with that 🤔. 

(IMO it also looks nicer on the website 😬)

rickie:
Thanks for the review @nathankooij 😄.

rickie:
Should we change the label to `chore` or `documentation`?

website/refasterrules.md
nathankooij: Agreed. Let's keep it like this. :)

nathankooij:
> Should we change the label to `chore` or `documentation`? To clarify, there are so many code changes I choose `improvement`.

I'd say `chore`. We're doing grunt-like work here, nor fixing bugs or introducing new features etc. `improvement` is a bit of a vacuous label to me IMO.

rickie:
Rebased.

website/_config.yml
```diff
@@ -4,7 +4,7 @@ logo: assets/images/favicon.svg
 url: https://error-prone.picnic.tech
 description: >-
   Error Prone extensions: extra bug checkers and a large battery of Refaster
-  templates.
+  rules.
```
Stephan202, line 7: Also updated the Github repo description.

Stephan202 (approved):
Tnx for picking this up!

Added a commit in which I also reflowed some text and renamed `RefasterTemplateModifiers{,Test}`, since that checker also covers class and `@Placeholder` method modifiers.

Stephan202:
Suggested commit message:
```
Consistently prefer "Refaster rule" over "Refaster template" (#286)

Because the former term references a class containing one or more
`@BeforeTemplate` methods, one or more `@Placeholder` methods and an optional
`@AfterTemplate`, while the latter term more narrowly references a single
`@BeforeTemplate` or `@AfterTemplate` method.
```

rickie:
@Stephan202 made a tiny tweak to not start the sentence with `because`. Sounds a little bit better to me, but WDYT?

Stephan202:
Works for me :smile:

PR #287, opened 2022-10-10
Accommodate rapid website development on the `website` branch

rickie:


.github/workflows/deploy-website.yaml
```diff
@@ -2,7 +2,7 @@ name: Update `error-prone.picnic.tech` website content
 on:
   pull_request:
   push:
-    branches: [ master ]
+    branches: [ master, develop ]
```
japborst, line 5: Why do we need `develop`? That branch didn't exist before.

.github/workflows/deploy-website.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
         with:
           path: ./website/_site
   deploy:
-    if: github.ref == 'refs/heads/master'
+    if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/website'
```
japborst, line 38: Also requires a change in our environment settings to allow this branch to deploy.

.github/workflows/deploy-website.yaml
japborst: Below we're mentioning `website`, so perhaps it should be `s/develop/website/`?

rickie: You are right, not sure why I typed `develop` 😅. Pushed the tweak.

japborst: I think we can omit this, as long as you open a draft PR. Would that work? Or is there a strong preference to keep using  a branch?

rickie: For now it would be nice to be able to quickly push something to that branch, such that it'll build. Making draft PRs is a bit more cumbersome. We can revert this after we have everything stable and merged the way we want it. 

For now it'll allow us to iterate quickly.

japborst: In that case, we don't even need to merge this PR. Just push this change on the `website` branch, and it'll work :)

Stephan202:
I would have expected this PR to be just `s/master/website/`, such that merges to the default branch don't revert the work on `website` :)

Assuming that's the intention, a suggested commit message could be:
```
Deploy the website from the `website` branch (#287)
```

rickie:
I understood something slightly different. Anyway, updated the PR.

Stephan202 (approved):

.github/workflows/deploy-website.yaml
```diff
@@ -2,7 +2,7 @@ name: Update `error-prone.picnic.tech` website content
 on:
   pull_request:
   push:
-    branches: [ master, website ]
+    branches: [ website ]
```
Badbond, line 5: IIUC, if we wish to run the `build` job still on master, it should be added to the branches here. This will not overwrite the website as the `deploy` job does not run on master.

```suggestion
    branches: [ master, website ]
```

rickie: Ah that would be nice actually, warns us whenever there is an error there. As long as it doesn't override the website for every new merge :).

Badbond (approved):
Based on my understanding, I assume this'll work as intended. :+1:

japborst (approved):

Stephan202:
@rickie will add one more commit "soon" :eyes:

PR #288, opened 2022-10-10
`s/templates/rules/` and `s/template/rule` for all files and classes

rickie:
❗This PR is on top of #286 ❗ 

This should be the last part of all the changes. Went over all the occurrences of "template{,s}" thrice. Could still be the case that I missed one 👀:smile:.


nathankooij:
> @nathankooij the changes from `RefasterTemplateModifiers.java` are there on purpose. There we specifically talk about _templates_, namely the before- and after-templates.

I realize it's a fine line, but some of the changes are on the rule level (e.g. adding `final` or removing `abstract`). I'll let you be the judge in that case. ;)

> Besides that; in `AssertJRules` I feel we are talking about the actual before-template instead of the rule.

Yes, was in dubio about this one. It's talking about the matching logic in the before template, but that also kind of makes a statement about the matching of the rules as a whole. Can go either way on that one.

nathankooij:
Pushed my changes, dropping one change in the modifiers test.

rickie:
Ah I updated the other branch to do the work there. So ported your commit to #286 , see [`a2182cf` (#286)](https://github.com/PicnicSupermarket/error-prone-support/pull/286/commits/a2182cf008dba0f59cd0f1d8530e11d8798f82e7).

rickie:
Will close this one in favor of #286 .

PR #289, opened 2022-10-10
Rename `Slf4JLogStatementTest` to `Slf4jLogStatementTest`

Badbond:
The test class name is inconsistent with the actual class name. Spotted while generating the documentation website. Decided to split it off given the unforeseeable future of the documentation generation branches. Thought to do it in scope of #288 as we were already renaming files there to make Refaster naming consistent, but this class is unrelated in that regard. 

Suggested commit message:

```
Rename `Slf4JLogStatementTest` to `Slf4jLogStatementTest` (#289)
```


Stephan202 (approved):
Tnx! (The build fails for unrelated reasons, so already approving.)

rickie (approved):

PR #292, opened 2022-10-12
Introduce Guava `Preconditions` Refaster rules

werli:
This set of Guava `Preconditions` Refaster rules is non-exhaustive, but covers the most common use case for rewrites to `Preconditions#check{Argument,State}()`.

I had pointed these four use cases out during code review multiple times by now, so adding a rule already. :+1: 

These templates only affect simple `if` constructs, not `if-else`. I think this should be part of another set of checks that removes the `else`-branch when the `if`-branch's responsibility is only throwing an exception (or returning something).


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PreconditionsRules.java
```diff
@@ -0,0 +1,91 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+import static com.google.errorprone.refaster.ImportPolicy.STATIC_IMPORT_ALWAYS;
+
+import com.google.common.base.Preconditions;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import com.google.errorprone.refaster.annotation.UseImportPolicy;
+
+/** Refaster templates related to expressions dealing with {@link Preconditions}. */
```
Stephan202:
```suggestion
/** Refaster templates related to statements dealing with {@link Preconditions}. */
```
:smile:

Stephan202 (commented):
Very nice! Pushed a commit with some tweaks in line with "how we do it elsewhere", particularly in terms of naming.

The format string variants require special handling and are perhaps better deferred to another ticket (though I have some ideas around that; can write some stuff about that after the meetings I have now).

I would suggest also adding the other simple cases, though; don't mind doing those later :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PreconditionsRules.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /** Prefer {@link Preconditions#checkArgument(boolean, Object)} over more verbose alternatives. */
+  static final class CheckArgumentWithMessage {
+    @BeforeTemplate
+    void before(boolean condition, String message) {
+      if (condition) {
+        throw new IllegalArgumentException(message);
+      }
+    }
```
werli, line 39: Opening a thread w/ @Stephan202's message:

> The format string variants require special handling and are perhaps better deferred to another ticket (though I have some ideas around that; can write some stuff about that after the meetings I have now).

I also thought of leaving many of those as follow up. :+1: 

One thought on this however. Wouldn't e.g. a `String#format()` call already partially be handled by the case I had pointed out here? As a result, we'd get
```java
checkArgument(!condition, String.format(message, obj1));
```
which I'd then expect to be re-written by _another_ `Preconditions` rule. Curious to hear the thinking.

Stephan202: Indeed! The philosophy of the Refaster rules defined in this repository are that they define the smallest/most generic possible improvement, so that in unison they improve the largest amount of code. So what you propose is the way to go.

There's one caveat here: not all `String.format` calls can be replaced. What I suspect we need is a `Matcher` that identifies valid [`Strings.lenientFormat`](https://guava.dev/releases/31.1-jre/api/docs/com/google/common/base/Strings.html#lenientFormat(java.lang.String,java.lang.Object...)) format strings; then we can instruct Refaster to rewrite only eligible expressions using `@Matches(IsLenientFormatTemplateString.class)`. (Best name TBD.)

Stephan202 (commented):
Added another commit.

werli:
Latest commits LGTM @Stephan202 :+1:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssortedRules.java
```diff
@@ -49,6 +49,26 @@ int after(int index, int size) {
...
 
+  /**
+   * Prefer {@link Objects#checkIndex(int, int)} over less descriptive or more verbose alternatives.
+   *
+   * <p>If a custom error message is desired, consider using Guava's {@link
+   * com.google.common.base.Preconditions#checkElementIndex(int, int, String)}.
+   */
+  static final class CheckIndexConditional {
+    @BeforeTemplate
+    void before2(int index, int size) {
```
werli: Nice to have added this one, thanks! We can however rename it to `before`. :ok_hand:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PreconditionsRules.java
Stephan202: ^ It'll be more "scalable" to handle the `String.format` logic as a `BugChecker`, so let's defer that to a separate PR.

Stephan202 (approved):
Rebased and added one more commit.

Suggested commit message:
```
Introduce Guava `Preconditions` Refaster rules (#292)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PreconditionsRules.java
```diff
@@ -0,0 +1,174 @@
...
+import static com.google.common.base.Preconditions.checkState;
+import static com.google.errorprone.refaster.ImportPolicy.STATIC_IMPORT_ALWAYS;
+
+import com.google.common.base.Preconditions;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import com.google.errorprone.refaster.annotation.UseImportPolicy;
+
+/** Refaster templates related to statements dealing with {@link Preconditions}. */
+final class PreconditionsRules {
```
Stephan202, line 18: One more thing:
```suggestion
@OnlineDocumentation
final class PreconditionsRules {
```

Stephan202 (commented):
Added :)

rickie (approved):
Added a commit. 

Nice improvements @werli 🚀 !

Suggested commit message LGTM, will merge soon.

rickie:
Hello, yes for `checkIndex()` we have [two rules](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssortedRules.java#L42-L74) in `AssortedRules.java`. 

Up until now we haven't yet considered `checkFromIndexSize` or `checkFromIndex`. Adding rules for that could be interesting though. Are you willing to [file a ticket](https://github.com/PicnicSupermarket/error-prone-support/issues/new/choose) for that? If you could provide some examples there as well, we can discuss adding such a rule. Maybe you would be up for opening a PR for that 😄?

Stephan202:
Something I just realized (and should have considered earlier): in a small number of cases the rules introduced here can cause functional or performance issues. See google/guava#5927 for an example of the former. An example of the latter would be a positional argument that is expensive to compute; such a computation now happens eagerly/unconditionally.

It might be good to test these changes internally before the next tag. (There's no nice way to add a local `@SuppressWarnings` for this case; perhaps we should look into a comment-based do-not-replace marker :thinking:.)

PR #293, opened 2022-10-12
Introduce `{Mono,Flux}#zipWith{,Iterable}` Refaster rules

werli:
These proposed Refaster rules are solely meant to increase the readability when zipping publishers.

In the PR description I'll explain the proposal for `Mono#zipWith()`, but the same applies for the other proposed rules.

---

For example, `Mono#zipWith()` usages make it look like both `Mono`s are not retrieved concurrently, but have a sequential dependency like usages of `Mono#zipWhen()` require.

The following example illustrates the limited readability of `Mono#zipWith()` usages:
```java
Mono<String> zipWithExample(String id) {
  return aService.getA(id).zipWith(bService.getB(id), this::combineAsString);
}
```
Reactive publishers should be "streams of data", but in this example it's not apparent that a tuple is created and ultimately mapped to a string.

This PR suggests to rewrite such examples to the following:
```java
Mono<String> zipExample(String id) {
  return Mono.zip(aService.getA(id), bService.getB(id))
    .map(this::combineAsString);
}
```
As a result, the following this should be more apparent:
1. Both `Mono`s are resolved concurrently to a tuple.
2. There is a non-reactive mapping step.

---

Suggested commit message:
```
Introduce `{Mono,Flux}#zipWith()` Refaster rules (#293)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -106,6 +106,53 @@ final class ReactorRules {
     }
   }
 
+  /**
+   * Prefer {@link Flux#zip(Publisher, Publisher)} over a chained {@link Flux#zipWith(Publisher)}.
+   */
+  static final class FluxZip<T, S> {
+    @BeforeTemplate
+    Flux<Tuple2<T, S>> before(Flux<T> flux, Publisher<S> other) {
+      return flux.zipWith(other);
+    }
+
+    @AfterTemplate
+    Flux<Tuple2<T, S>> after(Flux<T> flux, Publisher<S> other) {
+      return Flux.zip(flux, other);
+    }
+  }
+
+  /**
+   * Prefer {@link Flux#zip(Publisher, Publisher)} with a chained combinator over a chained {@link
+   * Flux#zipWith(Publisher, BiFunction)}.
+   */
+  static final class FluxZipWithCombinator<T, S, R> {
+    @BeforeTemplate
+    Flux<R> before(Flux<T> flux, Publisher<S> other, BiFunction<T, S, R> combinator) {
+      return flux.zipWith(other, combinator);
+    }
+
+    @AfterTemplate
+    Flux<R> after(Flux<T> flux, Publisher<S> other, BiFunction<T, S, R> combinator) {
+      return Flux.zip(flux, other).map(function(combinator));
+    }
+  }
+
+  /**
+   * Prefer {@link Flux#zipWithIterable(Iterable)} with a chained combinator over {@link
+   * Flux#zipWithIterable(Iterable, BiFunction)}.
+   */
+  static final class FluxZipWithIterable<T, S, R> {
+    @BeforeTemplate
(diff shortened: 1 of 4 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -68,6 +71,82 @@ Mono<T> after(Optional<T> optional) {
...
+  static final class FluxZipWithIterable<T, S, R> {
+    @BeforeTemplate
+    Flux<R> before(Flux<T> flux, Iterable<S> iterable, BiFunction<T, S, R> combinator) {
+      return flux.zipWithIterable(iterable, combinator);
+    }
+
+    @AfterTemplate
+    Flux<R> after(Flux<T> flux, Iterable<S> iterable, BiFunction<T, S, R> combinator) {
+      return flux.zipWithIterable(iterable).map(function(combinator));
+    }
```
werli, line 147: Note that this method is the only out of the added ones that could be annotated to statically import `TupleUtils#function()` without statically importing `{Mono,Flux}#zip()`. For consistency _within these changes_ I didn't annotate this method, but maybe we should.

```suggestion
    @AfterTemplate
    @UseImportPolicy(STATIC_IMPORT_ALWAYS)
    Flux<R> after(Flux<T> flux, Iterable<S> iterable, BiFunction<T, S, R> combinator) {
      return flux.zipWithIterable(iterable).map(function(combinator));
    }
```
Worth pointing out that `TupleUtils#function()` is to be statically imported anyways using EPS since it's part of the `StaticImport` bug checker.

werli (commented):
Some notes / reasoning for reviewers. :+1:

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -35,6 +36,28 @@ ImmutableSet<Mono<Integer>> testMonoFromOptional() {
         Mono.fromSupplier(() -> Optional.of(2).orElse(null)));
   }
 
+  Mono<Tuple2<String, Integer>> testMonoZip() {
+    return Mono.just("foo").zipWith(Mono.just(1));
+  }
+
+  Mono<String> testMonoZipWithCombinator() {
+    return Mono.just("foo").zipWith(Mono.just(2), (string, count) -> string.repeat(count));
```
Stephan202:
```suggestion
    return Mono.just("foo").zipWith(Mono.just(1), String::repeat);
```

Stephan202 (approved):
Thanks! I can see these changes being slightly controversial, but on the whole I agree with them :) Rebased and added a commit.

Suggested commit message:
```
Introduce `{Mono,Flux}#zipWith{,Iterable}` Refaster rules (#293)
```

werli:
^ Rebased to resolve conflicts. :+1:

werli:
^ Rebased to resolve conflicts v2. :+1:

rickie (approved):
Nice PR @werli. Thanks for the thorough explanation and examples. Really helped reviewing the PR 🚀 !

rickie:
Will merge once 🟢 .

PR #301, opened 2022-10-17
Run build workflow for non-LTS JDKs

rickie:


Stephan202 (approved):
Rebased and added a commit based on what we discussed offline (if I recall the outcome exactly, that is), plus some other tweaks. In summary:
1. Only support the latest non-LTS release.
2. Use the latest LTS for non-Linux builds.
3. Use specific macOS/Windows versions, like we do for Ubuntu.
4. Use Zulu only for the EA build (since Temurin doesn't provide one, IIUC).

Suggested commit message:
```
Build and test on additional platforms and against additional JDK versions (#301)
```

rickie:
Thanks for applying the suggestions from our last discussion 😄! Will mark it as ready for review :).

Badbond (approved):
LGTM. :+1: Alternative suggested commit message (not only different JDK version but also a different distribution, length also under 79 chars):

```
Build and test on additional platforms and against additional JDKs (#301)
```

Stephan202:
Merged the base branch into this branch (rebasing would require resolving a bunch of conflicts). Will merge once built.

PR #304, opened 2022-10-18
Make the build JDK 18+ compatible

rickie:
@werli reported that he got some errors when running Error Prone Support with JDK 18.0.2. 
 After some investigation I found these two JDK bug reports:
- https://bugs.openjdk.org/browse/JDK-8249634
- https://bugs.openjdk.org/browse/JDK-8160675

The first one flagged quite some violations. To resolve these violations, the default non-arg constructors are added _with_ Javadoc. This solution is not ideal. However by looking at the [docs of the javadoc](https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html) command, and more specifically the [options](https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDFDACB), I don't see a way to configure the check to go around this... 

For me, this check flags a lot of missing Javadocs during development. Therefore, it'd be a good idea to not disable this check and use this PR to make the code compatible with JDK 18.

```
$ java -version
openjdk version "18.0.2" 2022-07-19
IBM Semeru Runtime Open Edition 18.0.2.0 (build 18.0.2+9)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AmbiguousJsonCreator.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AmbiguousJsonCreator.java
@@ -36,6 +36,9 @@ public final class AmbiguousJsonCreator extends BugChecker implements Annotation
   private static final Matcher<AnnotationTree> IS_JSON_CREATOR_ANNOTATION =
       isType("com.fasterxml.jackson.annotation.JsonCreator");
 
+  /** Instantiates the default {@link AmbiguousJsonCreator}. */
+  public AmbiguousJsonCreator() {}
+
   @Override
   public Description matchAnnotation(AnnotationTree tree, VisitorState state) {
     if (!IS_JSON_CREATOR_ANNOTATION.matches(tree, state)) {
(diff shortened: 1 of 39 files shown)


werli:
I'll get back to this PR later on. I had checked out the branch however and the advertised `mvn clean install` now works using OpenJDK 18. :tada: 

```
$ java --version
openjdk 18.0.2 2022-07-19
OpenJDK Runtime Environment (build 18.0.2+0)
OpenJDK 64-Bit Server VM (build 18.0.2+0, mixed mode)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AmbiguousJsonCreator.java
```diff
@@ -36,6 +36,9 @@ public final class AmbiguousJsonCreator extends BugChecker implements Annotation
   private static final Matcher<AnnotationTree> IS_JSON_CREATOR_ANNOTATION =
       isType("com.fasterxml.jackson.annotation.JsonCreator");
 
+  /** Instantiates the default {@link AmbiguousJsonCreator}. */
```
Stephan202: "The default" hints at there being a non-default instance, while in fact there can be zero or more equivalent instances. How about:  
```suggestion
  /** Instantiates a new {@link AmbiguousJsonCreator} instance. */
```

(The three existing cases should be named "a default", as a contrast with "a customized".)

Stephan202 (approved):
It's a bit boiler-plate-y, but I agree it's nicer to keep the lint check enabled.

Suggested commit message:
```
Make the build JDK 18+ compatible (#304)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AmbiguousJsonCreator.java
rickie: Aha, I see. I used the same sentence that we already used in a few places 😬.

rickie:
Should've marked it ready a bit sooner. Will merge once last part is 🟢 .

PR #306, opened 2022-10-19
Introduce Reactor `Context#empty` Refaster rule

werli:
To ensure usage of the intended API.

---

Suggested commit message:
```
Introduce Reactor `Context#empty` Refaster rule (#306)
```


werli:
^ Rebased and (hopefully) fixed the build. :ok_hand:

rickie:
Added a commit to fix the build. 

If the formatter is removing an import in the output file, we can add it to the `elidedTypesAndStaticImports` to make sure the formatter doesn't want to remove the import. The import is unused, but Refaster doesn't remove the import for us.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -375,6 +377,19 @@ Flux<T> after(Flux<T> flux) {
     }
   }
 
+  /** Prefer {@link reactor.util.context.Context#empty()}} over more verbose alternatives. */
+  static final class ContextEmpty {
+    @BeforeTemplate
+    Context before() {
+      return Context.of(ImmutableMap.of());
```
Stephan202: Users outside of Picnic might be more likely to use `new HashMap<>()`; let's add a case for that.

Stephan202: (Better would be to introduce a `Matcher` for this, but that's something for a separate PR.)

Stephan202 (approved):
Added a commit. Nice contribution!

rickie (approved):
Nice contribution @werli 🚀 !

PR #308, opened 2022-10-23
By default, prevent `BugChecker`s from introducing new dependencies

Stephan202:
As-is, this is especially relevant for users who do not use Guava or the New Relic Agent API, as mentioned [here](https://www.reddit.com/r/java/comments/y2u8tv/comment/is50d1u/?utm_source=share&utm_medium=web2x&context=3). Note that most `BugChecker`s don't need to be modified, since any third-party references in their suggestions must already be present if the check is to trigger at all.

Suggested commit message:
```
By default, prevent `BugChecker`s from introducing new dependencies (#308)
```

We should implement something similar for Refaster, but that will look quite different (and is out of scope for this PR).

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ScheduledTransactionTrace.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ScheduledTransactionTrace.java
@@ -52,7 +52,7 @@ public final class ScheduledTransactionTrace extends BugChecker implements Metho
 
   @Override
   public Description matchMethod(MethodTree tree, VisitorState state) {
-    if (!ThirdPartyLibrary.NEW_RELIC_AGENT_API.canUse(state)
+    if (!ThirdPartyLibrary.NEW_RELIC_AGENT_API.isIntroductionAllowed(state)
         || !IS_SCHEDULED.matches(tree, state)) {
       return Description.NO_MATCH;
     }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ScheduledTransactionTrace.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ScheduledTransactionTrace.java
@@ -26,6 +26,7 @@ import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.AnnotationTree;
 import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
+import tech.picnic.errorprone.bugpatterns.util.ThirdPartyLibrary;
 
 /**
  * A {@link BugChecker} that flags methods with Spring's {@code @Scheduled} annotation that lack New
@@ -51,7 +52,8 @@ public final class ScheduledTransactionTrace extends BugChecker implements Metho
 
   @Override
   public Description matchMethod(MethodTree tree, VisitorState state) {
-    if (!IS_SCHEDULED.matches(tree, state)) {
+    if (!ThirdPartyLibrary.NEW_RELIC_AGENT_API.canUse(state)
+        || !IS_SCHEDULED.matches(tree, state)) {
       return Description.NO_MATCH;
     }
 
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
@@ -189,8 +189,7 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
           "INSTANCE",
           "newBuilder",
           "of",
-          "valueOf",
-          "values");
+          "valueOf");
 
   /** Instantiates a new {@link StaticImport} instance. */
   public StaticImport() {}
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
@@ -102,7 +102,8 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
           "org.springframework.http.MediaType",
           "org.testng.Assert",
           "reactor.function.TupleUtils",
-          "tech.picnic.errorprone.bugpatterns.util.MoreTypes");
+          "tech.picnic.errorprone.bugpatterns.util.MoreTypes",
+          "tech.picnic.errorprone.bugpatterns.util.ThirdPartyLibrary");
 
   /** Type members that should be statically imported. */
   @VisibleForTesting
@@ -189,7 +190,8 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
           "INSTANCE",
           "newBuilder",
           "of",
-          "valueOf");
+          "valueOf",
+          "values");
 
   /** Instantiates a new {@link StaticImport} instance. */
   public StaticImport() {}
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
@@ -102,8 +102,7 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
           "org.springframework.http.MediaType",
           "org.testng.Assert",
           "reactor.function.TupleUtils",
-          "tech.picnic.errorprone.bugpatterns.util.MoreTypes",
-          "tech.picnic.errorprone.bugpatterns.util.ThirdPartyLibrary");
+          "tech.picnic.errorprone.bugpatterns.util.MoreTypes");
 
   /** Type members that should be statically imported. */
   @VisibleForTesting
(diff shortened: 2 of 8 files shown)


rickie (commented):
Added a commit. 

First round of review 👀:smile:.

Stephan202 (commented):
Tnx for the typo fixes :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * Tells whether it is okay to introduce a dependency on this well-known third party library in
+   * the given context.
+   *
+   * @param state The context under consideration.
+   * @return {@code true} iff it is okay to assume or create a dependency on this library.
+   */
+  public boolean canUse(VisitorState state) {
```
Stephan202, line 63: Note to self: ponder a better name. I didn't select `isSupported` because it doesn't seem to jive with `IGNORE_CLASSPATH_COMPAT_FLAG`, but perhaps we should use that name anyway. Or perhaps something else?

oxkitsune (approved):
Cool stuff! Left some questions and opinions :smile: 

Would be interesting to see whether we could make a bug checker that enforces `ThirdPartyLibrary` usage in bug checkers :exploding_head:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java
oxkitsune: I think `canUse` is fine, especially if we plan to add more fine-grained control.
In this case I think `canUse` makes more sense than `isSupported` (or `isAvailable`) as the library might be available on the classpath, but like the XXX mentions new usages could be undesirable.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java
```diff
@@ -0,0 +1,98 @@
...
+    return shouldIgnoreClasspath(state) || isKnownClass(className, state);
+  }
+
+  /**
+   * Attempts to determine whether a class with the given FQCN is on the classpath.
+   *
+   * <p>The {@link VisitorState}'s symbol table is consulted first. If the type has not yet been
+   * loaded, then an attempt is made to do so.
+   */
+  private static boolean isKnownClass(String className, VisitorState state) {
```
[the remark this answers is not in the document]
oxkitsune: I prefer `className` as it's more intuitive than `fqcn` :+1:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java
rickie: Yeah I was also thinking about this. 

`isAllowedToUse` (is allowed might a bit more "accurate" as `shouldIgnoreClasspath` is part of that check, although it is very similar to `canUse`)?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java
```diff
@@ -0,0 +1,99 @@
...
+      return false;
+    }
+  }
+
+  private static boolean shouldIgnoreClasspath(VisitorState state) {
+    return state
+        .errorProneOptions()
+        .getFlags()
+        .getBoolean(IGNORE_CLASSPATH_COMPAT_FLAG)
+        .orElse(Boolean.FALSE);
```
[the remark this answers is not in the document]
rickie: Because the `getBoolean` expects a `Boolean` and otherwise there would be an implicit cast from the boxed `false` variant.

rickie (approved):
Some final notes, PR looks nice!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ScheduledTransactionTrace.java
```diff
@@ -48,7 +49,8 @@ public final class ScheduledTransactionTrace extends BugChecker implements Metho
 
   @Override
   public Description matchMethod(MethodTree tree, VisitorState state) {
-    if (!IS_SCHEDULED.matches(tree, state)) {
+    if (!ThirdPartyLibrary.NEW_RELIC_AGENT_API.canUse(state)
```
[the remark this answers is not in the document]
Stephan202: I see a +1, so will do.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
```diff
@@ -102,7 +102,8 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
           "org.springframework.http.MediaType",
           "org.testng.Assert",
           "reactor.function.TupleUtils",
-          "tech.picnic.errorprone.bugpatterns.util.MoreTypes");
+          "tech.picnic.errorprone.bugpatterns.util.MoreTypes",
+          "tech.picnic.errorprone.bugpatterns.util.ThirdPartyLibrary");
```
Stephan202, line 106: This doesn't cause the enum values to be statically imported. Need to investigate before merging this PR.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
```diff
@@ -189,7 +190,8 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
           "INSTANCE",
           "newBuilder",
           "of",
-          "valueOf");
+          "valueOf",
+          "values");
```
Stephan202, line 194: ... but `ThirdPartyLibrary.values()`, _would_ be statically imported, which is not something we want.

Stephan202 (commented):
Rebased and added a commit. We could also move the `StaticImport` changes and investigation to a separate PR. (If we do want to keep it here, then eventually the suggested commit message should be updated, too.)

rickie:
Note that I mention in [this comment](https://github.com/PicnicSupermarket/error-prone-support/pull/308#discussion_r1003514595) that I want to add it to the `STATIC_IMPORT_EXEMPTED_MEMBERS` because I _don't_ like statically importing this one. Primarily as it would mean loosing some context. 

> ... but ThirdPartyLibrary.values(), would be statically imported, which is not something we want.

That is not true right? It would be prevented in the `isCandidate` method because it sees that `values` is in `STATIC_IMPORT_EXEMPTED_IDENTIFIERS`. 

So I want to propose that I add a commit to move it. Which means that we defer looking into the issue with not importing 🤔. WDYT?

rickie:
I see that we currently only allow to exempt specific _members_ using `STATIC_IMPORT_EXEMPTED_MEMBERS`, which we could do as there are only 4 types in the `enum`. Nevertheless, let's defer it all together and investigate the issue + introduce support for easily adding a class to disallow all members thereof.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java
rickie: Read the Javadocs and goal of the class again.

Have three other options: `isUsageAllowed`, `isReferencingAllowed`, `isReferenceAllowed`. 

To me, the `ThirdPartyLibrary#isIntroductionAllowed` sounds a bit like the question: Is it okay if we introduce this third party library (at all)? The name of the method does not 100% match with the Javadoc of the class, as it mentions "introducing **references**". So I'd say that `isReferencingAllowed` could make the whole intent of the class even clearer (and is for that reason the one I like most). WDYT?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
Stephan202: This should also be reverted.

Stephan202 (commented):
Rebased and added a commit. As discussed offline: will merge once built.

> Note that I mention in [this comment](https://github.com/PicnicSupermarket/error-prone-support/pull/308#discussion_r1003514595) that I want to add it to the `STATIC_IMPORT_EXEMPTED_MEMBERS` because I _don't_ like statically importing this one. Primarily as it would mean loosing some context.

Okay, I misread that (but find it rather surprising; I don't see any issue with statically importing these, as long as it'd be done consistently :shrug:).
 
> > ... but ThirdPartyLibrary.values(), would be statically imported, which is not something we want.
> 
> This is not true right? It would be prevented in the `isCandidate` method because it sees that `values` is in `STATIC_IMPORT_EXEMPTED_IDENTIFIERS`.

Yes, because I [added](https://github.com/PicnicSupermarket/error-prone-support/pull/308/commits/13c862beaa90f40b8d59af949bfd209ab605b10e#diff-e0df674a0d2abd3814f43a388eb9244a2becdb9e634ec98f9081d3e980a94381R194) `values` in the context of this PR after making this observation...

rickie:
> I don't see any issue with statically importing these

I think it is a bit less clear 🤷🏻.

> Yes, because I [added](https://github.com/PicnicSupermarket/error-prone-support/pull/308/commits/13c862beaa90f40b8d59af949bfd209ab605b10e#diff-e0df674a0d2abd3814f43a388eb9244a2becdb9e634ec98f9081d3e980a94381R194) values in the context of this PR after making this observation...

Ahh I missed that change 😂, that's why I didn't revert it haha. Thanks for fixing that 😉.

PR #310, opened 2022-10-24
Don't update `project.build.outputTimestamp` on `mvn versions:set`

Stephan202:
Noticed while testing #309. Suggested commit message:
```
Don't update `project.build.outputTimestamp` on `mvn versions:set` (#310)

As we rely on the value of `git.commit.time` instead.
```


Badbond (approved):
Looking good! Thanks for fixing this :pray: 

Slightly tweaked the suggested commit message to not include the property substitution details of `${}`. Feel free to revert if you do prefer that.

PR #314, opened 2022-10-25
Prefer `Flux#take(long, boolean)` over `Flux#take(long)` to limit upstream generation

eric-picnic:
In Reactor versions prior to 3.5.0, `Flux#take(n)` makes an unbounded request upstream, and is equivalent to `Flux#take(n, false)`. In 3.5.0, the behavior of `Flux#take(n)` will change to that of `Flux#take(n, true)`.

The intent with this Refaster rule is to get the new behavior before upgrading to Reactor 3.5.0.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1,6 +1,7 @@
 package tech.picnic.errorprone.refasterrules;
 
 import static com.google.common.collect.MoreCollectors.toOptional;
+import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
 import static com.google.errorprone.refaster.ImportPolicy.STATIC_IMPORT_ALWAYS;
 import static java.util.function.Function.identity;
 import static org.assertj.core.api.Assertions.assertThat;
@@ -28,7 +29,9 @@ import reactor.core.publisher.Mono;
 import reactor.test.StepVerifier;
 import reactor.test.publisher.PublisherProbe;
 import reactor.util.context.Context;
+import tech.picnic.errorprone.refaster.annotation.Description;
 import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
+import tech.picnic.errorprone.refaster.annotation.Severity;
 import tech.picnic.errorprone.refaster.matchers.ThrowsCheckedException;
 
 /** Refaster rules related to Reactor expressions and statements. */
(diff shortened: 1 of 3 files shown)


eric-picnic:
Suggested commit message:
```
Prefer `Flux#take(n, true)` over `Flux#take(n)` to limit generation
    
In Reactor versions prior to 3.5.0, `Flux#take(n)` makes an unbounded
request upstream, and is equivalent to `Flux#take(n, false)`.
In 3.5.0, the behavior of `Flux#take(n)` will change to that of
`Flux#take(n, true)`.

The intent with this Refaster rule is to get the new behavior
before upgrading to Reactor 3.5.0.
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -144,6 +144,18 @@ Mono<S> after(Mono<T> mono, S object) {
...
 
+  static final class FluxTakeGenerationLimit<T> {
+    @BeforeTemplate
+    Flux<T> before(Flux<T> flux, long n) {
+      return flux.take(n);
+    }
+
+    @AfterTemplate
+    Flux<T> after(Flux<T> flux, long n) {
+      return flux.take(n, /* limitRequest= */ true);
```
rickie, line 155: Gotta love the `self-check` profile, didn't expect this to show up in the after-template.

Fun fact; if we really want to add this comment in the actual output of applying a Refaster rule, we can use `Refaster#emitComment` or `Refaster#emitCommentBefore`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -144,6 +144,18 @@ Mono<S> after(Mono<T> mono, S object) {
     }
   }
 
+  static final class FluxTakeGenerationLimit<T> {
```
rickie, line 147:
```suggestion
  static final class FluxTake<T> {
```

rickie: We use content of the after-template to determine how we should name the Refaster rule. We can simply write `FluxTake` here as we see in the after-template. 

(We are working on a naming algorithm that will be enforced.)

rickie, line 147: Let's add a Javadoc here as well :).

rickie (approved):
Added a commit.

Have some suggestions, but overall, looks really good!

Congrats on your first Error Prone Support contribution 🎉 !

W.r.t. the suggested commit message, need to think about it 👀.

rickie:
I'm thinking about the suggested commit message. I think it'd be better to add that information to the Javadoc as this information is good to have in the project. Otherwise it would get "lost" in the git history. So I'd go for moving it to there and simplify the suggested commit message. WDYT @eric-picnic ?

eric-picnic:
>  I think it'd be better to add that information to the Javadoc as this information is good to have in the project.

I agree. I have moved it from the suggested commit message to the Javadoc of `FluxTake`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -144,6 +144,28 @@ Mono<S> after(Mono<T> mono, S object) {
...
+   * Prefer {@link Flux#take(long, boolean)} over {@link Flux#take(long)}.
+   *
+   * <p>In Reactor versions prior to 3.5.0, `Flux#take(long)` makes an unbounded request upstream,
+   * and is equivalent to `Flux#take(long, false)`. In 3.5.0, the behavior of `Flux#take(long)` will
+   * change to that of `Flux#take(long, true)`.
+   *
+   * <p>The intent with this Refaster rule is to get the new behavior before upgrading to Reactor
+   * 3.5.0.
+   */
+  static final class FluxTake<T> {
```
Stephan202, line 166: This is one of the few (perhaps the first?) Refaster rules that is intentionally not behavior preserving. A good opportunity to use the `@Description` and `@Severity` annotations introduced in #255. :smile:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -144,6 +144,28 @@ Mono<S> after(Mono<T> mono, S object) {
...
+    @BeforeTemplate
+    Flux<T> before(Flux<T> flux, long n) {
+      return flux.take(n);
+    }
+
+    @AfterTemplate
+    Flux<T> after(Flux<T> flux, long n) {
+      return flux.take(n, /* limitRequest= */ true);
```
werli, line 165: IMO, the comment doesn't add much. The description is already clear enough.
```suggestion
      return flux.take(n, true);
```

werli (approved):
Nice! Sweet to see this suggestion I had made quite some time ago make it to EPS! :rocket:

Sweet work @eric-picnic :muscle: Congrats on the first contribution here.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: This is suggested by the `BooleanParameter` check (see comment above) :)

Stephan202 (approved):
Rebased and added a commit. Very nice!

I tweaked the suggested commit message, but feel free to revert.

PR #319, opened 2022-10-27
Introduce JUnit factory method BugChecker

eric-picnic:
The BugChecker flags JUnit factory methods which do not have a comment with the parameter names of the test method, or which do not follow the naming convention of suffixing the test method name with `TestCases`.


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitFactoryMethodDeclaration.java
```diff
@@ -0,0 +1,285 @@
...
+import java.util.Optional;
+import java.util.stream.Stream;
+import tech.picnic.errorprone.bugpatterns.util.JUnit;
+import tech.picnic.errorprone.bugpatterns.util.MoreASTHelpers;
+
+/**
+ * A {@link BugChecker} that flags non-canonical JUnit factory method declarations.
+ *
+ * <p>A canonical JUnit factory method is one which - has the same name as the test method it
+ * provides test cases for, but with a `TestCases` suffix, and - has a comment which connects the
```
rickie, line 54: It makes more sense to list the requirements in an (un)ordered list. As this is quite an opinionated check we could decide to write something like "we consider a canonical". WDYT?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitFactoryMethodDeclaration.java
```diff
@@ -0,0 +1,285 @@
...
+
+    /* Even though we match on the test method, none of the fixes apply to it directly, so we report
+    the fixes separately using `VisitorState::reportMatch`, and return `Description.NO_MATCH`. */
+    fixes.forEach(state::reportMatch);
+    return Description.NO_MATCH;
+  }
+
+  private ImmutableList<Description> getSuggestedFixes(
+      MethodTree tree,
+      VisitorState state,
```
rickie, line 132: We always list the `VisitorState` as last parameter.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitFactoryMethodDeclaration.java
```diff
@@ -0,0 +1,285 @@
...
+    ImmutableList<Description> fixes =
+        getSuggestedFixes(
+            tree,
+            state,
+            methodSourceAnnotation,
+            factoryMethodName.orElseThrow(),
+            Iterables.getOnlyElement(factoryMethods));
+
+    /* Even though we match on the test method, none of the fixes apply to it directly, so we report
+    the fixes separately using `VisitorState::reportMatch`, and return `Description.NO_MATCH`. */
```
rickie, line 125:
```suggestion
    the fixes separately using `VisitorState::reportMatch`. */
```
This is kind of implied :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitFactoryMethodDeclaration.java
rickie: (Also updated in a few other places).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitFactoryMethodDeclaration.java
```diff
@@ -0,0 +1,285 @@
...
+
+    return comments.stream()
+        .map(Tokens.Comment::getText)
+        .anyMatch(comment -> comment.equals(expectedComment));
+  }
+
+  private static ImmutableList<Tokens.Comment> extractReturnStatementComments(
+      VisitorState state, int startPosition, int endPosition) {
+    return state.getOffsetTokens(startPosition, endPosition).stream()
+        .filter(t -> t.kind() == Tokens.TokenKind.RETURN)
```
rickie, line 263:
```suggestion
        .filter(t -> t.kind() == TokenKind.RETURN)
```
We can remove some of the qualifiers :). Also `Tokens.Comment`.

rickie (changes requested):
It's a very interesting check with some questions we (either deferred) or haven't had to ask ourselves _yet_. Additionally, this is quite a Picnic-opinionated check 😄.

To make it easier to review this PR + getting it merged, I think we can do two things:

1. Change the base from this PR to #214 there we have some first changes that we should do either way and make this PR a bit easier. 
2. Secondly, extract the things from `JUnit` into a different PR.  

Added a commit with some suggestions. Didn't really review functionality wise yet but wanted to start the discussion already.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/JUnit.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.AnnotationTree;
+import com.sun.source.tree.AssignmentTree;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodTree;
+import com.sun.tools.javac.code.Type;
+import java.util.Optional;
+import javax.lang.model.type.TypeKind;
+
+/** A set of JUnit-specific helpers for working with the AST. */
```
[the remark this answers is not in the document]
eric-picnic: I'm fine with either option, but I think it makes sense to parallel `JUnitMatchers` here. Let's start with renaming it `MoreJUnitMatchers`, and get back to it when we have a clearer idea of in which order the currently open JUnit-related BugChecker PRs should be merged and what code we want to share between the BugCheckers.

eric-picnic:
I have broken out some of the changes to a separate PR (https://github.com/PicnicSupermarket/error-prone-support/pull/335), and rebased this branch on top of them.

oxkitsune (approved):

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitFactoryMethodDeclaration.java
```diff
@@ -0,0 +1,283 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Stream;
+import tech.picnic.errorprone.bugpatterns.util.MoreASTHelpers;
+import tech.picnic.errorprone.bugpatterns.util.MoreJUnitMatchers;
+
+/**
+ * A {@link BugChecker} that flags non-canonical JUnit factory method declarations.
+ *
+ * <p>At Picnic, we consider a JUnit factory method canonical if it
```
rickie, line 55:
```suggestion
 * <p>At Picnic, we consider a JUnit factory method canonical if it
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitFactoryMethodDeclarationTest.java
```diff
@@ -0,0 +1,280 @@
...
+            "A.java",
+            "import static org.junit.jupiter.params.provider.Arguments.arguments;",
+            "",
+            "import java.util.List;",
+            "import java.util.stream.Stream;",
+            "import org.junit.jupiter.params.ParameterizedTest;",
+            "import org.junit.jupiter.params.provider.Arguments;",
+            "import org.junit.jupiter.params.provider.MethodSource;",
+            "",
+            "class A {",
```
rickie, line 129: In the `identification` test we already nicely show what exact cases are flagged. What we usually do in the `replacement` test is verify the actual rewrite for one or two cases. In this example I _think_ we can get away with using only one example in the `replacement` test that shows all the rewrites. This means we can simplify the test code a lot.

rickie (commented):
Added a commit. 

Simplified some parts of the code. Flushing what I already have to show some progress :). 
Will need more time for this nice check 😄.

rickie:
So went through a big rebase. Not sure what made it harder, but I think because there were a lot of overlapping changes.

Let's see what Pitest has to say and let's focus on this one again.

rickie (commented):
Just flushing some comments I have.

Introducing a new utility (`ConflictDetection`) comes with the requirement to add a test for that as well 😉.

eric-picnic:
@rickie, I have pushed a commit for refining some of the tests to get better mutation testing coverage.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/ConflictDetectionTest.java
```diff
@@ -0,0 +1,62 @@
...
+            "  private void foo2t() {}",
+            "",
+            "  // BUG: Diagnostic contains: [RenameBlockerFlagger] `foo3t` is already statically imported",
+            "  private void foo3() {}",
+            "",
+            "  // BUG: Diagnostic contains: [RenameBlockerFlagger] `int` is not a valid identifier",
+            "  private void in() {}",
+            "}")
+        .addSourceLines(
+            "/staticimport/B.java",
```
rickie:
```suggestion
            "/pkg/B.java",
```
In other classes we use `pkg` for this :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreJUnitMatchers.java
```diff
@@ -78,4 +83,24 @@ private static String toMethodSourceFactoryName(
...
+  /**
+   * Extracts the name of the JUnit factory method from a {@link
+   * org.junit.jupiter.params.provider.MethodSource} annotation.
+   *
+   * @param methodSourceAnnotation The {@link org.junit.jupiter.params.provider.MethodSource}
+   *     annotation to extract a method name from.
+   * @return The name of the factory method referred to in the annotation. Alternatively, {@link
+   *     Optional#empty()} if there is more than one.
+   */
+  public static Optional<String> extractSingleFactoryMethodName(
```
rickie, line 96: We should add a test case for this in the `MoreJUnitMatchersTest`, are you up for that?

rickie (changes requested):
Added a commit and left some comments. Had to first go over the code with style things before diving in deeper. Cool stuff!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitFactoryMethodDeclaration.java
```diff
@@ -0,0 +1,249 @@
...
+            .map(VariableTree::getName)
+            .map(Object::toString)
+            .collect(toImmutableList());
+
+    String expectedComment = parameterNames.stream().collect(joining(", ", "/* { ", " } */"));
+
+    List<? extends StatementTree> statements = factoryMethod.getBody().getStatements();
+
+    Stream<? extends StatementTree> returnStatementsNeedingComment =
+        Streams.mapWithIndex(statements.stream(), IndexedStatement::new)
```
rickie, line 181: Can you elaborate a bit on this part? Why do we need an `IndexedStatement` exactly? What are we trying to achieve here?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitFactoryMethodDeclaration.java
```diff
@@ -0,0 +1,249 @@
...
+
+  private ImmutableList<Description> getReturnStatementCommentFixes(
+      MethodTree testMethod, MethodTree factoryMethod, VisitorState state) {
+    ImmutableList<String> parameterNames =
+        testMethod.getParameters().stream()
+            .map(VariableTree::getName)
+            .map(Object::toString)
+            .collect(toImmutableList());
+
+    String expectedComment = parameterNames.stream().collect(joining(", ", "/* { ", " } */"));
```
rickie, line 176: We can use the `collect` in the statement above directly.

rickie: Maybe extracting to a method would make the intent a bit more clear.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitFactoryMethodDeclaration.java
```diff
@@ -0,0 +1,249 @@
...
+
+    List<? extends StatementTree> statements = factoryMethod.getBody().getStatements();
+
+    Stream<? extends StatementTree> returnStatementsNeedingComment =
```
rickie, line 180: Can there be more than one statement that needs a comment? Don't think so right? If so, it would require an extra test for sure (or I am missing something). Otherwise, we can improve the code a bit I think.

rickie (changes requested):
Some more comments and a commit. PTAL :).

PR #322, opened 2022-10-28
Introduce Refaster rules to streamline `java.time` type creation

CoolTomatos:


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
```diff
@@ -64,8 +65,47 @@ ZoneOffset after() {
...
+  /** Prefer more fluent way of converting an instant to a local date. */
+  static final class InstantToLocalDate {
+    @BeforeTemplate
+    LocalDate before(Instant instant, ZoneOffset zoneOffset) {
+      return LocalDate.ofInstant(instant, zoneOffset);
+    }
+
+    @AfterTemplate
+    LocalDate after(Instant instant, ZoneOffset zoneOffset) {
+      return instant.atOffset(zoneOffset).toLocalDate();
```
Stephan202, line 77: While this is more "OOP-y" than the original, I'm not yet convinced that these rewrites are an improvement:
- The old code better conveys intent (IMHO).
- The old code seems more efficient.

So I wonder: shouldn't we define these rules in the opposite direction?

Stephan202 (commented):
Thanks for the PR, @CoolTomatos! I can you share some further details on why he "fluent" expressions should be preferred? (If helpful, feel free to reach out on Slack to link to an associated internal discussion.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
CoolTomatos: Hi @Stephan202,
The four templates I added here is to extend the existing `InstantAtOffset` (which I renamed).
I don't have strong arguments about one way or another.

Cc @rickie

rickie: I have no strong preference here to be honest. So let's do it the other way around then :).

Stephan202: Doing it the other way around SGTM! (I think that the existing `OffsetDateTime.ofInstant(instant, zoneOffset)` -> `instant.atOffset(zoneOffset)` makes sense, but for the new cases the static factory method seems to be the way to go :+1:.)

CoolTomatos: can do

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
```diff
@@ -64,8 +66,67 @@ ZoneOffset after() {
     }
   }
 
+  static final class InstantToLocalDate {
```
Stephan202: We generally name the templates after the "shape" of the code in the `@AfterTemplate`s (with the idea that eventually we'll automate this, because naming is hard). Will propose something :)

Stephan202 (commented):
Added a commit. I suspect we may need to tweak the `@BeforeTemplate`s a bit, but need to leave the metro now. More later (after a bunch of meetings).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
```diff
@@ -64,6 +66,65 @@ ZoneOffset after() {
     }
   }
 
+  static final class LocalDateOfInstant {
+    @BeforeTemplate
+    LocalDate before(Instant instant, ZoneId zoneId) {
+      return Refaster.anyOf(
+          instant.atZone(zoneId).toLocalDate(),
+          instant.atZone(zoneId).toOffsetDateTime().toLocalDate(),
```
Stephan202: We can have a separate `instant.atZone(zoneId).toOffsetDateTime()` -> `OffsetDateTime.ofInstant(instant, zoneId)` rule :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
```diff
@@ -64,6 +66,65 @@ ZoneOffset after() {
...
+  }
+
+  static final class LocalTimeOfInstant {
+    @BeforeTemplate
+    LocalTime before(Instant instant, ZoneId zoneId) {
+      return Refaster.anyOf(
+          instant.atZone(zoneId).toLocalTime(),
+          instant.atZone(zoneId).toOffsetDateTime().toLocalTime(),
+          LocalDateTime.ofInstant(instant, zoneId).toLocalTime());
+    }
```
Stephan202: One could also go via `OffsetTime`; will add.

Stephan202 (approved):
Rebased and added a commit. Nice set of rules like this!

Suggested commit message:
```
Introduce Refaster rules for canonical `java.time` type creation (#322)
``` 

(Not yet really happy with this suggestion; ideas welcome.)

rickie (approved):
One question, other than that, LGTM! 

Nice PR @CoolTomatos 🚀 !

rickie:
@chamil-prabodha anything from your side :)?

chamil-prabodha (approved):
Looks good to me. I added one small comment in the tests. 
Nice PR! This would definitely help us a lot to clean up the code 😄

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
```diff
@@ -64,7 +66,83 @@ ZoneOffset after() {
...
 
-  /** Prefer {@link Instant#atOffset(ZoneOffset)} over the more verbose alternative. */
+  /** Prefer {@link LocalDate#ofInstant(Instant, ZoneId)} over more indirect alternatives. */
+  static final class LocalDateOfInstant {
+    @BeforeTemplate
+    LocalDate before(Instant instant, ZoneId zoneId) {
+      return Refaster.anyOf(
+          instant.atZone(zoneId).toLocalDate(),
+          OffsetDateTime.ofInstant(instant, zoneId).toLocalDate(),
+          LocalDateTime.ofInstant(instant, zoneId).toLocalDate());
```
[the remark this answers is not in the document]
Stephan202: @rickie works for me! Wanna add that commit? :smile:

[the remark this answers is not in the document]
Stephan202: Actually, added it ;)

Stephan202 (approved):
And rebased. :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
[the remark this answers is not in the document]
rickie: Thanks for already doing so @Stephan202 😄

rickie:
I can only come up with: `to streamline 'java.time' type creation`. 
Without looking at `InstantAtZone`, as it is an exception to the rule (or suggested commit message for that matter), it would make sense to call it something like `... prefer '#ofInstant' method for creating 'java.time' types`. 

(I'm using ' in the examples to not break the ` `)

Anyway, if you don't like the suggestion, I'd suggest to roll with the proposed one.

@CoolTomatos thanks for the extra rebase 😉.

CoolTomatos:
I don't have merge rights so please go ahead with your suggestion captain Rick 😁

Stephan202:
So either:
* ```Introduce Refaster rules for canonical `java.time` type creation (#322)```
* ```Introduce Refaster rules to streamline `java.time` type creation (#322)```

I guess I'm still tending to the first option. Hard to put my finger on it, but it seems to make a more modest claim :thinking:. Regardless: @rickie will let you merge :)

rickie:
> @rickie will let you merge :) 

Hehe it that case I'm going with the second one 👀:smile:.

PR #323, opened 2022-10-29
Introduce `IsInstanceLambdaUsage` check

Ptijohn:
Creating as draft as I'm not super comfortable in the usage of `Expression`s and all the nitty gritty details of that...

I'm not happy in the way I'm actually matching the second case of `isInstance(Object)` :/ 
But at the same time I'm not sure I see a way to do that in a recursive way 🤔

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
@@ -3,8 +3,6 @@ package tech.picnic.errorprone.bugpatterns;
 import static com.google.errorprone.BugPattern.LinkType.CUSTOM;
 import static com.google.errorprone.BugPattern.SeverityLevel.SUGGESTION;
 import static com.google.errorprone.BugPattern.StandardTags.STYLE;
-import static com.sun.source.tree.Tree.Kind.INSTANCE_OF;
-import static com.sun.source.tree.Tree.Kind.LAMBDA_EXPRESSION;
 import static tech.picnic.errorprone.bugpatterns.util.Documentation.BUG_PATTERNS_BASE_URL;
 
 import com.google.auto.service.AutoService;
@@ -16,11 +14,19 @@ import com.google.errorprone.fixes.SuggestedFix;
 import com.google.errorprone.matchers.Description;
 import com.sun.source.tree.InstanceOfTree;
 import com.sun.source.tree.LambdaExpressionTree;
+import com.sun.source.tree.Tree.Kind;
+import tech.picnic.errorprone.bugpatterns.util.SourceCode;
 
-/** A {@link BugChecker} that aligns usages of T.class::isInstance. */
+/**
+ * A {@link BugChecker} that flags lambda expressions that can be replaced with a method reference
+ * of the form {@code T.class::isInstance}.
+ *
+ * @see MethodReferenceUsage
+ */
+// XXX: Consider folding this logic into the `MethodReferenceUsage` check.
 @AutoService(BugChecker.class)
 @BugPattern(
-    summary = "Use Class::isInstance where possible",
+    summary = "Prefer `Class::isInstance` method reference over equivalent lambda expression",
     link = BUG_PATTERNS_BASE_URL + "IsInstanceUsage",
     linkType = CUSTOM,
     severity = SUGGESTION,
@@ -33,19 +39,15 @@ public final class IsInstanceUsage extends BugChecker implements LambdaExpressio
 
   @Override
   public Description matchLambdaExpression(LambdaExpressionTree tree, VisitorState state) {
-    if (LAMBDA_EXPRESSION == tree.getKind() && INSTANCE_OF == tree.getBody().getKind()) {
-      return constructDescription(
-          tree, constructFix(tree, ((InstanceOfTree) tree.getBody()).getType()));
+    if (tree.getKind() != Kind.LAMBDA_EXPRESSION || tree.getBody().getKind() != Kind.INSTANCE_OF) {
+      return Description.NO_MATCH;
     }
-    return Description.NO_MATCH;
-  }
 
-  private Description constructDescription(
-      LambdaExpressionTree tree, SuggestedFix.Builder fixBuilder) {
-    return describeMatch(tree, fixBuilder.build());
-  }
-
-  private static SuggestedFix.Builder constructFix(LambdaExpressionTree tree, Object target) {
-    return SuggestedFix.builder().replace(tree, target + ".class::isInstance");
+    return describeMatch(
+        tree,
+        SuggestedFix.replace(
+            tree,
+            SourceCode.treeToString(((InstanceOfTree) tree.getBody()).getType(), state)
+                + ".class::isInstance"));
   }
 }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
@@ -1,0 +1,87 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import static com.google.errorprone.BugPattern.LinkType.CUSTOM;
+import static com.google.errorprone.BugPattern.SeverityLevel.SUGGESTION;
+import static com.google.errorprone.BugPattern.StandardTags.STYLE;
+import static com.sun.source.tree.Tree.Kind.IDENTIFIER;
+import static com.sun.source.tree.Tree.Kind.INSTANCE_OF;
+import static com.sun.source.tree.Tree.Kind.LAMBDA_EXPRESSION;
+import static com.sun.source.tree.Tree.Kind.MEMBER_SELECT;
+import static com.sun.source.tree.Tree.Kind.METHOD_INVOCATION;
+import static tech.picnic.errorprone.bugpatterns.util.Documentation.BUG_PATTERNS_BASE_URL;
+
+import com.google.auto.service.AutoService;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.LambdaExpressionTreeMatcher;
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.fixes.SuggestedFixes;
+import com.google.errorprone.matchers.Description;
+import com.sun.source.tree.IdentifierTree;
+import com.sun.source.tree.InstanceOfTree;
+import com.sun.source.tree.LambdaExpressionTree;
+import com.sun.source.tree.MemberSelectTree;
+import com.sun.source.tree.MethodInvocationTree;
+
+/** A {@link BugChecker} that aligns usages of T.class::isInstance. */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Use Class::isInstance where possible",
+    link = BUG_PATTERNS_BASE_URL + "IsInstanceUsage",
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = STYLE)
+public final class IsInstanceUsage extends BugChecker implements LambdaExpressionTreeMatcher {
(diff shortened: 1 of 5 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
```diff
@@ -0,0 +1,87 @@
...
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.fixes.SuggestedFixes;
+import com.google.errorprone.matchers.Description;
+import com.sun.source.tree.IdentifierTree;
+import com.sun.source.tree.InstanceOfTree;
+import com.sun.source.tree.LambdaExpressionTree;
+import com.sun.source.tree.MemberSelectTree;
+import com.sun.source.tree.MethodInvocationTree;
+
+/** A {@link BugChecker} that aligns usages of T.class::isInstance. */
```
Ptijohn, line 29: Can't `@link` this one 🤔 (AFAICS)
And I don't want to do `@link Class#isInstance(Object)` as this is precisely one of the ones that we're getting rid of.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Override
+  public Description matchLambdaExpression(LambdaExpressionTree tree, VisitorState state) {
+    if (ImmutableSet.of(LAMBDA_EXPRESSION, METHOD_INVOCATION).contains(tree.getKind())) {
+      if (INSTANCE_OF == tree.getBody().getKind()) {
+        return constructDescription(
+            tree, state, constructFix(tree, ((InstanceOfTree) tree.getBody()).getType()));
+      } else if (METHOD_INVOCATION == tree.getBody().getKind()) {
+        return treatMethodInvocation(tree, (MethodInvocationTree) tree.getBody(), state);
+      }
```
Stephan202, line 51: IIUC the `i -> Integer.class.isInstance(i)` case will be flagged and fixed by [MethodReferenceUsage](https://github.com/PicnicSupermarket/error-prone-support/blob/8fa3ff3702f48d77a23bb12401d1ac4a4993f71a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MethodReferenceUsage.java), so it seems we can simplify this check by dropping support for that case :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+    MemberSelectTree methodSelect = (MemberSelectTree) methodInvocationTree.getMethodSelect();
+    if (MEMBER_SELECT != methodSelect.getExpression().getKind()
+        || IDENTIFIER
+            != ((MemberSelectTree) methodSelect.getExpression()).getExpression().getKind()) {
+      return Description.NO_MATCH;
+    }
+    IdentifierTree identifierTree =
+        (IdentifierTree) ((MemberSelectTree) methodSelect.getExpression()).getExpression();
+    return constructDescription(tree, state, constructFix(tree, identifierTree.getName()));
```
Ptijohn, line 72: That's the part I'm quite unhappy about... Looks super specific :/

Stephan202 (commented):
Tnx for filing the PR we discussed! I'm currently reviewing something else; will have a look at this one later :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
Ptijohn: Oh, that would be awesome :D 
Let me try to extend the test coverage of MethodReferenceUsage then ;)

Stephan202: If there is no suitable/equivalent test yet (didn't check): cool!

Ptijohn: Yep exactly :)

Ptijohn: Added this test to `MethodReferenceUsage`:
```
            "    Stream.of(1).filter(v -> Integer.class.isInstance(v));",`
```
And it was actually not flagged 🤔 
I'm wondering if it's because of the `.class.isInstance` 🤔 
But then it indeed feels like it should be `MethodReferenceUsage` that should be extended there.

Stephan202: > But then it indeed feels like it should be `MethodReferenceUsage` that should be extended there.

Indeed! There's [a branch](https://github.com/PicnicSupermarket/error-prone-support/compare/sschroevers/MethodReferenceUsageCheck-improvements) with other improvements for that check, so _perhaps_ it works there. If not, I'd add an `XXX` comment to that effect; it's very tricky code, so I caution against going down that rabbit hole. (Though if you want to... I won't stop you ;).)

Ptijohn: Hahaha, yeah I actually got inspired from this code, and got slightly afraid 😅 
I'll leave that out for now.
That being said, I'll check in the branch 👌 
And then adapt this code.
Thanks for the quick feedback!

Ptijohn: Not covered in the new branch, will drop an XXX on `MethodReferenceUsage` and adapt the rest!

Ptijohn: Done, PTAL :)

Stephan202: Tnx! On the review list it goes :smile:

Ptijohn: Thanks :D

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
```diff
@@ -0,0 +1,57 @@
...
+          tree, state, constructFix(tree, ((InstanceOfTree) tree.getBody()).getType()));
+    }
+    return Description.NO_MATCH;
+  }
+
+  private Description constructDescription(
+      LambdaExpressionTree tree, VisitorState state, SuggestedFix.Builder fixBuilder) {
+    SuggestedFix fix = fixBuilder.build();
+    return SuggestedFixes.compilesWithFix(
+            fix, state, ImmutableList.of(), /* onlyInSameCompilationUnit= */ true)
```
Stephan202, line 49: Aye, `compilesWithFix` is very convenient, but also _very_ expensive. Unfortunately we'll have to find a way to avoid this. @Ptijohn which issue did you hit without this?

(In case you're wondering: indeed the `MethodReferenceUsage` check also relies on `compilesWithFix`, and this is why it hasn't been enabled yet for the internal code base. It's the primary reason for the work on [this branch](https://github.com/PicnicSupermarket/error-prone-support/compare/sschroevers/MethodReferenceUsageCheck-improvements).)

Ptijohn: To be honest? It's mainly that I'm discovering the tool 😅 (and thus I didn't know about the limitations of this one)
I will dig to see what's best in this case 👌

Ptijohn: Oh so IIUC you just exact that we can just drop it right? 
Will push it :) 
I indeed thought for a sec that the `compilexWithFix` was a mandatory step in the validation of such a mechanism, but makes sense that it's not the case was "simpler" changes like this one.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
```diff
@@ -0,0 +1,51 @@
...
+
+/** A {@link BugChecker} that aligns usages of T.class::isInstance. */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Use Class::isInstance where possible",
+    link = BUG_PATTERNS_BASE_URL + "IsInstanceUsage",
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = STYLE)
+public final class IsInstanceUsage extends BugChecker implements LambdaExpressionTreeMatcher {
```
rickie, line 28: As this is quite specific to lambda's, we could update the name to reflect that. Maybe something like `LambdaIsInstanceUage` or `IsInstanceLambdaUsage`?

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsageTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Test
+  void identification() {
+    compilationTestHelper
+        .addSourceLines(
+            "A.java",
+            "import reactor.core.publisher.Flux;",
+            "",
+            "class A {",
+            "  void m() {",
```
rickie, line 21: It'd be nice to also add a non-Reactor related example.

Will push something.

rickie (commented):
Added a commit.

Left some questions and considerations :).

Cool check @Ptijohn 🚀 !

Suggested commit message (name subject to change):
```
Introduce `IsInstanceUsage` check (#323)
```

Stephan202 (commented):
@rickie since you didn't push yet, let me push what I had (not flushed earlier because I had to leave the train).

rickie:
Will start resolving the conflicts 😄. 
Okay we _almost_ had the exact same changes so basically nothing is left haha.

For me the only things left are the [naming discussion](https://github.com/PicnicSupermarket/error-prone-support/pull/323#discussion_r1013400508) and the [tag discussion](https://github.com/PicnicSupermarket/error-prone-support/pull/323#discussion_r1013388030).

Stephan202:
> Okay we _almost_ had the exact same changes so basically nothing is left haha.

GMTA!

(Will have a look at the open points in a bit.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
```diff
@@ -0,0 +1,51 @@
...
+import com.sun.source.tree.LambdaExpressionTree;
+
+/** A {@link BugChecker} that aligns usages of T.class::isInstance. */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Use Class::isInstance where possible",
+    link = BUG_PATTERNS_BASE_URL + "IsInstanceUsage",
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = STYLE)
```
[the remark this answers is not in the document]
Stephan202: I see that `MethodReferenceUsage` does also use `STYLE`, I think in this case `SIMPLIFICATION` would work too. (In the general case method references may be longer, and are perhaps not always "simpler".)

Stephan202 (approved):
Added a commit and tweaked the suggested commit message :)

rickie (approved):

Ptijohn:
Wow thank you both for the reviews and extra commits @rickie and @Stephan202 💪 
I still need to go fully through them, but I think there's a lot of learning for me there :D 
If you want to merge before that, don't hesitate, I'm quite full today so I don't know when I'll to it 🤔

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceUsage.java
Ptijohn: 🤦

Ptijohn (commented):
Alright, can't approve myself, but got through the feedback, I learned a lot for next time :D 
Thanks! 💪

rickie:
That's the spirit 🚀 😄  @Ptijohn !

PR #326, opened 2022-10-31
Don't sort `@SpringPostgresTestContainer#flywaySchemas`

EnricSala:
Because the first element is treated as a default.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationAttributeListing.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationAttributeListing.java
@@ -64,6 +64,7 @@ public final class LexicographicalAnnotationAttributeListing extends BugChecker
       ImmutableSet.of(
           // XXX: unless JsonPropertyOrder#alphabetic is true...
           "com.fasterxml.jackson.annotation.JsonPropertyOrder#value",
+          "com.picnic.jooq.annotation.SpringPostgresTestContainer#flywaySchemas",
           "io.swagger.annotations.ApiImplicitParams#value",
           "io.swagger.v3.oas.annotations.Parameters#value",
           "javax.xml.bind.annotation.XmlType#propOrder",


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationAttributeListing.java
```diff
@@ -64,6 +64,7 @@ public final class LexicographicalAnnotationAttributeListing extends BugChecker
       ImmutableSet.of(
           // XXX: unless JsonPropertyOrder#alphabetic is true...
           "com.fasterxml.jackson.annotation.JsonPropertyOrder#value",
+          "com.picnic.jooq.annotation.SpringPostgresTestContainer#flywaySchemas",
```
Stephan202, line 67: There's also `#flywayLocations`, which eventually delegates to [org.flywaydb.core.api.configuration.FluentConfiguration.html#locations(java.lang.String...)](https://flywaydb.org/documentation/usage/api/javadoc/org/flywaydb/core/api/configuration/FluentConfiguration.html#locations(java.lang.String...)). Not directly clear from the code, but I _assume_ order also matters for that one. If so, perhaps we should add it?

Stephan202: That said (and sorry for not checking this before replying on Slack): in this case I think we should instead configure the [`-XepOpt:LexicographicalAnnotationAttributeListing:Excludes`](https://github.com/PicnicSupermarket/error-prone-support/blob/9c330981eaff6186383d7a22537017eb72e68f8f/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationAttributeListingTest.java#L18) flag in the internal code base, since this class is not open source.

Stephan202 (commented):
(But thanks for the quick turnaround :smile:.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationAttributeListing.java
EnricSala: Sounds good, let's do that :) Closing the PR.

EnricSala: About `#flywayLocations`, I did check it but concluded order didn't matter. I think Flyway uses it's own sorting logic (i.e. first numbered versioned migrations, then repeatable, etc), but can have a closer look :eyes:

Stephan202: Ah, good point about file names mattering; in my clicking through the code I didn't get to that point, but it makes sense!

PR #327, opened 2022-11-02
Introduce `Optional{Filter,Map}` Refaster rules

werli:
After I had found this construct during code review.

A mixture of `Optional#map()` and `Optional#filter()` is however not supported with these rules. At this point it seems that Refaster won't be enough, I suspect?
```java
optional.stream().map(x -> x).filter(x -> true).find{Any,First}
```

Suggested commit message
```
Introduce `OptionalStream{Filter,Map}` Refaster rules (#327)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
@@ -355,6 +355,41 @@ final class OptionalRules {
     }
   }
 
+  /**
+   * Avoid unnecessary {@link Optional} to {@link Stream} conversion when filtering that ultimately
+   * returns the same result.
+   */
+  static final class OptionalStreamFilter<T> {
+    @BeforeTemplate
+    Optional<T> before(Optional<T> optional, Predicate<? super T> predicate) {
+      return Refaster.anyOf(
+          optional.stream().filter(predicate).findAny(),
+          optional.stream().filter(predicate).findFirst());
+    }
+
+    @AfterTemplate
+    Optional<T> after(Optional<T> optional, Predicate<? super T> predicate) {
+      return optional.filter(predicate);
+    }
+  }
+
+  /**
+   * Avoid unnecessary {@link Optional} to {@link Stream} conversion when mapping that ultimately
+   * returns the same result.
+   */
+  static final class OptionalStreamMap<S, T> {
+    @BeforeTemplate
+    Optional<? extends T> before(Optional<S> optional, Function<? super S, ? extends T> function) {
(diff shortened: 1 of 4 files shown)


werli:
I'll try to circle back to this PR tomorrow to ensure that the build is running through. Local `mvn clean install` ran through. :+1:

rickie:
Yeah it's the `self-check` that is failing, I think there is some overlapping check or rule that rewrites the code. 
One can use this command to run the self-check:
```
mvn clean install -Dverification.warn -Perror-prone-fork -Pnon-maven-central -Pself-check -s settings.xml
```

Stephan202 (approved):
Rebased and added a commit with a proposal. Nice one!

Suggested commit message:
```
Introduce `Optional{Filter,Map}` Refaster rules (#327)
```

werli:
Thanks for the suggestions, @Stephan202. After re-thinking parts of it, I agree with your proposed changes. Let's roll with it. :+1:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -355,6 +355,42 @@ Optional<T> after(Optional<T> optional) {
...
+    Optional<T> after(Optional<T> optional, Predicate<? super T> predicate) {
+      return optional.filter(predicate);
+    }
+  }
+
+  /**
+   * Avoid unnecessary {@link Optional} to {@link Stream} conversion when mapping a value of the
+   * former type.
+   */
+  // XXX: If `StreamMapFirst` also simplifies `.findAny()` expressions, then this rule can be
```
rickie, line 380: IIUC in that case the `self-check` would identify this case right? 

In that case I'd argue that it would be better to drop this XXX 😬. 
@Stephan202

rickie (approved):
Nice @werli, another contribution! 

Looks good 😄. 

Suggested commit message LGTM.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
Stephan202: I agree that we _could_ drop it for that reason, but without it on might ask "why not add `.findFirst()`?" IMHO now it's clearer why this rule exists in the first place :shrug:

rickie: Okay okay. Agreed, it doesn't really fit in a `@Description` or something like that, so XXX is most reasonable.

rickie:
Rebased, will merge once 🟢.

PR #334, opened 2022-11-04
Fixed the link for ErrorProneTestHelperSourceFormat

gtoison:
The link for the ErrorProneTestHelperSourceFormat is broken: https://error-prone.picnic.tech/bugpatterns/ErrorProneTestHelperSourceFormat

It should be: https://error-prone.picnic.tech/bugpatterns/ErrorProneHelperSourceFormat/

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ErrorProneTestHelperSourceFormat.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ErrorProneTestHelperSourceFormat.java
@@ -51,7 +51,7 @@ import java.util.Optional;
 @AutoService(BugChecker.class)
 @BugPattern(
     summary = "Test code should follow the Google Java style",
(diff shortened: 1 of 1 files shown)


rickie:
Nice catch again @gtoison!

It's funny, this is actually an error on our side. We were kind of in a hurry to get the website done before the actual open-sourcing.
So you spotted a very specific bug in the code for the website generation. 
We checked it out and @Badbond _just_ pushed a [fix](https://github.com/PicnicSupermarket/error-prone-support/commit/210db2d473d5a95ba7b60f11c7f8d14a5692b01e) for this and the **correct** URL now works.

Will close this PR as it is now resolved. However, we really want to thank you for filing a PR for this bug 😄. 

(See here: https://error-prone.picnic.tech/bugpatterns/ErrorProneTestHelperSourceFormat/)

rickie:
(Slightly tweaked PR title for the release note generation.)

gtoison:
Thanks for the quick fix!

PR #337, opened 2022-11-05
Improve and extend Refaster `Map` rules

Stephan202:
Suggested commit message:
```
Improve and extend Refaster `Map` rules

Summary of changes:
- Move relevant rules from `AssertJRules` to `AssertJMapRules` and introduce associated tests.
- Extract relevant rules from `AssortedRules` to the new `MapRules` Refaster collection.
- Add a few new rules to both `AssertJMapRules` and `MapRules`.
```

PR #339, opened 2022-11-07
Introduce `MapIsEmpty` Refaster rule

mlrprananta:
Adds an additional refaster Map rule that builts on https://github.com/PicnicSupermarket/error-prone-support/pull/337

Suggested commit message:
```
Introduce `map.keySet().isEmpty()` to `map.isEmpty()` refaster rule
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
@@ -110,7 +110,7 @@ final class MapRules {
   static final class MapIsEmpty<K, V> {
     @BeforeTemplate
     boolean before(Map<K, V> map) {
-      return map.keySet().isEmpty();
+      return Refaster.anyOf(map.keySet(), map.values(), map.entrySet()).isEmpty();
     }
 
     @AfterTemplate
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
@@ -41,6 +41,19 @@ final class MapRules {
     }
   }
 
+  /** Prefer {@link Map#isEmpty()} over more contrived alternatives. */
+  static final class MapIsEmpty<K, V> {
+    @BeforeTemplate
+    boolean before(Map<K, V> map) {
+      return Refaster.anyOf(map.keySet(), map.values(), map.entrySet()).isEmpty();
+    }
+
+    @AfterTemplate
+    boolean after(Map<K, V> map) {
+      return map.isEmpty();
+    }
+  }
+
   /** Prefer {@link Map#size()} over more contrived alternatives. */
   static final class MapSize<K, V> {
     @BeforeTemplate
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
```diff
@@ -105,4 +105,17 @@ Stream<V> after(Map<K, V> map) {
       return map.values().stream();
     }
   }
+
+  /** Prefer {@link Map#isEmpty()} over more contrived alternatives. */
+  static final class MapIsEmpty<K, V> {
+    @BeforeTemplate
+    boolean before(Map<K, V> map) {
+      return map.keySet().isEmpty();
```
Stephan202, line 113: Like above:
```suggestion
      return Refaster.anyOf(map.keySet(), map.values(), map.entrySet()).isEmpty();
```

Stephan202 (commented):
Nice! Will do a more thorough review later :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
mlrprananta: Applied here and to tests ✅

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/MapRulesTestInput.java
```diff
@@ -44,4 +44,15 @@ Stream<String> testMapKeyStream() {
   Stream<Integer> testMapValueStream() {
     return ImmutableMap.of("foo", 1).entrySet().stream().map(Map.Entry::getValue);
   }
+
+  Stream<boolean> testMapIsEmpty() {
```
rickie, line 48: This wasn't flagged because of a tiny line of code missing in the base PR 😬. Will drop this method :).

rickie (approved):
Added a commit.

Suggested commit message:
```
Prefer `Map#isEmpty()` over `Map#{keySet,values,entrySet}.isEmpty()` (#339)
```
(Not sure how to chain method calls with the `#`...)

Alternative would be: 
```
Introduce `MapIsEmpty` Refaster rule (#339)
```

One last thing to consider, we could sort the `keySet`, `values` and `entrySet` lexicographically but I think this order also makes sense 🤔.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/MapRulesTestInput.java
mlrprananta: Thanks, overlooked that

mlrprananta:
Yeah I was thinking of ordering it lexicographically, but I wanted to stick with the order Stephan already proposed in the other rule

Stephan202 (approved):
W.r.t. ordering: it seems we often order by some fuzzy "frequency" metric. :shrug:

Added one commit; don't ask me to _exactly_ explain this ordering; also just following some existing fuzzy logic :grimacing:. (Eventually we'll auto-sort this stuff.)

I think ```Introduce `MapIsEmpty` Refaster rule (#339)``` matches what we've done before :+1:

mlrprananta:
Rebased

PR #346, opened 2022-11-12
Report mutation test coverage of proposed changes

Stephan202:
Suggested commit message:
```
Report mutation test coverage of proposed changes (#346)

Thanks to a free Arcmutate OSS license, GitHub Actions now runs Pitest against
files changed in the context of a PR. 

While there, update Pitest's configuration to ignore Refaster rule collection
classes, as mutations of those classes will not impact the associated unit
tests.

See:
- https://www.arcmutate.com
- https://pitest.org
```

### DIFF
--- a/.github/workflows/pitest-analyze-pr.yml
+++ b/.github/workflows/pitest-analyze-pr.yml
@@ -29,7 +29,7 @@ jobs:
         uses: actions/upload-artifact@v3.1.1
         with:
           name: pitest-reports
-          path: ./target/pit-reports-ci/
+          path: ./target/pit-reports-ci
   # XXX: Drop the job below. Here only for testing purposes.
   update-pr:
     permissions:
@@ -54,4 +54,4 @@ jobs:
           name: pitest-reports
           path: target/pit-reports-ci
       - name: Update PR
-        run: mvn pitest-github:updatePR
+        run: mvn -DrepoToken="${{ secrets.GITHUB_TOKEN }}" pitest-github:updatePR
--- a/.github/workflows/pitest-analyze-pr.yml
+++ b/.github/workflows/pitest-analyze-pr.yml
@@ -35,6 +35,7 @@ jobs:
     permissions:
       actions: read
       checks: write
+      issues: write
     needs: analyze-pr
     runs-on: ubuntu-22.04
     steps:
--- a/.github/workflows/pitest-analyze-pr.yml
+++ b/.github/workflows/pitest-analyze-pr.yml
@@ -34,6 +34,7 @@ jobs:
   update-pr:
     permissions:
       actions: read
+      checks: write
     needs: analyze-pr
     runs-on: ubuntu-22.04
     steps:
--- a/.github/workflows/pitest-analyze-pr.yml
+++ b/.github/workflows/pitest-analyze-pr.yml
@@ -32,6 +32,8 @@ jobs:
           path: ./target/pit-reports-ci/
   # XXX: Drop the job below. Here only for testing purposes.
   update-pr:
+    permissions:
+      actions: read
     needs: analyze-pr
     runs-on: ubuntu-22.04
     steps:
--- a/.github/workflows/pitest-analyze-pr.yml
+++ b/.github/workflows/pitest-analyze-pr.yml
@@ -49,6 +49,5 @@ jobs:
           pr: ${{github.event.pull_request.number}}
           name: pitest-reports
           path: target/pit-reports-ci
-          workflow_conclusion: success
       - name: Update PR
         run: mvn -DrepoToken=${{ secrets.GITHUB_TOKEN }} pitest-github:updatePR
--- a/.github/workflows/pitest-analyze-pr.yml
+++ b/.github/workflows/pitest-analyze-pr.yml
@@ -2,7 +2,7 @@
 # uploads the results. The associated PR is subsequently updated by the
 # `pitest-update-pr.yml` workflow. See https://blog.pitest.org/oss-pitest-pr/
 # for details.
-name: "Mutation testing"
+name: "Mutation testing: analysis"
 on:
   pull_request:
 permissions:
@@ -24,7 +24,7 @@ jobs:
       - name: Run Pitest
         run: mvn test pitest:mutationCoverage -DargLine.xmx=2048m -Dverification.skip -Dfeatures="+GIT(from[HEAD~1]), +gitci"
       - name: Aggregate Pitest reports
-        run: mvn pitest-git:aggregate -DtrailingText="Mutation testing report by [Pitest](https://pitest.org/). Review any surviving mutants by inspecting the line comments under [_Files changed_](${{ github.event.number }}/files)."
+        run: mvn pitest-git:aggregate -DkilledEmoji=":tada:" -DmutantEmoji=":zombie:" -DtrailingText="Mutation testing report by [Pitest](https://pitest.org/). Review any surviving mutants by inspecting the line comments under [_Files changed_](${{ github.event.number }}/files)."
       - name: Upload Pitest reports as artifact
         uses: actions/upload-artifact@v3.1.1
         with:
--- a/.github/workflows/pitest-analyze-pr.yml
+++ b/.github/workflows/pitest-analyze-pr.yml
@@ -22,6 +22,9 @@ jobs:
           distribution: temurin
           cache: maven
       - name: Run Pitest
+        # By running with feature `+GIT(from[HEAD~1]), +gitci`, Pitest only
+        # analyzes changed lines in a PR as GitHub places them in a single
+        # commit. See https://blog.pitest.org/pitest-pr-setup for details.
         run: mvn test pitest:mutationCoverage -DargLine.xmx=2048m -Dverification.skip -Dfeatures="+GIT(from[HEAD~1]), +gitci"
       - name: Aggregate Pitest reports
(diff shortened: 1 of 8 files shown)


.github/workflows/pitest-update-pr.yml
```diff
@@ -0,0 +1,35 @@
+# Updates a pull request based on the corresponding mutation testing analysis
+# performed by the `pitest-analyze-pr.yml` workflow. See
+# https://blog.pitest.org/oss-pitest-pr/ for details.
```
Stephan202, line 3: The changes in this file will only take effect once merged to the default branch. See the code reverted in the most recent commit to see how I tested a variation of this code.

pom.xml
```diff
@@ -1268,8 +1279,12 @@
                     <version>1.9.9</version>
                     <configuration>
                         <excludedClasses>
+                            <!-- AutoValue generated classes. -->
                             <excludedClass>*.AutoValue_*</excludedClass>
+                            <!-- Refaster rule collections. -->
+                            <excludedClass>*.refaster*.*Rules*</excludedClass>
```
Stephan202, line 1285: Arguably this should be moved to a separate PR :shrug:

Stephan202 (commented):
Before reviewing this PR, consider reading the corresponding [blog post](https://blog.pitest.org/oss-pitest-pr/).

.github/workflows/pitest-analyze-pr.yml
```diff
@@ -0,0 +1,32 @@
+# Performs mutation testing analysis on the files changed by a pull request and
+# uploads the results. The associated PR is subsequently updated by the
+# `pitest-update-pr.yml` workflow. See https://blog.pitest.org/oss-pitest-pr/
+# for details.
+name: "Mutation testing"
```
rickie, line 5: To be more similar to the other file we could do `: analysis` here.

.github/workflows/pitest-analyze-pr.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Set up JDK
+        uses: actions/setup-java@v3.6.0
+        with:
+          java-version: 17
+          distribution: temurin
+          cache: maven
+      - name: Run Pitest
+        run: mvn test pitest:mutationCoverage -DargLine.xmx=2048m -Dverification.skip -Dfeatures="+GIT(from[HEAD~1]), +gitci"
+      - name: Aggregate Pitest reports
+        run: mvn pitest-git:aggregate -DtrailingText="Mutation testing report by [Pitest](https://pitest.org/). Review any surviving mutants by inspecting the line comments under [_Files changed_](${{ github.event.number }}/files)."
```
rickie, line 27: Seems fun to me to add some values for `killedEmoji` and `mutantEmoji`. Will propose something 👀. 

Relevant documentation [here](https://docs.arcmutate.com/docs/github/github-maven-plugin.html).

rickie (commented):
Started reviewing this one, cool stuff!

Added a commit.

This might break things, let's see later when testing the changes 👀.

rickie:
Rebased :).

cdg-pitest-license.txt
```diff
@@ -0,0 +1,7 @@
+#License file for pitest extensions
+#Mon Nov 07 08:41:17 GMT 2022
+expires=07/11/2023
+keyVersion=1
+signature=MhZxMbnO6UovNfllM0JuVWkZyvRT3/G5o/uT0Mm36c7200VpZNVu03gTAGivnl9W5RzvZhfpIHccuQ5ctjQkrqhsFSrl4fyqPqu3y5V2fsHIdFXP/G72EGj6Kay9ndLpaEHalqE0bEwxdnHMzEYq5y3O9vUPv8MhUl57xk+rvBo\=
+packages=tech.picnic.errorprone.*
+type=OSSS
```
Stephan202, line 7: @rickie following your last commit the codes doesn't work anymore; the plugin looks for the exact original file. Now all analysis is skipped. From the logs:
```
Licenced for evaluation and demonstration purposes only.
Expires: 2023-09-20
For use with packages: com.example.*
```

.github/workflows/pitest-analyze-pr.yml
```diff
@@ -0,0 +1,32 @@
+# Performs mutation testing analysis on the files changed by a pull request and
+# uploads the results. The associated PR is subsequently updated by the
+# `pitest-update-pr.yml` workflow. See https://blog.pitest.org/oss-pitest-pr/
+# for details.
+name: "Mutation testing: analysis"
```
Stephan202, line 5: I had that `: analysis` suffix in an earlier commit, and I removed it on purpose, IIRC. Something to do with how the result is displayed in the GitHub UI. (But would have to revert the last commit to see how/what.)
```suggestion
name: "Mutation testing"
```

Stephan202 (commented):
I'll let you do the revert and test the other changes :upside_down_face:

cdg-pitest-license.txt
```diff
@@ -0,0 +1,7 @@
+#License file for pitest extensions
```
Stephan202, line 1: To be sure: I'd also revert the change to this file. It doesn't follow our standards anyway (missing dot at end of sentence, auto-generated timestamp), so I'd rather keep it exactly as provided.

Stephan202: Or we should replace all comments with something like:
```
# Arcmutate license for Error Prone Support, requested by sending an email to
# support@arcmutate.com.
```
(Assuming the plugin doesn't care about the comments :eyes:)

pom.xml
rickie: It's okay to leave in IMO.

.github/workflows/pitest-update-pr.yml
```diff
@@ -0,0 +1,35 @@
...
+  update-pr:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3.1.0
+      - name: Set up JDK
+        uses: actions/setup-java@v3.6.0
+        with:
+          java-version: 17
```
rickie: Here we should use the same version as in the existing workflow `17.0.4`, right?

rickie (approved):
Added a commit. Also reverted the test changes. 

The setup looks good to me, very cool to have this as part of the build 😄.

Doubting a little about the suggested commit message, we could provide some more information about Arcmutate and the setup we now have. I'm thinking there is a reason that you kept it high-level, so first asking whether there is a particular reason for it 👀 @Stephan202.

Stephan202:
Latest changes LGTM! Also updated the suggested commit message. PTAL :)

rickie:
Nice, looks splendid 😄!

.github/workflows/pitest-analyze-pr.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          fetch-depth: 2
+      - name: Set up JDK
+        uses: actions/setup-java@v3.6.0
+        with:
+          java-version: 17.0.4
+          distribution: temurin
+          cache: maven
+      - name: Run Pitest
+        run: mvn test pitest:mutationCoverage -DargLine.xmx=2048m -Dverification.skip -Dfeatures="+GIT(from[HEAD~1]), +gitci"
```
Badbond, line 25: At first, the `+GIT(from[HEAD~1])` triggered me thinking this would only apply on the latest commit of the feature branch, and that we instead wish to diff between the head of the target and the head of the feature branch. However, this actually refers to the entire diff as it is squashed into a single commit (see [here](https://blog.pitest.org/pitest-pr-setup/#:~:text=This%20runs%20pitest%2C%20producing%20additional%20output%20which%20is%20then%20used%20by%20the%20maven%20plugin.%20Although%20at%20first%20glance%20it%20looks%20like%20only%20the%20most%20recent%20change%20is%20analysed%2C%20in%20fact%20all%20changes%20in%20the%20PR%20will%20be%20considered%20as%20github%20places%20them%20in%20a%20single%20commit.)):

> This runs pitest, producing additional output which is then used by the maven plugin. Although at first glance it looks like only the most recent change is analysed, in fact all changes in the PR will be considered as github places them in a single commit.

Perhaps worth of a comment to avoid future confusion.

Badbond (approved):
Looking good. Nice suggested commit message too. Didn't dive to deep into all the possbile configuration to see if we can improve further. I've got two comments, but will approve nonetheless as they are not blocking this PR for me.

.github/workflows/pitest-analyze-pr.yml
```diff
@@ -0,0 +1,32 @@
...
+# for details.
+name: "Mutation testing"
+on:
+  pull_request:
+permissions:
+  contents: read
+jobs:
+  analyze-pr:
+    runs-on: ubuntu-22.04
+    steps:
```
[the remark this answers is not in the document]
rickie: Fair point :). 

However, somewhat regularly we have Renovate updates, like for Pitest or its JUnit plugin, where we would want to see that the setup still works with the new version. So I think it's better to leave it as is 😄.

.github/workflows/pitest-analyze-pr.yml
rickie: The blogpost is mentioned in a comment on the second line, so I'm in doubt whether we need a comment on top of that. Both options are fine by me though.

Badbond: What I am linking here is a different blog post though.

Badbond: Added a commit. PTAL.

Stephan202:
^ Will review the latest discussions EOD.

.github/workflows/pitest-analyze-pr.yml
rickie: LGTM! Nice.

Badbond: Very nice. Much more details but still concise. :+1: LGTM.

PR #347, opened 2022-11-13
Improve `Tree` deletion suggestions

Stephan202:
Suggested commit message:
```
Improve `Tree` deletion suggestions (#347)

When suggesting to remove a method or method annotation, also remove any
trailing whitespace. This avoids the possible introduction of an empty
line right at the start of a code block.
```

See also [this comment](https://github.com/PicnicSupermarket/error-prone-support/pull/198#discussion_r1019570033).

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/SourceCode.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/SourceCode.java
@@ -1,7 +1,12 @@
 package tech.picnic.errorprone.bugpatterns.util;
 
+import static com.sun.tools.javac.util.Position.NOPOS;
+
+import com.google.common.base.CharMatcher;
 import com.google.errorprone.VisitorState;
+import com.google.errorprone.fixes.SuggestedFix;
 import com.sun.source.tree.Tree;
+import com.sun.tools.javac.util.JCDiagnostic.DiagnosticPosition;
 
 /**
  * A collection of Error Prone utility methods for dealing with the source code representation of
@@ -9,6 +14,9 @@ import com.sun.source.tree.Tree;
  */
 // XXX: Can we locate this code in a better place? Maybe contribute it upstream?
 public final class SourceCode {
+  /** The complement of {@link CharMatcher#whitespace()}. */
+  private static final CharMatcher NON_WHITESPACE_MATCHER = CharMatcher.whitespace().negate();
+
   private SourceCode() {}
 
   /**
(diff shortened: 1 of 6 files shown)


error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/SourceCodeTest.java
```diff
@@ -0,0 +1,198 @@
...
+            "",
+            "  class MultipleMethods {",
+            "    void middleMethod() {}",
+            "  }",
+            "}")
+        .doTest(TestMode.TEXT_MATCH);
+  }
+
+  /**
+   * Uses {@link SourceCode#deleteWithTrailingWhitespace(Tree, VisitorState)} to suggest the
```
rickie, line 170: For all `@BugPattern`s we start with `A {@link BugChecker} that`, so would suggest to do the same here.

rickie (approved):
Rebased and added a commit.

WDYT about removing the `XXX` at the top of the class of `SourceCode`? I think by now it kinda "fits" in the `util` package together with the other utilities.

Nice and creative way of testing this 🚀 !

Stephan202 (commented):
> Rebased and added a commit.

I don't see a second commit :eyes: 

> WDYT about removing the `XXX` at the top of the class of `SourceCode`? I think by now it kinda "fits" in the `util` package together with the other utilities.

Works for me!

Rebased and added a commit ;)

rickie:
Oops, I don't know what happened but thanks for applying 😉.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/SourceCode.java
```diff
@@ -24,4 +31,31 @@ public static String treeToString(Tree tree, VisitorState state) {
...
+    int endPos = state.getEndPosition(tree);
+    if (sourceCode == null || endPos == NOPOS) {
+      /* We can't identify the trailing whitespace; delete just the tree. */
+      return SuggestedFix.delete(tree);
+    }
+
+    int actualEnd = NON_WHITESPACE_MATCHER.indexIn(sourceCode, endPos);
+    return actualEnd == -1
+        ? SuggestedFix.delete(tree)
+        : SuggestedFix.replace(((DiagnosticPosition) tree).getStartPosition(), actualEnd, "");
```
Stephan202, line 59: Not that I think we ever hit this case, but I think we should use `sourceCode.length()` if `actualEnd == -1`.

Stephan202 (commented):
I'll be afk for the next >24 hours; can add a commit later.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/SourceCode.java
Stephan202: Rebased and added a commit.

ibabiankou (approved):
Looks good 🚀

PR #357, opened 2022-11-18
Have Renovate file `ruby/setup-ruby` upgrade PRs at most once a month

rickie:
As discussed [here](https://github.com/PicnicSupermarket/error-prone-support/pull/345#pullrequestreview-1178125382).

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -10,7 +10,8 @@
     },
     {
       "matchPackagePatterns": [
-        "^com\\.palantir\\.baseline:baseline-error-prone$"
+        "^com\\.palantir\\.baseline:baseline-error-prone$",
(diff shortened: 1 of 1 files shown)


Stephan202 (approved):
Suggested commit message:
```
Have Renovate file `ruby/setup-ruby` upgrade PRs at most once a month (#357)
```

Badbond (approved):
LGTM :+1:. [As mentioned before](https://github.com/PicnicSupermarket/error-prone-support/pull/125#discussion_r896862688), this can cause this dependency to be skipped if the first day of the month falls on a non-working-day resulting in 60+ days of no upgrades.

rickie:
EPS gets Renovate upgrades every day of the week 😉, so we will be safe 😄.

PR #358, opened 2022-11-18
Extend `MonoFlux` Refaster rule

werli:
As I had noticed during code review. :+1: 

Suggested commit message
```
Extend `MonoFlux` Refaster rule (#358)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -559,7 +559,7 @@ final class ReactorRules {
   static final class MonoFlux<T> {
     @BeforeTemplate
     Flux<T> before(Mono<T> mono) {
-      return Flux.concat(mono);
+      return Refaster.anyOf(Flux.concat(mono), mono.flatMapMany(Flux::just));
     }
 
     @AfterTemplate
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -559,7 +559,7 @@ Flux<S> after(Flux<T> flux) {
   static final class MonoFlux<T> {
     @BeforeTemplate
     Flux<T> before(Mono<T> mono) {
-      return Flux.concat(mono);
+      return Refaster.anyOf(Flux.concat(mono), mono.flatMapMany(Flux::just));
```
rickie, line 562: We usually list `Mono`s before `Flux`es, so will change the order :).

rickie (approved):
Rebased and added a commit.

Nice addition @werli 🚀!
Tried to think of other possible variants but couldn't think of them.

Alternative suggested commit message (or we should just use the already proposed one 🤔):
```
Have `MonoFlux` flag `Mono#FlatMapMany(Flux::just)` cases
```

Stephan202 (approved):
Added a commit.

Suggested commit message, in light of further additions:
```
Extend `MonoFlux` Refaster rule (#358)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -559,7 +559,7 @@ Flux<S> after(Flux<T> flux) {
   static final class MonoFlux<T> {
     @BeforeTemplate
     Flux<T> before(Mono<T> mono) {
-      return Flux.concat(mono);
+      return Refaster.anyOf(mono.flatMapMany(Flux::just), Flux.concat(mono));
```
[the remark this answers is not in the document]
werli: Good one. :+1: Thanks!

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
werli: Noted for the future. :memo:

PR #374, opened 2022-11-26
Introduce some Refaster rules that avoid nested `Publisher`s

werli:
### Context

These proposed Refaster rules primarily aim to make Reactor code more readable as nested `Publisher`s are avoided (e.g. `Mono<Mono<String>>`).

They may result in a minor performance improvement due to the dropped `map()` calls.

### Summary

I've noticed an occurrence of the following code pattern:
```java
Mono<String> method1() {
  Mono.just("foo")
    .map(str -> method2(str))
    .flatMap(identity());
}

Mono<String> method2(String str) {
  ...
} 
```
which these changes rewrite to:
```java
Mono<String> method1() {
  Mono.just("foo")
    .flatMap(str -> method2(str));
}
```
Conceptionally, the same applies for:
- `Mono#flatMap(Function)`
- `Mono#flatMapMap(Function)`
- `Flux#concatMap(Function)`
- `Flux#concatMap(Function, int)`

and in theory also for some others (e.g. `Flux#flatMap()`), but these are already re-written by other rules IIUC.

### Suggested commit message
```
Introduce some Refaster rules that avoid nested `Publisher`s (#374)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -639,6 +639,58 @@ final class ReactorRules {
     }
   }
 
+  /** Prefer {@link Mono#flatMap(Function)} over more contrived alternatives. */
+  static final class MonoFlatMapIdentity<S, T> {
+    @BeforeTemplate
+    Mono<T> before(Mono<S> mono, Function<S, Mono<T>> function) {
+      return mono.map(function).flatMap(identity());
+    }
+
+    @AfterTemplate
+    Mono<T> after(Mono<S> mono, Function<S, Mono<T>> function) {
+      return mono.flatMap(function);
+    }
+  }
+
+  /** Prefer {@link Mono#flatMapMany(Function)} over more contrived alternatives. */
+  static final class FlatMapManyIdentity<S, T> {
+    @BeforeTemplate
+    Flux<T> before(Mono<S> mono, Function<S, ? extends Publisher<T>> function) {
+      return mono.map(function).flatMapMany(identity());
+    }
+
+    @AfterTemplate
+    Flux<T> after(Mono<S> mono, Function<S, ? extends Publisher<T>> function) {
+      return mono.flatMapMany(function);
+    }
+  }
+
+  /** Prefer {@link Flux#concatMap(Function)} over more contrived alternatives. */
+  static final class ConcatMapIdentity<S, T> {
+    @BeforeTemplate
+    Flux<T> before(Flux<S> flux, Function<S, ? extends Publisher<T>> function) {
+      return flux.map(function).concatMap(identity());
+    }
+
+    @AfterTemplate
+    Flux<T> after(Flux<S> flux, Function<S, ? extends Publisher<T>> function) {
+      return flux.concatMap(function);
+    }
+  }
+
+  /** Prefer {@link Flux#concatMap(Function, int)} over more contrived alternatives. */
+  static final class ConcatMapIdentityWithPrefetch<S, T> {
+    @BeforeTemplate
+    Flux<T> before(Flux<S> flux, Function<S, ? extends Publisher<T>> function, int prefetch) {
+      return flux.map(function).concatMap(identity(), prefetch);
+    }
+
+    @AfterTemplate
+    Flux<T> after(Flux<S> flux, Function<S, ? extends Publisher<T>> function, int prefetch) {
+      return flux.concatMap(function, prefetch);
+    }
+  }
+
   /**
    * Prefer {@link Flux#concatMapIterable(Function)} over alternatives that require an additional
    * subscription.
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -320,7 +320,10 @@ final class ReactorRules {
   static final class FluxConcatMap<T, S> {
     @BeforeTemplate
     Flux<S> before(Flux<T> flux, Function<? super T, ? extends Publisher<? extends S>> function) {
-      return Refaster.anyOf(flux.flatMap(function, 1), flux.flatMapSequential(function, 1));
+      return Refaster.anyOf(
+          flux.flatMap(function, 1),
+          flux.flatMapSequential(function, 1),
+          flux.map(function).concatMap(identity()));
     }
 
     @AfterTemplate
@@ -337,7 +340,9 @@ final class ReactorRules {
         Function<? super T, ? extends Publisher<? extends S>> function,
         int prefetch) {
       return Refaster.anyOf(
-          flux.flatMap(function, 1, prefetch), flux.flatMapSequential(function, 1, prefetch));
+          flux.flatMap(function, 1, prefetch),
+          flux.flatMapSequential(function, 1, prefetch),
+          flux.map(function).concatMap(identity(), prefetch));
     }
 
     @AfterTemplate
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -615,6 +615,58 @@ Flux<S> after(Flux<T> flux) {
...
+  static final class FlatMapManyIdentity<S, T> {
+    @BeforeTemplate
+    Flux<T> before(Mono<S> mono, Function<S, ? extends Publisher<T>> function) {
+      return mono.map(function).flatMapMany(identity());
+    }
+
+    @AfterTemplate
+    Flux<T> after(Mono<S> mono, Function<S, ? extends Publisher<T>> function) {
+      return mono.flatMapMany(function);
+    }
```
werli, line 641: In the case of `function` being of type `Function<S, Mono<T>>` first this rule and then `MonoFlatMapToFlux` are applied, which ultimately rewrite this code piece
```java
mono.map(function).flatMapMany(identity());
```
to
```java
mono.flatMap(function).flux();
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -615,6 +615,58 @@ Flux<S> after(Flux<T> flux) {
...
+    @AfterTemplate
+    Mono<T> after(Mono<S> mono, Function<S, Mono<T>> function) {
+      return mono.flatMap(function);
+    }
+  }
+
+  /** Prefer {@link Mono#flatMapMany(Function)} over more contrived alternatives. */
+  static final class FlatMapManyIdentity<S, T> {
+    @BeforeTemplate
+    Flux<T> before(Mono<S> mono, Function<S, ? extends Publisher<T>> function) {
```
werli, line 634: Admittedly, I couldn't figure out why
```suggestion
    Flux<T> before(Mono<S> mono, Function<S, Publisher<T>> function) {
```
alone wouldn't do it. Anyone an idea what I'm missing here?

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -615,6 +615,58 @@ Flux<S> after(Flux<T> flux) {
     }
   }
 
+  /** Prefer {@link Mono#flatMap(Function)} over more contrived alternatives. */
+  static final class MonoFlatMapIdentity<S, T> {
```
Stephan202: We generally name the patterns after the `@AfterTemplate`. So this would be `MonoFlatMap`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -615,6 +615,58 @@ Flux<S> after(Flux<T> flux) {
...
+    }
+
+    @AfterTemplate
+    Mono<T> after(Mono<S> mono, Function<S, Mono<T>> function) {
+      return mono.flatMap(function);
+    }
+  }
+
+  /** Prefer {@link Mono#flatMapMany(Function)} over more contrived alternatives. */
+  static final class FlatMapManyIdentity<S, T> {
```
Stephan202: W.r.t. the use of generic type parameters: further up we use `T` and `S` the other way around... perhaps something we should try to "auto-fix" somehow.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: In the latter case only function that return precisely a `Publisher` (so not a subtype) are matched.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -615,6 +615,58 @@ Flux<S> after(Flux<T> flux) {
     }
   }
 
+  /** Prefer {@link Mono#flatMap(Function)} over more contrived alternatives. */
+  static final class MonoFlatMapIdentity<S, T> {
+    @BeforeTemplate
+    Mono<T> before(Mono<S> mono, Function<S, Mono<T>> function) {
```
Stephan202: This matches more, and is consistent with the relevant parameter types:
```suggestion
    Mono<T> before(Mono<S> mono, Function<? super S, ? extends Mono<? extends T>> function) {
```

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -199,6 +200,28 @@ Flux<Number> testFluxCast() {
...
 
+  Mono<String> testMonoFlatMapIdentity() {
+    return Mono.just("foo").map(Mono::just).flatMap(identity());
+  }
+
+  ImmutableSet<Flux<String>> testFlatMapManyIdentity() {
+    return ImmutableSet.of(
+        Mono.just("foo").map(Mono::just).flatMapMany(identity()),
+        Mono.just("foo").map(Flux::just).flatMapMany(identity()));
+  }
```
Stephan202: One test suffices for this case :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -615,6 +615,58 @@ Flux<S> after(Flux<T> flux) {
...
+    @BeforeTemplate
+    Flux<T> before(Flux<S> flux, Function<S, ? extends Publisher<T>> function) {
+      return flux.map(function).concatMap(identity());
+    }
+
+    @AfterTemplate
+    Flux<T> after(Flux<S> flux, Function<S, ? extends Publisher<T>> function) {
+      return flux.concatMap(function);
+    }
+  }
```
Stephan202: This one can be merged with the existing `FluxConcatMap` rule. Likewise for the one below.

Stephan202 (approved):
Added a commit with some suggestions and also tweaked the suggested commit message; PTAL. :)

Tnx!

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
werli: :bulb: Thanks a lot - clear.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
werli: I didn't get this pattern beforehand - notes :memo:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
werli: Thx - I didn't get this apparently.

werli:
Thx @Stephan202 for the good comments. Updated commit message LGTM, that's closer to what I had in mind anyways.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: No worries; we plan to automate this :smile:

rickie (approved):
Nice @werli , you are on a roll! Keep it going 🚀 ! 

Suggested commit message LGTM!

(Rebased, will merge once 🍏)

Issue #336, opened 2022-11-05
Specify `Locale` when calling `String#to{Lower,Upper}Case`


rickie:
### Problem

When calling `String#to{Lower,Upper}Case` without arguments, the system's `Locale` is used by default.
To prevent this we want our code to specify `Locale.ROOT`. 
Therefore, we want to rewrite occurrences that invoke the aforementioned methods without `Locale` and introduce `Locale.ROOT`. 

### Description of the proposed new feature

- [ ] Support a stylistic preference.
- [x] Avoid a common gotcha, or potential problem.

I would like to rewrite the following code:
```java
"A".toLowerCase();
"B".toUpperCase();
```

  to:
```java
"A".toLowerCase(Locale.ROOT);
"B".toUpperCase(Locale.ROOT);
```

### Considerations

Just a consideration; instead of `Locale.ROOT` one can also use other `Locale`s. I _assume_ we want to leave those as is, right? 

### Participation

This is a nice opportunity for contributing an Error Prone check 😄.

Stephan202:
> Just a consideration; instead of `Locale.ROOT` one can also use other `Locale`s. I _assume_ we want to leave those as is, right?

Certainly!

> This is a nice opportunity for contributing a Refaster rule :smile:.

I'd rather see this implemented using a `BugChecker` which suggest the `Locale.ROOT` fix by default, but proposes `Locale.getDefault()` as a secondary fix. The latter is (a) behavior preserving, (b) may be more appropriate in certain contexts (such as desktop applications) and (c) might make for an easier migration in large code bases. (I.e.: first do the safe thing by using `Locale.getDefault()` everywhere, disallow new `.to{Lower,Upper}Case()` calls, then review and replace such usages if/when possible

rickie:
That makes sense, let's create a BugChecker for this then. I updated the description 👍🏻.

mlrprananta:
I'm looking to pick this one up, any pointers before I begin?

rickie:
Yes @mlrprananta for sure.

I would say this will be quite similar to the `FluxFlatMapUsage`, `RefasterAnyOfUsage` and (a little more complex maybe:) the `IdentityConversion` checks. 

On a first glance, it sounds like the `BugChecker` needs to implement the `MethodInvocationTreeMatcher` and use a custom `Matcher` (see the aforementioned checks for some examples) and match on the methods from this ticket. 
[...]

PR #376, opened 2022-11-28
Introduce `StringCaseLocaleUsage` check

mlrprananta:
PR for this [Issue](https://github.com/PicnicSupermarket/error-prone-support/issues/336).

In short, a default Locale should be specified when using either `String#toLowerCase()` or `String#toUpperCase()`.

## Suggested commit message
```
Introduce `StringCaseLocaleUsage` check (#376)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitLocaleUsage.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitLocaleUsage.java
@@ -58,6 +58,10 @@ public final class ImplicitLocaleUsage extends BugChecker implements MethodInvoc
   }
 
   private static Fix suggestLocale(MethodInvocationTree tree, String locale, VisitorState state) {
+    // XXX: The logic that replaces the first parenthesis assumes that `tree` does not have a source
+    // code representation such as `str.toLowerCase/* Some comment with parens (). */()`. In such a
+    // case the comment, rather than the method invocation arguments, will be modified. Implement a
+    // generic solution for this.
     return SuggestedFix.builder()
         .addImport("java.util.Locale")
         .replace(tree, SourceCode.treeToString(tree, state).replaceFirst("\\(", '(' + locale))
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitLocaleUsage.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitLocaleUsage.java
@@ -23,7 +23,10 @@ import tech.picnic.errorprone.bugpatterns.util.SourceCode;
  * A {@link BugChecker} that flags calls to {@link String#toLowerCase()} and {@link
  * String#toUpperCase()}, as these methods implicitly rely on the environment's default locale.
  */
-// XXX: Also flag `String::toLowerCase` and `String::toUpperCase` method references.
+// XXX: Also flag `String::toLowerCase` and `String::toUpperCase` method references. For these cases
+// the suggested fix should introduce a lambda expression with a parameter of which the name does
+// not coincide with the name of an existing variable name. Such functionality should likely be
+// introduced in a utility class.
 @AutoService(BugChecker.class)
 @BugPattern(
     summary = "Specify a `Locale` when calling `String#to{Lower,Upper}Case`",
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitLocaleUsage.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitLocaleUsage.java
@@ -27,11 +27,11 @@ import tech.picnic.errorprone.bugpatterns.util.SourceCode;
 @AutoService(BugChecker.class)
 @BugPattern(
     summary = "Specify a `Locale` when calling `String#to{Lower,Upper}Case`",
-    link = BUG_PATTERNS_BASE_URL + "SpecifyLocale",
+    link = BUG_PATTERNS_BASE_URL + "ImplicitLocaleUsage",
     linkType = CUSTOM,
     severity = WARNING,
     tags = FRAGILE_CODE)
-public final class SpecifyLocale extends BugChecker implements MethodInvocationTreeMatcher {
+public final class ImplicitLocaleUsage extends BugChecker implements MethodInvocationTreeMatcher {
   private static final long serialVersionUID = 1L;
   private static final Matcher<ExpressionTree> DEFAULT_LOCALE_CASE_CONVERSION =
       instanceMethod()
@@ -39,8 +39,8 @@ public final class SpecifyLocale extends BugChecker implements MethodInvocationT
           .namedAnyOf("toLowerCase", "toUpperCase")
           .withNoParameters();
 
-  /** Instantiates a new {@link SpecifyLocale} instance. */
-  public SpecifyLocale() {}
+  /** Instantiates a new {@link ImplicitLocaleUsage} instance. */
+  public ImplicitLocaleUsage() {}
 
   @Override
   public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
@@ -1,7 +1,7 @@
 package tech.picnic.errorprone.bugpatterns;
 
 import static com.google.errorprone.BugPattern.LinkType.CUSTOM;
-import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
+import static com.google.errorprone.BugPattern.SeverityLevel.SUGGESTION;
 import static com.google.errorprone.BugPattern.StandardTags.LIKELY_ERROR;
 import static com.google.errorprone.matchers.method.MethodMatchers.instanceMethod;
 import static tech.picnic.errorprone.bugpatterns.util.Documentation.BUG_PATTERNS_BASE_URL;
@@ -21,19 +21,18 @@ import tech.picnic.errorprone.bugpatterns.util.SourceCode;
 
 /**
  * A {@link BugChecker} that flags {@link String#toLowerCase()} or {@link String#toUpperCase()}
- * which do not specify a Locale.
+ * which do not specify a {@code Locale}.
  */
 @AutoService(BugChecker.class)
 @BugPattern(
-    summary =
-        "Specify `Locale.ROOT` or `Locale.getDefault()` when calling `String#to{Lower,Upper}Case` without a specific Locale",
+    summary = "Specify a `Locale` when calling `String#to{Lower,Upper}Case`",
     link = BUG_PATTERNS_BASE_URL + "SpecifyLocale",
     linkType = CUSTOM,
-    severity = WARNING,
+    severity = SUGGESTION,
     tags = LIKELY_ERROR)
 public final class SpecifyLocale extends BugChecker implements MethodInvocationTreeMatcher {
   private static final long serialVersionUID = 1L;
-  private static final Matcher<ExpressionTree> STRING_TO_UPPER_OR_LOWER_CASE =
+  private static final Matcher<ExpressionTree> STRING_TO_LOWER_OR_UPPER_CASE =
       instanceMethod()
           .onExactClass(String.class.getName())
           .namedAnyOf("toLowerCase", "toUpperCase")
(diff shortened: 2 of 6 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
```diff
@@ -0,0 +1,63 @@
...
+ * which do not specify a Locale.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary =
+        "Specify `Locale.ROOT` or `Locale.getDefault()` when calling `String#to{Lower,Upper}Case` without a specific Locale",
+    link = BUG_PATTERNS_BASE_URL + "SpecifyLocale",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = LIKELY_ERROR)
```
mlrprananta, line 33: Not sure if `LIKELY_ERROR` or `FRAGILE_CODE` is more appropriate here

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
```diff
@@ -0,0 +1,63 @@
...
+      instanceMethod()
+          .onExactClass(String.class.getName())
+          .namedAnyOf("toLowerCase", "toUpperCase");
+
+  /** Instantiates a new {@link SpecifyLocale} instance. */
+  public SpecifyLocale() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (STRING_TO_UPPER_OR_LOWER_CASE.matches(tree, state) && tree.getArguments().isEmpty()) {
```
oxkitsune, line 46:
```suggestion
          .namedAnyOf("toLowerCase", "toUpperCase")
          .withNoParameters();

  /** Instantiates a new {@link SpecifyLocale} instance. */
  public SpecifyLocale() {}

  @Override
  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
    if (STRING_TO_UPPER_OR_LOWER_CASE.matches(tree, state)) {
```

We can use `withNoParameters()` to only match invocations that don't provide any arguments.

mlrprananta: Applied 👍

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/SpecifyLocaleTest.java
```diff
@@ -0,0 +1,93 @@
...
+      CompilationTestHelper.newInstance(SpecifyLocale.class, getClass());
+  private final BugCheckerRefactoringTestHelper refactoringTestHelper =
+      newInstance(SpecifyLocale.class, getClass());
+
+  @Test
+  void identification() {
+    compilationTestHelper
+        .addSourceLines(
+            "A.java",
+            "class A {",
```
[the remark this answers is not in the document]
rickie: Now that there are quite some characters of the alphabet I'm not sure if it's still the easiest / best way to use them in that way. Although at first it absolutely made sense :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import tech.picnic.errorprone.bugpatterns.util.SourceCode;
+
+/**
+ * A {@link BugChecker} that flags {@link String#toLowerCase()} or {@link String#toUpperCase()}
+ * which do not specify a Locale.
```
rickie: Hmm `{@link Locale}` requires an important so not too nice, but maybe `code` is nice?

rickie (commented):
Hey @mlrprananta thanks for opening the PR, it's a nice one 🚀 !

Made some suggestions and some open questions. Some of them I applied already but they are not final, as in still open for discussion 😄. Don't have time to fully finalize the review but it's in a good state.

The biggest point for me is the name of the check 🤔. Let's see :).

Added a commit.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary =
+        "Specify `Locale.ROOT` or `Locale.getDefault()` when calling `String#to{Lower,Upper}Case` without a specific Locale",
+    link = BUG_PATTERNS_BASE_URL + "SpecifyLocale",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = LIKELY_ERROR)
+public final class SpecifyLocale extends BugChecker implements MethodInvocationTreeMatcher {
```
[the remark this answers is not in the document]
mlrprananta: Likewise not sure about naming, I'm open for suggestions

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
```diff
@@ -0,0 +1,64 @@
...
+          .build();
+    }
+    return Description.NO_MATCH;
+  }
+
+  private static Fix buildFix(
+      String localeToSpecify, MethodInvocationTree tree, VisitorState state) {
+    return SuggestedFix.builder()
+        .replace(
+            tree, SourceCode.treeToString(tree, state).replace("()", "(" + localeToSpecify + ")"))
```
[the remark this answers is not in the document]
mlrprananta: Hmm I don't have a huge preference, but I find my way a bit more clear in what's happening

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/SpecifyLocaleTest.java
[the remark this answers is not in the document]
mlrprananta: Did you have an alternative in mind?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
```diff
@@ -0,0 +1,62 @@
...
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Specify a `Locale` when calling `String#to{Lower,Upper}Case`",
+    link = BUG_PATTERNS_BASE_URL + "SpecifyLocale",
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = LIKELY_ERROR)
+public final class SpecifyLocale extends BugChecker implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<ExpressionTree> STRING_TO_LOWER_OR_UPPER_CASE =
```
Stephan202:
```suggestion
  private static final Matcher<ExpressionTree> DEFAULT_LOCALE_CASE_CONVERSION =
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
```diff
@@ -0,0 +1,62 @@
...
+    return buildDescription(tree)
+        .addFix(suggestLocale("Locale.ROOT", tree, state))
+        .addFix(suggestLocale("Locale.getDefault()", tree, state))
+        .build();
+  }
+
+  private static Fix suggestLocale(String locale, MethodInvocationTree tree, VisitorState state) {
+    return SuggestedFix.builder()
+        .addImport("java.util.Locale")
+        .replace(tree, SourceCode.treeToString(tree, state).replace("()", "(" + locale + ")"))
```
Stephan202: This assumes that the code is properly formatted. More robust:
```suggestion
        .replace(tree, SourceCode.treeToString(tree, state).replace("(", '(' + locale ))
```

Due to auto-formatting of our test code we can't just add an extra space, but this case can be modelled using an extra comment. Will push something.

Edit: even this is too fragile. Consider `str.toUpperCase( /* () */ )`. So it should be:
```suggestion
        .replace(tree, SourceCode.treeToString(tree, state).replaceFirst("\\(", '(' + locale))
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
```diff
@@ -0,0 +1,62 @@
...
+      instanceMethod()
+          .onExactClass(String.class.getName())
+          .namedAnyOf("toLowerCase", "toUpperCase")
+          .withNoParameters();
+
+  /** Instantiates a new {@link SpecifyLocale} instance. */
+  public SpecifyLocale() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
```
Stephan202, line 46: Next to a `MethodInvocationTree` we should also handle the associated method reference.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
Stephan202: Would go for `WARNING` and `FRAGILE_CODE`, since the default often _usually_ does what one wants.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import tech.picnic.errorprone.bugpatterns.util.SourceCode;
+
+/**
+ * A {@link BugChecker} that flags {@link String#toLowerCase()} or {@link String#toUpperCase()}
+ * which do not specify a {@code Locale}.
```
Stephan202:
```suggestion
 * A {@link BugChecker} that flags calls to {@link String#toLowerCase()} and {@link
 * String#toUpperCase()}, as these methods implicitly rely on the environment's default locale.
```

Stephan202 (commented):
Rebased and added a commit.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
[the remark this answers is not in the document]
Stephan202: Also not sure. Best I got for now is `ImplicitLocaleUsage`; will push, but also open to suggestions.

Stephan202 (commented):
Pushed renaming suggestion. Only open point from my side is the method reference handling. Could also be deferred (I already added an `XXX` comment), but would like to make a habit of covering this case.

Stephan202 (approved):
Better to defer; expanded the comment. Suggested commit message:
```
Introduce `ImplicitLocaleUsage` check (#376)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitLocaleUsage.java
```diff
@@ -0,0 +1,66 @@
...
+    return buildDescription(tree)
+        .addFix(suggestLocale(tree, "Locale.ROOT", state))
+        .addFix(suggestLocale(tree, "Locale.getDefault()", state))
+        .build();
+  }
+
+  private static Fix suggestLocale(MethodInvocationTree tree, String locale, VisitorState state) {
+    return SuggestedFix.builder()
+        .addImport("java.util.Locale")
+        .replace(tree, SourceCode.treeToString(tree, state).replaceFirst("\\(", '(' + locale))
```
Stephan202, line 67: Even this isn't correct. Added another comment for further follow-up work.

rickie (commented):
Looked at the changes and they LGTM! 

We could also decide to make GH issues out of the `XXX`es, but not sure if necessary for both though. 
I think deferring the `XXX`'es is a smart thing to do as it indeed probably involves a new utility class which are usually not that easy. 

One thing a like about `ImplicitLocaleUsage` is the word implicit. What I do not really like is that `String` is "missing". As in, it captures the essense, but the focus is clearly on `Strings` which is not really clear from the name. Therefore I'm proposing two others:
- `StringLocaleUsage`
- `StringCaseLocaleUsage`

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpecifyLocale.java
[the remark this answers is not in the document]
mlrprananta: I agree with @rickie that we should specify `String` as well to make it more specific, unless we want to extend this rule  to other methods/classes as well later?

Would `StringCaseImplicitLocaleUsage` be too long? 😅

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitLocaleUsage.java
mlrprananta: Just to be clear, should the `XXX` be tackled in this PR?

rickie:
> W.r.t. the checker name: StringCaseLocaleUsage work for me. Or perhaps ImplicitStringCaseLocale? :thinking:

To be honest, I prefer the `Usage` one as it is more in line with some other names we used till now.

rickie (approved):
Nice, thanks for being fast with replying and applying @mlrprananta  😉. 

You're probably the first one to receive the new Christmas sticker 🎄 🎅🏻 !

PR #384, opened 2022-12-04
Introduce `MonoFromOptionalSwitchIfEmpty` and `OptionalMapMonoJust` Refaster rules

werli:
I've spotted the following interesting pattern:
```java
Optional.of(someNullableValue())
  .map(Mono::just)
  .orElse(someMono());
```
The mix between `Optional` and `Mono` API is unnecessary.

These rules aim to rewrite these to:
```java
Mono.justOrEmpty(someNullableValue())
  .switchIfEmpty(someMono());
```

---
Suggested commit message:
```
Introduce `MonoFromOptionalSwitchIfEmpty` Refaster rule (#384)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -119,20 +119,34 @@ final class ReactorRules {
   }
 
   /**
+   * Try to avoid expressions of type {@code Optional<Mono<T>>}, but if you must map an {@link
+   * Optional} to this type, prefer using {@link Mono#just(Object)}.
+   */
+  static final class OptionalMapMonoJust<T> {
+    @BeforeTemplate
+    Optional<Mono<T>> before(Optional<T> optional) {
+      return optional.map(Mono::justOrEmpty);
+    }
+
+    @AfterTemplate
+    Optional<Mono<T>> after(Optional<T> optional) {
+      return optional.map(Mono::just);
+    }
+  }
+
+  /**
    * Prefer a {@link Mono#justOrEmpty(Optional)} and {@link Mono#switchIfEmpty(Mono)} chain over
    * more contrived alternatives.
    */
   static final class MonoFromOptionalSwitchIfEmpty<T> {
     @BeforeTemplate
-    Mono<T> before(Optional<T> optional, Mono<T> alternative) {
-      return optional
-          .map(Refaster.<Function<T, Mono<T>>>anyOf(Mono::just, Mono::justOrEmpty))
-          .orElse(alternative);
+    Mono<T> before(Optional<T> optional, Mono<T> mono) {
+      return optional.map(Mono::just).orElse(mono);
     }
 
     @AfterTemplate
-    Mono<T> after(Optional<T> optional, Mono<T> alternative) {
-      return Mono.justOrEmpty(optional).switchIfEmpty(alternative);
+    Mono<T> after(Optional<T> optional, Mono<T> mono) {
+      return Mono.justOrEmpty(optional).switchIfEmpty(mono);
     }
   }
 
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -78,6 +78,24 @@ Mono<T> after(Optional<T> optional) {
...
 
+  /**
+   * Prefer a {@link Mono#justOrEmpty(Optional)} and {@link Mono#switchIfEmpty(Mono)} chain over
+   * more contrived alternatives.
+   */
+  static final class MonoFromOptionalSwitchIfEmpty<T> {
+    @BeforeTemplate
+    Mono<T> before(Optional<T> optional, Mono<T> alternative) {
+      return optional
+          .map(Refaster.<Function<T, Mono<T>>>anyOf(Mono::just, Mono::justOrEmpty))
```
Stephan202: The `Mono::justOrEmpty` -> `Mono::just` case can be its own "precursor rule", following our approach to defining the smallest improvement per Refaster rule. Will push a proposal.

Stephan202 (approved):
Thanks for opening a PR!

Added one commit :)

Tweaked suggested commit message (a bit long, but alas):
```
Introduce `OptionalMapMonoJust` and `MonoFromOptionalSwitchIfEmpty` Refaster rules
(#384)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
werli: Good one, thanks. I didn't see that initially.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -78,6 +78,38 @@ Mono<T> after(Optional<T> optional) {
...
+    }
+
+    @AfterTemplate
+    Optional<Mono<T>> after(Optional<T> optional) {
+      return optional.map(Mono::just);
+    }
+  }
+
+  /**
+   * Prefer a {@link Mono#justOrEmpty(Optional)} and {@link Mono#switchIfEmpty(Mono)} chain over
```
rickie, line 98: I know this is our usual way of defining the Javadoc. 
However, it couldn't hurt to mention the exact reason either as `@Description` or instead of `over more contrived alternatives`. Not mixing the `Optional` and `Mono` API is a good argument and nice to document here as well.

We could IMO at least start with adding `@Description`s here and there.

rickie (approved):
Got one question.

Besides that, no remarks. LGTM! 

W.r.t. suggested commit message, LGTM as well.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
werli: I can try to come up with something later today. :crossed_fingers:

rickie: Let's go with Javadoc then, deal?

Stephan202: Rebased and added a commit; PTAL.

Stephan202:
> Unable to resolve action `actions/setup-java@v3.7.0`, unable to find version `v3.7.0`

Build fails due to actions/setup-java#422...

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
rickie: LGTM!

werli: Yeah, would indeed feel like a different concern. Would have been weird starting this now in this PR. :thinking:

rickie:
Swapped the order of the names to make 'em sorted 😉.

Issue #387, opened 2022-12-04
Rewrite `collection.stream().forEach(consumer)` to `collection.forEach(consumer)`


Stephan202:
### Problem

Summary says it; this is a good candidate for a Refaster rule.

### Description of the proposed new feature

- [x] Support a stylistic preference.
- [ ] Avoid a common gotcha, or potential problem.

Use Refaster to rewrite code of the form:
```java
collection.stream().forEach(consumer);
```
to
```java
collection.forEach(consumer);
```

### Considerations

_In theory_ this rewrite rule could change the code's semantics, but this is unlikely in practice.

amestoyg:
I'd like to pick this one up.

PR #390, opened 2022-12-05
Introduce `CollectionForEach` Refaster rule

amestoyg:
Related to the issue #387, I added a new Refaster rule to rewrite the use of `forEach` when working with collections.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
@@ -17,6 +17,7 @@ import java.util.Optional;
 import java.util.Queue;
 import java.util.Set;
 import java.util.SortedSet;
+import java.util.function.Consumer;
 import java.util.function.IntFunction;
 import java.util.stream.Stream;
 import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
@@ -401,6 +402,19 @@ final class CollectionRules {
     @AfterTemplate
     Optional<T> after(Queue<T> queue) {
       return Optional.ofNullable(queue.poll());
+    }
+  }
+
+  /** Prefer calling {@link Collection#forEach} ()} over more contrived alternatives. */
+  static final class CollectionForEach<T> {
+    @BeforeTemplate
+    void before(Collection<T> collection, Consumer<T> consumer) {
+      collection.stream().forEach(consumer);
+    }
+
+    @AfterTemplate
+    void after(Collection<T> collection, Consumer<T> consumer) {
+      collection.forEach(consumer);
     }
   }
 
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
```diff
@@ -404,6 +405,19 @@ Optional<T> after(Queue<T> queue) {
     }
   }
 
+  /** Prefer calling {@link Collection#forEach} ()} over more contrived alternatives. */
```
Stephan202, line 408:
```suggestion
  /** Prefer {@link Collection#forEach(Consumer)} over more contrived alternatives. */
```
(I see we use the extra word "calling" in some places in this class, but elsewhere we don't. Something to unify, perhaps in an automated fashion :smile:.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
```diff
@@ -404,6 +405,19 @@ Optional<T> after(Queue<T> queue) {
...
+    @BeforeTemplate
+    void before(Collection<T> collection, Consumer<T> consumer) {
+      collection.stream().forEach(consumer);
+    }
+
+    @AfterTemplate
+    void after(Collection<T> collection, Consumer<T> consumer) {
+      collection.forEach(consumer);
+    }
+  }
```
Stephan202, line 419: The `forEach` methods accept a `Consumer<? super T>`. So we can define a more generic rule:

```suggestion
  static final class CollectionForEach<S, T extends S> {
    @BeforeTemplate
    void before(Collection<T> collection, Consumer<S> consumer) {
      collection.stream().forEach(consumer);
    }

    @AfterTemplate
    void after(Collection<T> collection, Consumer<S> consumer) {
      collection.forEach(consumer);
    }
  }
```

Stephan202 (commented):
Nice; left two suggestions. Congrats on your first contribution! :tada:

NB: The build fails because of actions/setup-java#422; if that issue doesn't resolve itself soon we'll open a downgrade PR for this repo.

werli (approved):
Optimistically approving assuming @Stephan202's suggestions will be taken care of. :slightly_smiling_face:

amestoyg:
Thanks for the comments @Stephan202. I pushed the suggested changes 🙂

Stephan202:
Nice! Suggested commit message:
```
Introduce `CollectionForEach` Refaster rule (#390)
```

rickie (approved):
Nice @amestoyg, congrats on your first PR 🎉! 

@Stephan202 tweaked suggested commit message.
I linked the PR already, but still I think we should have the habit to refer to the tickets in the commit message.

Do we either prefer the terminology "Fixes", "Resolves" or "Closes"?

Issue #386, opened 2022-12-04
Refaster rules that encourage filtering before sorting


Stephan202:
### Problem

Given a stream of elements that must be filtered and sorted, it is generally more efficient to do the former first, so that the subsequent sorting step involves fewer elements. This applies to `Stream`, `IntStream`, `LongStream`, `DoubleStream` and `Flux` types.

### Description of the proposed new feature

- [ ] Support a stylistic preference.
- [x] Avoid a common gotcha, or potential problem.

Rewrite:
```
stream.sorted().filter(pred)
```
to:
```
stream.filter(pred).sorted()
```

Do the same for the other types mentioned above, as well as `Comparator`-accepting sort method overloads.

### Considerations

One nice advantage of the proposed rewrite rules is that [these simplifications](https://github.com/PicnicSupermarket/error-prone-support/blob/5afa7e1878453f97996ed73c36847844e07074cb/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java#L222-L2460) might subsequently kick in.

CoolTomatos:
I would like to pick this one up for the cute new Christmax edition sticker haha.

PR #393, opened 2022-12-06
Refaster rules that encourage filtering before sorting.

CoolTomatos:
Closes PicnicSupermarket/error-prone-support#386.


Stephan202:
Rebased to test whether I can push to the branch, which worked :heavy_check_mark: Adding the PR to my review list :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -167,6 +167,34 @@ Stream<R> after(Stream<T> stream, Function<? super S, ? extends Stream<? extends
     }
   }
 
+  /** Apply filtering before sorting to reduce the number of elements to sort. */
```
Stephan202: Perhaps slightly more idiomatic:
```suggestion
  /** Filter before sorting to reduce the number of elements to sort. */
```

Stephan202: Alternatively:
```suggestion
  /** Apply {@link Stream#filter(Predicate)} before {@link Stream#sorted()} to reduce the number of elements to sort. */
```

In most cases we do reference the relevant methods, so let's go with this one.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -167,6 +167,34 @@ Stream<R> after(Stream<T> stream, Function<? super S, ? extends Stream<? extends
...
+  static final class SortAfterFilter<T> {
```
Stephan202: (We still have to automate this, but) our current naming scheme would dictate that this class is named `StreamFilterSorted`.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StreamRulesTestInput.java
```diff
@@ -130,4 +131,14 @@ ImmutableSet<Boolean> testStreamAllMatch() {
...
     return Stream.of("foo").noneMatch(s -> !s.isBlank());
   }
+
+  Stream<Integer> testSortAfterFilter() {
+    return Stream.of(1, 4, 3, 2).sorted().filter(i -> i % 2 == 0);
+  }
+
+  Stream<Integer> testSortWithComparatorAfterFilter() {
+    return Stream.of(1, 4, 3, 2)
+        .sorted(Comparator.comparingInt(Integer::intValue).reversed())
```
Stephan202: If we use `reverseOrder()` we can simplify `elidedTypesAndStaticImports` above.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1246,4 +1247,36 @@ Duration after(StepVerifier.LastStep step, Duration duration) {
       return step.verifyTimeout(duration);
     }
   }
```
Stephan202: Right now we have the `StepVerifier`-related rules at the bottom; let's keep that.

Stephan202 (approved):
Added one more commit. Very nice contribution @CoolTomatos!

Suggested commit message:
```
Introduce Refaster `Flux` and `Stream` rules to suggest filtering before sorting (#393)

Fixes #386.
```

rickie (approved):
Rebased and added a commit.

Changes LGTM! 

Made one tweak to the suggested commit message to be more inline with the previous commit messages (IMO).

Thanks for opening a PR Sir @CoolTomatos 😉!

rickie:
Rebased, will merge once 💚.

PR #394, opened 2022-12-06
Prefer `BigDecimal.valueOf(double)` over `new BigDecimal(double)`

Venorcis:
A very nice rule to rewrite `new BigDecimal(long/int)` to `BigDecimal.valueOf(long/int)` already existed. As using `new BigDecimal(double)` is [quite discouraged](https://rules.sonarsource.com/java/RSPEC-2111), this PR extends the existing rule to rewrite `new BigDecimal(double/long/int)` to `BigDecimal.valueOf(double/long/int)`

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
@@ -54,7 +54,7 @@ final class BigDecimalRules {
   // XXX: Ideally we also rewrite `new BigDecimal("<some-integer-value>")` in cases where the
   // specified number can be represented as an `int` or `long`, but that requires a custom
   // `BugChecker`.
-  static final class BigDecimalFactoryMethod {
+  static final class BigDecimalValueOf {
     @BeforeTemplate
     BigDecimal before(double value) {
       return new BigDecimal(value);
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
@@ -51,8 +51,9 @@ final class BigDecimalRules {
   }
 
   /** Prefer {@link BigDecimal#valueOf(double)} over the associated constructor. */
-  // XXX: Ideally we'd also rewrite `new BigDecimal("<some-integer-value>")`, but it doesn't
-  // appear that's currently possible with Error Prone.
+  // XXX: Ideally we also rewrite `new BigDecimal("<some-integer-value>")` in cases where the
+  // specified number can be represented as an `int` or `long`, but that requires a custom
+  // `BugChecker`.
   static final class BigDecimalFactoryMethod {
     @BeforeTemplate
     BigDecimal before(double value) {
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
@@ -50,17 +50,17 @@ final class BigDecimalRules {
     }
   }
 
-  /** Prefer {@link BigDecimal#valueOf(long)} over the associated constructor. */
-  // XXX: Ideally we'd also rewrite `BigDecimal.valueOf("<some-integer-value>")`, but it doesn't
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
```diff
@@ -51,9 +51,7 @@ BigDecimal after() {
   }
 
   /** Prefer {@link BigDecimal#valueOf(long)} over the associated constructor. */
-  // XXX: Ideally we'd also rewrite `BigDecimal.valueOf("<some-integer-value>")`, but it doesn't
```
Venorcis, line 54: AFAIK this actually works rn.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/BigDecimalRulesTestInput.java
```diff
@@ -17,7 +17,11 @@ ImmutableSet<BigDecimal> testBigDecimalTen() {
     return ImmutableSet.of(BigDecimal.valueOf(10), BigDecimal.valueOf(10L), new BigDecimal("10"));
   }
 
-  ImmutableSet<BigDecimal> testBigDecimalFactoryMethod() {
-    return ImmutableSet.of(new BigDecimal(0), new BigDecimal(0L));
+  ImmutableSet<BigDecimal> testBigDecimalLongFactoryMethod() {
+    return ImmutableSet.of(new BigDecimal(2), new BigDecimal(2L));
```
Venorcis, line 21: felt `2` was a nicer value to use here as there is a rule to rewrite 0 to `ZERO` as well

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
```diff
@@ -50,17 +50,17 @@ BigDecimal after() {
     }
   }
 
-  /** Prefer {@link BigDecimal#valueOf(long)} over the associated constructor. */
-  // XXX: Ideally we'd also rewrite `BigDecimal.valueOf("<some-integer-value>")`, but it doesn't
+  /** Prefer {@link BigDecimal#valueOf(double)} over the associated constructor. */
+  // XXX: Ideally we'd also rewrite `new BigDecimal("<some-integer-value>")`, but it doesn't
```
Venorcis, line 54: think this was meant here, not sure if this is a nice comment to keep around anyway, but I'll leave that up to you 😉

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/BigDecimalRulesTestInput.java
```diff
@@ -18,6 +18,6 @@ ImmutableSet<BigDecimal> testBigDecimalTen() {
   }
 
   ImmutableSet<BigDecimal> testBigDecimalFactoryMethod() {
-    return ImmutableSet.of(new BigDecimal(0), new BigDecimal(0L));
+    return ImmutableSet.of(new BigDecimal(2), new BigDecimal(2L), new BigDecimal(2.0));
```
Venorcis, line 21: felt `2` was a nicer value to use here as there is a rule above to rewrite `0` to `ZERO` as well

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
Stephan202: Hmm, I wonder whether the comment meant a bit of both: if `"<some-integer-value>"` represents a value that can be expressed as a `long`, then passing in a `long`/`int` is nicer than a string. (But such a change can't be done using Refaster; it'd require a `BugChecker`.)

Stephan202 (commented):
Nice one @Venorcis! I'll circle back later for a full review :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
Stephan202: Ah, I see that `BigDecimal.valueOf` doesn't exist. So certainly that was _not_ meant :smile:

Stephan202 (commented):
Added a small commit. Have a meeting now; will finalize review afterwards :)

PR #397, opened 2022-12-07
Prevent `ReverseOrder` Refaster rule from introducing a static import

Stephan202:
This is a workaround for google/error-prone#3584.

Suggested commit message:
```
Prevent `ReverseOrder` Refaster rule from introducing a static import (#397)

This is a workaround for the issue resolved by google/error-prone#3584.

After application of this Refaster rule, any static imports of
`java.util.Collections.reverseOrder` are obsolete. These can be removed by
running Google Java Format or Error Prone's `RemoveUnusedImports` check.

Where possible, subsequent application of the `StaticImport` check will
statically import `java.util.Comparator.reverseOrder`.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ComparatorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ComparatorRules.java
@@ -62,8 +62,9 @@ final class ComparatorRules {
           Comparator.<T>naturalOrder().reversed());
     }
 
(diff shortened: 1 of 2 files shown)


rickie (approved):
Thanks for the quick turn around and fixing this!

W.r.t. suggested commit message made one tweak.

rickie:
@Stephan202 shouldn't it be both "check"?

rickie:
Rebased.

Stephan202:
Doh, I misread the diff. I'm fine either way (I also prefer `check`, but the interface is named `BugChecker`.) Anyway, fixed.

eric-picnic (approved):

PR #400, opened 2022-12-08
Fix suggestions emitted by the `StringCaseLocaleUsage` check

rickie:
Encountered this while testing the upcoming release on downstream code.

The following code:
```
return this.name().toLowerCase();
```
Got rewritten to:
```
return this.name(Locale.ROOT).toLowerCase();
```
But should've been: 
```
return this.name().toLowerCase(Locale.ROOT);
```

After some testing, it turns out that searching for the _first_ occurrence of `(` is much more error-prone than searching for the last occurrence. Therefore I propose to change the implementation accordingly. 

NB: To be (arguably) more correct, I changed the implementation to find the first closing bracket after the last opening bracket. There are cases where this can still cause problems. An example:
```
// This code:
"a".toLowerCase(/* Comment with parens: ((. */ );
// would currently result in:
"a".toLowerCase(/* Comment with parens: ((Locale.ROOT);  
// at the end we are missing: `*/...`
```
Yes we should tackle this, but I think it'd be better for now to leave this as is and create a ticket to properly fix this.

So this is indeed not the optimal solution and we should implement a more correct one. However, for now I'd argue this is a clear improvement.


error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/StringCaseLocaleUsageTest.java
```diff
@@ -62,6 +62,13 @@ void replacementFirstSuggestedFix() {
...
             "    \"a\".toLowerCase(/* Comment with parens: (). */ );",
             "    \"b\".toUpperCase();",
+            "",
+            "    getString().toLowerCase();",
+            "    getString().toUpperCase /* Comment with parens: (). */();",
+            "  }",
+            "",
+            "  private String getString() {",
+            "    return \"\";",
             "  }",
```
Stephan202: We can simply use `toString()` instead of a custom `getString()`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StringCaseLocaleUsage.java
```diff
@@ -58,13 +58,22 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
     return SuggestedFix.builder()
         .addImport("java.util.Locale")
-        .replace(tree, SourceCode.treeToString(tree, state).replaceFirst("\\(", '(' + locale))
+        .replace(
+            tree,
+            source.substring(0, indexOfLastOpeningBracket)
+                + "("
+                + locale
+                + sourceAfterLastOpeningBracket.substring(indexOfClosingBracket))
         .build();
```
Stephan202: If a known-hacky solution bit us once, then we shouldn't replace it with another hacky solution. It seem better to look only at the final closing paren. Additionally, we can take inspiration from Error Prone's (Java)doc-related checks to see how we can robustly avoid matching comments. Will push a proposal.

Stephan202 (approved):
Added a commit.

Suggested commit message:
```
Fix suggestions emitted by `StringCaseLocaleUsage` check (#400)

The suggested `Locale` arguments are now always located in the correct place.
```

Stephan202:
^ The four surviving mutants are in code that handles the absence of source code; not sure how we can test those cases without jumping through lots of hoops. Not worth the effort :)

Badbond (approved):
Nice bug fix and general improvement :+1: Suggested commit message LGTM. Didn't have much input but for _maybe_ one test case that can be added.

rickie:
Added commit. 

Will merge once 🟢 .

Added one word to the suggested commit message 👀.

Issue #399, opened 2022-12-08
Have `IsInstanceLambdaUsage` handle extra method invocations


rickie:
### Describe the bug

- [x] I have verified that the issue is reproducible against the latest version
      of the project.
- [x] I have searched through existing issues to verify that this issue is not
      already known.

### Minimal Reproducible Example

The `IsInstanceLambdaUsage` check works for _most_ cases. However, when we invoke a method on the object passed to the filter, the suggestion is incorrect, which leads to non-compilable code.

This code:
```java
Stream.of(1).filter(i -> i.getClass() instanceof Class);

// Given `locations` is a list of `Location` objects.
Stream.of(locations).filter(location -> location.getName() instanceof String);
```
Is rewritten to:
```java
Stream.of(1).filter(Class.class::isInstance);

// Given `locations` is a list of `Location` objects.
Stream.of(locations).filter(String.class::isInstance);
```
This is **not** behavior preserving, as we are now checking whether `i` is an instance of `Class`. This usually leads to non-compilable code. To clarify, in the second example we are now checking that `location` is an instance of `String` instead of the `location.getName()`.

### Expected behavior

I think the best option for now would be to **not** flag these cases. We probably can't make a clear improvement here.

### Setup

- Operating system: `Linux Ubuntu 20.04`
- Java version: `openjdk 17.0.3 2022-04-19 LTS`
- Error Prone version: `2.16`.
- Error Prone Support version: `0.5.0`.

PR #401, opened 2022-12-08
Improve `IsInstanceLambdaUsage` check

Ptijohn:
By checking `isinstanceof` usages only against identifiers.

Fixes [this issue](https://github.com/PicnicSupermarket/error-prone-support/issues/399).


error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/IsInstanceLambdaUsageTest.java
```diff
@@ -23,6 +23,7 @@ void identification() {
             "    // BUG: Diagnostic contains:",
             "    Stream.of(1).filter(i -> i instanceof Integer);",
```
Stephan202: The following expression is also flagged, while it shouldn't:
```java
Stream.of(1).filter(i -> someOtherLocalVariable instanceof Integer);
```

Similarly, we should not flag `BiPredicate`s such as in:
```java
Flux.just(1, "foo").distinctUntilChanged(v -> v, (a, b) -> a instanceof Integer);

Comment (Stephan202) [reply error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/IsInstanceLambdaUsageTest.java]: While there: in `identification` tests we often list the "good" cases first; I'll shuffle om things.

File: error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceLambdaUsage.java [hunk 2]
```diff
@@ -39,7 +39,9 @@ public IsInstanceLambdaUsage() {}
 
   @Override
   public Description matchLambdaExpression(LambdaExpressionTree tree, VisitorState state) {
-    if (tree.getKind() != Kind.LAMBDA_EXPRESSION || tree.getBody().getKind() != Kind.INSTANCE_OF) {
+    if (tree.getKind() != Kind.LAMBDA_EXPRESSION
+        || tree.getBody().getKind() != Kind.INSTANCE_OF
+        || ((InstanceOfTree) tree.getBody()).getExpression().getKind() != Kind.IDENTIFIER) {
       return Description.NO_MATCH;
```
Stephan202: While we're here, let's kill the highlighted mutants.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceLambdaUsage.java
```diff
@@ -39,7 +39,9 @@ public IsInstanceLambdaUsage() {}
 
   @Override
   public Description matchLambdaExpression(LambdaExpressionTree tree, VisitorState state) {
-    if (tree.getKind() != Kind.LAMBDA_EXPRESSION || tree.getBody().getKind() != Kind.INSTANCE_OF) {
+    if (tree.getKind() != Kind.LAMBDA_EXPRESSION
+        || tree.getBody().getKind() != Kind.INSTANCE_OF
```
Stephan202: Side-remark: at some point we should take a stance around testing the `Kind` versus performing an `instanceof` check. I expect the former to be faster (didn't check), but the latter is easier to understand for new users. Also, once we set JDK 17 as a baseline we may naturally start to use the latter i.c.w. pattern matching.

Stephan202: It looks like `tree.getBody().getKind() == Kind.INSTANCE_OF` implies `tree.getKind() == Kind.LAMBDA_EXPRESSION`, so we can skip the latter check.

Stephan202 (approved):
```
Improve `IsInstanceLambdaUsage` check (#401)

Fixes #399.
```

Ptijohn:
Oh well, thanks a lot @Stephan202 :D 
Was planning on digging on those mutants and all today, but I guess I'll learn from you instead ;) (which also works ;) )

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceLambdaUsage.java
rickie: I was looking at this but I think this is not correct. Isn't it because we are implementing a `LambdaExpressionTreeMatcher`, that makes the first condition always true?

rickie:
Rebased. Will merge once 🍏.

rickie:
Oh and btw, thanks for the **epic** fixes @Stephan202. Shows us we need to think _even more_ about all the edge cases.

rickie (approved):
Thanks @Ptijohn for picking this up so fast 🚀 !

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceLambdaUsage.java
Stephan202: ^ Discussed offline: in case of a "block lambda" we do hit this code, but `tree.getBody().getKind()` will then be `    BLOCK`.

PR #405, opened 2022-12-09
Introduce `MonoIdentity` and `MonoThen` Refaster rules

pacbeckh:
We came across this pattern in our code base after some (automatic) refactorings.

```java
mono
    .flatMap(operationReturningMonoVoid)
    .flux()
    .then();
```

`.flux().then()` doesn't add anything, similarly to calling `.then` on a `Mono<Void>`.

Introduced two rules for both steps

---
Suggested commit message:
```
Introduce `MonoFluxThen` and `MonoVoidThen` Refaster rules (#405)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -675,6 +675,34 @@ final class ReactorRules {
     }
   }
 
+  /** Don't unnecessarily convert {@link Mono} to {@link Flux}. */
+  @SuppressWarnings("VoidMissingNullable")
+  static final class MonoFluxThen<T> {
+    @BeforeTemplate
+    Mono<Void> before(Mono<T> mono) {
+      return mono.flux().then();
+    }
+
+    @AfterTemplate
+    Mono<Void> after(Mono<T> mono) {
+      return mono.then();
+    }
+  }
+
+  /** Don't unnecessarily call {@link Mono#then()} on a {@code Mono<Void>} . */
+  @SuppressWarnings("VoidMissingNullable")
+  static final class MonoVoidThen {
+    @BeforeTemplate
+    Mono<Void> before(Mono<Void> mono) {
+      return mono.then();
+    }
+
+    @AfterTemplate
+    Mono<Void> after(Mono<Void> mono) {
+      return mono;
+    }
+  }
+
   /**
    * Prefer a collection using {@link MoreCollectors#toOptional()} over more contrived alternatives.
    */
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -365,11 +365,19 @@ final class ReactorRules {
     }
   }
 
-  /** Don't unnecessarily pass an empty publisher to {@link Mono#switchIfEmpty(Mono)}. */
-  static final class MonoSwitchIfEmptyOfEmptyPublisher<T> {
+  /** Don't unnecessarily transform a {@link Mono} to an equivalent instance. */
+  static final class MonoIdentity<T> {
     @BeforeTemplate
     Mono<T> before(Mono<T> mono) {
       return mono.switchIfEmpty(Mono.empty());
+    }
+
+    // XXX: Review the suppression once NullAway has better support for generics. Keep an eye on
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -675,6 +675,34 @@ Flux<T> after(Mono<T> mono) {
...
+    }
+
+    @AfterTemplate
+    Mono<Void> after(Mono<T> mono) {
+      return mono.then();
+    }
+  }
+
+  /** Don't unnecessarily call {@link Mono#then()} on a {@code Mono<Void>} . */
+  @SuppressWarnings("VoidMissingNullable")
```
pacbeckh, line 693: Introduced this because I got into a fight with NullAway:
```
[NullAway] dereferenced expression mono is @Nullable
```
"mono" was nullable:
```
    @BeforeTemplate
    Mono<@Nullable Void> before(Mono<@Nullable Void> mono) {
      return mono.then();
    }
```
    
regardless, I wasn't really sure why we need `@Nullable` on the `Void` type in this context

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -221,6 +221,14 @@ ImmutableSet<Flux<String>> testMonoFlux() {
         Flux.concat(Mono.just("baz")));
   }
 
+  Mono<Void> testMonoFluxThen() {
+    return Mono.just("foo").flux().then();
+  }
+
+  Mono<Void> testMonoVoidThen() {
+    return Mono.just("foo").then().then();
```
pacbeckh, line 229: The most concise way I could think of creating a `Mono<Void>`, not the prettiest

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: IIUC this is a bug in NullAway; indeed [their support for generics is limited](https://github.com/uber/NullAway/issues?q=is%3Aopen+generics). Let's add the `@Nullable` annotations (they are correct) but also keep the suppression (only on this class).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -675,6 +675,34 @@ Flux<T> after(Mono<T> mono) {
...
+  static final class MonoVoidThen {
+    @BeforeTemplate
+    Mono<Void> before(Mono<Void> mono) {
+      return mono.then();
+    }
+
+    @AfterTemplate
+    Mono<Void> after(Mono<Void> mono) {
+      return mono;
+    }
```
Stephan202: We can combine this rule with `MonoSwitchIfEmptyOfEmptyPublisher` above, as we generally try to have one rule for each "target shape" in an `@AfterTemplate`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -675,6 +675,34 @@ Flux<T> after(Mono<T> mono) {
     }
   }
 
+  /** Don't unnecessarily convert {@link Mono} to {@link Flux}. */
+  @SuppressWarnings("VoidMissingNullable")
+  static final class MonoFluxThen<T> {
```
Stephan202: We haven't fully migrated yet, but the naming standard we're thinking of settling on would have this rule be named `MonoThen`, derived from the code in the `@AfterTemplate`.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
Stephan202: How about `Mono.<Void>empty()`? :)

Stephan202 (approved):
Thanks for your contribution @pacbeckh!

I rebased and added a commit. Perhaps the `MonoIdentity` rule should now be moved (we have some "fuzzy logic" for that), but as-is the diff is easier to review.

Suggested commit message:
```
Introduce `MonoIdentity` and `MonoThen` Refaster rules (#405)

The `MonoIdentity` rule is a generalization of the existing
`MonoSwitchIfEmptyOfEmptyPublisher` rule.
```

rickie (approved):
Changes LGTM!

Nice contribution @pacbeckh !

PR #407, opened 2022-12-11
Improve `IdentityConversion` check

Stephan202:
Resolves an issue found while testing Error Prone Support against Checkstyle (currently testing on [this branch](https://github.com/PicnicSupermarket/error-prone-support/compare/sschroevers/integration-tests)).

Suggested commit message:
```
Improve `IdentityConversion` check (#407)

If the result of an explicit boxing operation is immediately
dereferenced, then the explicit conversion operation is not redundant.
```


rickie (approved):
Pretty fun edge case. Fix looks good to me.

chamil-prabodha (approved):

rickie:
Hmm, is this an `improvement` or a `bug fix`?

Stephan202:
- Label-wise: bug fix :+1:
- Suggested commit wise: would leave as-is.

rickie:
Ah yes I only meant label-wise 😄. Nice.

PR #412, opened 2022-12-14
Prevent `NestedOptionals` from throwing an NPE

gtoison:
While testing version 0.6.0 I encountered a `NullPointerException` thrown by `NestedOptinal` when analysing this code (please don't judge, it's a bug sample!):

```
void m() {
    java.time.Duration.of(1, ChronoUnit.YEARS);
}
```

This PR adds a unit test reproducing the problem and a proposed fix.
Thanks a lot for providing this project!

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedOptionals.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedOptionals.java
@@ -41,8 +41,11 @@ public final class NestedOptionals extends BugChecker implements MethodInvocatio
 
   @Override
   public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
-    return state.getTypes().isSubtype(ASTHelpers.getType(tree), OPTIONAL_OF_OPTIONAL.get(state))
-        ? describeMatch(tree)
-        : Description.NO_MATCH;
+    Type stateType = OPTIONAL_OF_OPTIONAL.get(state);
+    if (stateType != null && state.getTypes().isSubtype(ASTHelpers.getType(tree), stateType)) {
+      return describeMatch(tree);
+    } else {
+      return Description.NO_MATCH;
+    }
   }
 }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedOptionals.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedOptionals.java
@@ -41,11 +41,11 @@ public final class NestedOptionals extends BugChecker implements MethodInvocatio
 
   @Override
   public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
-    Type stateType = OPTIONAL_OF_OPTIONAL.get(state);
-    if (stateType != null && state.getTypes().isSubtype(ASTHelpers.getType(tree), stateType)) {
-      return describeMatch(tree);
-    } else {
+    Type type = OPTIONAL_OF_OPTIONAL.get(state);
+    if (type == null || !state.getTypes().isSubtype(ASTHelpers.getType(tree), type)) {
       return Description.NO_MATCH;
     }
+
+    return describeMatch(tree);
   }
 }
--- a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/NestedOptionalsTest.java
+++ b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/NestedOptionalsTest.java
@@ -39,4 +39,19 @@ final class NestedOptionalsTest {
             "}")
         .doTest();
   }
+
+  @Test
+  void failingTypeCheck() {
+    compilationTestHelper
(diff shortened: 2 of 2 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedOptionals.java
```diff
@@ -41,8 +41,11 @@ public NestedOptionals() {}
...
   public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
-    return state.getTypes().isSubtype(ASTHelpers.getType(tree), OPTIONAL_OF_OPTIONAL.get(state))
-        ? describeMatch(tree)
-        : Description.NO_MATCH;
+    Type stateType = OPTIONAL_OF_OPTIONAL.get(state);
+    if (stateType != null && state.getTypes().isSubtype(ASTHelpers.getType(tree), stateType)) {
+      return describeMatch(tree);
+    } else {
+      return Description.NO_MATCH;
+    }
```
Stephan202: Alright, so I guess that the issue happens if the `Optional` class isn't loaded during compilation at all. That realization also lead to [this code](https://github.com/PicnicSupermarket/error-prone-support/blob/96114235c50217b09cb2e5dfd77b36d803ae614c/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java#L77-L85).

Stephan202: We do have a preference for an early return with `Description.NO_MATCH`:
 ```suggestion
    Type type = OPTIONAL_OF_OPTIONAL.get(state);
    if (type == null || !state.getTypes().isSubtype(ASTHelpers.getType(tree), type)) {
      return Description.NO_MATCH;
    }

    return describeMatch(tree);
```

File: error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/NestedOptionalsTest.java [hunk 2]
```diff
@@ -39,4 +39,19 @@ void identification() {
...
+  @Test
+  void failingTypeCheck() {
+    compilationTestHelper
+        .addSourceLines(
+            "B.java",
+            "import java.time.temporal.ChronoUnit;",
+            "",
+            "class B {",
+            "  void m() {",
+            "    java.time.Duration.of(1, ChronoUnit.YEARS);",
```
Stephan202: I suggest using a slightly more realistic case ;)
```
jshell> java.time.Duration.of(1, java.time.temporal.ChronoUnit.YEARS)
|  Exception java.time.temporal.UnsupportedTemporalTypeException: Unit must not have an estimated duration
|        at Duration.plus (Duration.java:717)
|        at Duration.of (Duration.java:312)
|        at (#1:1)
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/NestedOptionalsTest.java
```diff
@@ -39,4 +39,19 @@ void identification() {
             "}")
         .doTest();
   }
+
+  @Test
+  void failingTypeCheck() {
```
Stephan202: Slightly more in line which what we do in a few other places
```suggestion
  void identificationOptionalTypeNotLoaded() {
```

(Funny: I tried to replace the `Duration` expression below with `Stream.empty()`, but then the NPE didn't happen. I think because `Stream` does reference `Optional`, while `Duration` (and `ChronoUnit`) don't.

Stephan202 (approved):
Thanks for trying the new release _and_ filing a bug fix @gtoison! Let us know if you find other issues (or have requests) :)

I added a commit with some suggestions.

Suggested commit message:
```
Prevent `NestedOptionals` from throwing an NPE (#412)

Previously, a `NullPointerException` was thrown if during compilation the
`java.util.Optional` class was not loaded at all.
```

(I suspect there may be similar such booby traps in other checks :thinking:.

@gtoison let us know if you'd like to have a 0.6.1 patch release with this change. I might be able to make time for that in the next 1-2 days.

gtoison:
Thanks for the review and the fixes!
The integration tests I made only use very small bug samples. They check that some rules work as intented, so it's far from exhaustive.

A patch release would be nice but there's no need to rush things. I was planning to upgrade error-prone when there's a new release

rickie (approved):
Thanks for opening a PR @gtoison 🚀 ! 

The changes look good to me 😄.
Also suggested commit message LGTM! 

I'll rebase and merge once the build is green. 

(Ah I don't need to rebase.)

rickie:
Ah got one tweak for the suggested commit message @Stephan202 I think it should be "a NPE" instead of "an NPE" 👀?

Stephan202:
@rickie for acronyms it depends on pronunciation, and I personally literally say "en-pee-ee", so then "an" is correct. I think this is a gray area.

rickie:
Aha, I don't do it like that. Anyway, let's roll with what is suggested then :).

Issue #402, opened 2022-12-08
Rewrite JUnit `Assertions.assertThrows()` to AssertJ `assertThatThrownBy`


oxkitsune:
### Problem

We prefer to use AssertJ for our assertions. So we could use a Refaster rule that automatically prefers this over the assertions library built into JUnit!

### Description of the proposed new feature

- [x] Support a stylistic preference.
- [ ] Avoid a common gotcha, or potential problem.

I would like to rewrite the following code:
```java
org.junit.jupiter.api.Assertions.assertThrows(
    RuntimeException.class,
    () -> {
      throw new RuntimeException("bar");
    });
```

to: 
```java
assertThatThrownBy(() -> {
	throw new RuntimeException("bar");
})
.isInstanceOf(RuntimeException.class):
```

Stephan202:
Good idea! I would like to suggest that we expand the scope of this issue a bit to cover "some sizable subset" of the `org.junit.jupiter.api.Assertions` API. Having a separate PR for each method in that class is not efficient :sweat_smile:.

giall:
Interested in picking this up! Below are the methods in `org.junit.jupiter.api.Assertions` and the equivalent AssertJ code. There are quite a few but quite straight-forward so they could all be part of one big PR.

Conditions:
`assertTrue(condition)` -> `assertThat(condition).isTrue()`
`assertFalse(condition)` -> `assertThat(condition).isFalse()`
`assertNull(obj)` -> `assertThat(obj)isNull()`
`assertNotNull(obj)` -> `assertThat(obj)isNotNull()`
`assertInstanceOf(clazz, obj)` -> `assertThat(actual).isInstanceOf(clazz)` (or `isExactlyInstanceOf`?)

Equality:
`assertEquals(expected, actual)` -> `assertThat(actual).isEqualTo(expected)`
`assertNotEquals(expected, actual)` -> `assertThat(actual).isNotEqualTo(expected)`
`assertSame(expected, actual)` -> `assertThat(actual).isSameAs(expected)`
`assertNotSame(expected, actual)` -> `assertThat(actual).isNotSameAs(expected)`
`assertArrayEquals(expected, actual)` -> `assertThat(actual).containsExactly(expected)`
`assertIterableEquals(expected, actual)` -> `assertThat(actual).containsExactly(expected)`
`assertLinesMatch` -> ?

Exceptions:
[...]

PR #417, opened 2022-12-15
Introduce `JUnitToAssertJRules` Refaster rule collection

giall:
Related issue: https://github.com/PicnicSupermarket/error-prone-support/issues/402

Left `XXX` comments on `org.junit.jupiter.api.Assertions` methods that do not (yet) have rules.

Decided to leave the equality-related ones for a follow-up PR to reduce the scope a little bit, as those also need a bit more work (equality with deltas, etc.).
* `assertEquals`
* `assertArrayEquals`
* `assertIterableEquals`
* `assertNotEquals`

The following methods don't seem to have direct AssertJ equivalents, but could also be handled later.
* `assertLinesMatch`
* `assertAll`
* `assertTimeout`
* `assertTimeoutPreemptively`

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
@@ -73,157 +73,157 @@ final class JUnitToAssertJRules {
 
   static final class AssertThatIsTrue {
     @BeforeTemplate
-    void before(boolean condition) {
-      assertTrue(condition);
+    void before(boolean actual) {
+      assertTrue(actual);
     }
 
     @AfterTemplate
     @UseImportPolicy(STATIC_IMPORT_ALWAYS)
-    void after(boolean condition) {
-      assertThat(condition).isTrue();
+    void after(boolean actual) {
+      assertThat(actual).isTrue();
     }
   }
 
   static final class AssertThatWithFailMessageStringIsTrue {
     @BeforeTemplate
-    void before(boolean condition, String message) {
-      assertTrue(condition, message);
+    void before(boolean actual, String message) {
+      assertTrue(actual, message);
     }
 
     @AfterTemplate
     @UseImportPolicy(STATIC_IMPORT_ALWAYS)
-    void after(boolean condition, String message) {
-      assertThat(condition).withFailMessage(message).isTrue();
+    void after(boolean actual, String message) {
+      assertThat(actual).withFailMessage(message).isTrue();
     }
   }
 
   static final class AssertThatWithFailMessageSupplierIsTrue {
     @BeforeTemplate
-    void before(boolean condition, Supplier<String> supplier) {
-      assertTrue(condition, supplier);
+    void before(boolean actual, Supplier<String> supplier) {
+      assertTrue(actual, supplier);
     }
 
     @AfterTemplate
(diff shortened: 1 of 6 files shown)


rickie (commented):
Added a few commits and rebased. 

I'm slightly in doubt whether we should add Javadocs here @Stephan202. 

For ordering parameters, should we adhere to the order in which they occur in the before or after template? We kind of use both options. 

As we base _most_ thinks on the content of the after template, it could make sense to do the same for this. This would also make more sense given that you can have multiple before templates (this also holds for aftertemplates technically speaking, but we don't do that within Picnic currently).

Not completely done with reviewing the PR but flushing my comments and considerations :).

rickie (approved):
The changes LGTM. Thanks for the PR @giall 🚀 ! Let's get rid of those type of assertions 😄!

Suggested commit message:
```
Introduce `JUnitToAssertJRules` rule collection (#417)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
```diff
@@ -0,0 +1,491 @@
...
+    @AfterTemplate
+    @UseImportPolicy(STATIC_IMPORT_ALWAYS)
+    void after(Object object) {
+      assertThat(object).isNotNull();
+    }
+  }
+
+  static final class AssertThatWithFailMessageStringIsNotNull {
+    @BeforeTemplate
+    void before(Object object, String message) {
```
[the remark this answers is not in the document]
rickie: (We plan to automate the naming of the parameter names and the order in which they should be defined.)

rickie:
Note to self: we should add Javadoc to the Refaster rules that are added.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
```diff
@@ -0,0 +1,491 @@
...
+    void after(Object actual, Supplier<String> supplier) {
+      assertThat(actual).withFailMessage(supplier).isNotNull();
+    }
+  }
+
+  // XXX: Rewrite `org.junit.jupiter.api.Assertions.assertEquals`.
+  // XXX: Rewrite `org.junit.jupiter.api.Assertions.assertArrayEquals`.
+  // XXX: Rewrite `org.junit.jupiter.api.Assertions.assertIterableEquals`.
+  // XXX: Rewrite `org.junit.jupiter.api.Assertions.assertLinesMatch`.
+  // XXX: Rewrite `org.junit.jupiter.api.Assertions.assertNotEquals`.
```
Stephan202: This listing isn't exhaustive (especially when we consider the various overloads), and making it so doesn't seem worth the hassle, if done manually. Let's call out the incompleteness at the start of the class, and defer further work to another PR.

Stephan202: I started work on this idea [here](https://github.com/PicnicSupermarket/error-prone-support/compare/sschroevers/introduce-refaster-type-migration-annotation) (heavy WIP, and some changes should be moved to separate PRs, but the gist is there).

rickie: Wow, that branch is awesome! Looks like some nice changes @Stephan202 !

PR #419, opened 2022-12-17
Disable failing JDK 20-ea build for now

Stephan202:
See the analysis in #418 for context.

Suggested commit message:
```
Disable failing JDK 20-ea build for now (#419)

The build fails due to
openjdk/jdk20@2cb64a75578ccc15a1dfc8c2843aa11d05ca8aa7; the upcoming
Error Prone release includes a workaround for this:
google/error-prone@df033f03cb55c3f26221660ac5ea49bbaeeab289.
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,10 +22,11 @@ jobs:
             jdk: 17.0.4
             distribution: temurin
             experimental: false
-          - os: ubuntu-22.04
-            jdk: 20-ea
-            distribution: zulu
-            experimental: true
+          # XXX: Reenable this build after upgrading to Error Prone 2.17.
+          #- os: ubuntu-22.04
+          #  jdk: 20-ea
+          #  distribution: zulu
+          #  experimental: true
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.experimental }}
     steps:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,7 +22,7 @@ jobs:
             jdk: 17.0.4
             distribution: temurin
             experimental: false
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -22,10 +22,11 @@ jobs:
             jdk: 17.0.4
             distribution: temurin
             experimental: false
-          - os: ubuntu-22.04
-            jdk: 20-ea
-            distribution: zulu
-            experimental: true
+          # XXX: Reenable this build after upgrading to Error Prone 2.17.
```
rickie, line 25: Based on some googling I think "re-enable" is a lot more common to use than "reenable". WDYT?

rickie (approved):

.github/workflows/build.yaml
Stephan202: No strong opinion :upside_down_face:

Badbond: We can circumvent this oddity to use `Reactivate` instead as suggested [here](https://forum.wordreference.com/threads/re-enable-vs-reenable.1510386/post-7623614). :smile:

Badbond (commented):
LGTM. Maybe `s/fix/workaround` in the suggested commit message in line with the wording of the [commit](https://github.com/google/error-prone/commit/df033f03cb55c3f26221660ac5ea49bbaeeab289) in error prone?

Badbond (approved):
> Maybe `s/fix/workaround` in the suggested commit message in line with the wording of the [commit](https://github.com/google/error-prone/commit/df033f03cb55c3f26221660ac5ea49bbaeeab289) in error prone?

Changed it. Tweaks LGTM. Approving assuming build passes :heavy_check_mark:

rickie:
The enable terminology is used more often I would say, so went with the hyphen option :).

Issue #340, opened 2022-11-08
Drop redundant `Matchers.{allOf,anyOf}` containing a single `Matcher`


oxkitsune:
### Problem
Declaring a `Matcher` using `Matchers.{allOf,anyOf}(Matcher...)` containing  just a single `Matcher`, is redundant.

### Description of the proposed new feature

- [ ] Support a stylistic preference.
- [x] Avoid a common gotcha, or potential problem.

We can rewrite
```java
allOf(annotations(AT_LEAST_ONE, isType("org.junit.jupiter.params.provider.MethodSource")));
anyOf(annotations(AT_LEAST_ONE, isType("org.junit.jupiter.params.provider.MethodSource")));
```

to:
```java
annotations(AT_LEAST_ONE, isType("org.junit.jupiter.params.provider.MethodSource"));
annotations(AT_LEAST_ONE, isType("org.junit.jupiter.params.provider.MethodSource"));
```

### Considerations

This should support the method overloads `Matchers.{allOf,anyOf}(Iterable<? extends Matcher>)`.

### Participation

This is a nice opportunity to get your hands dirty with `BugChecker`s :rocket:

rickie:
A related example of what this check would flag: https://github.com/PicnicSupermarket/error-prone-support/pull/335#discussion_r1016351033.

EvgheniiShipilov:
I'd like to pick this one up. Anything I should be aware of?

rickie:
> I'd like to pick this one up. 

Nice! 

Ah I think you found the most similar `RefasterAnyOfUsage` check already. Another one would be the `IdentityConversion` check. As that check also flags enclosing method invocations that can be dropped. 

Well, actually now that I'm thinking of that and looking into the tests of that check; maybe this would be as simple as adding these exact methods to `IS_CONVERSION_METHOD` 🤔. 

(Not sure if the naming would still be correct for `IS_CONVERSION_METHOD` but that's another concern).

Currently not able to try it out, but might be later / tomorrow.

rickie:
Sorry, a little later than expected, but I tried it out.

Indeed, we can use the `IS_CONVERSION_METHOD` and the `IdentityConversion` check to solve this problem. 

With this information, do you have enough context to proceed with opening a PR 😉?

EvgheniiShipilov:
I think this should be enough, thanks for looking into it!

PR #420, opened 2022-12-19
Have `IdentityConversion` flag `com.google.errorprone.matchers.Matchers#{allOf,anyOf}`

EvgheniiShipilov:
Check for cases when `allOf` and `anyOf` Matchers have a single parameter. Such cases can be rewritten by simply omitting the respective method.

* Solves https://github.com/PicnicSupermarket/error-prone-support/issues/340

Suggested commit message:
```
Drop redundant Matchers.{allOf,anyOf} containing a single Matcher
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IdentityConversion.java
```diff
@@ -70,7 +70,10 @@ public final class IdentityConversion extends BugChecker implements MethodInvoca
           staticMethod()
               .onClass("reactor.core.publisher.Flux")
               .namedAnyOf("concat", "firstWithSignal", "from", "merge"),
-          staticMethod().onClass("reactor.core.publisher.Mono").namedAnyOf("from", "fromDirect"));
+          staticMethod().onClass("reactor.core.publisher.Mono").namedAnyOf("from", "fromDirect"),
+          staticMethod()
+              .onClass("com.google.errorprone.matchers.Matchers")
```
rickie: We sort this list lexicographical, so we should put this one after the `"com.google.common"` ones. Moving it up a bit :).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/IdentityConversionTest.java
```diff
@@ -157,6 +163,11 @@ void identification() {
             "    Mono<Integer> m1 = Mono.from(Mono.just(1));",
             "    // BUG: Diagnostic contains:",
             "    Mono<Integer> m2 = Mono.fromDirect(Mono.just(1));",
+            "",
+            "    // BUG: Diagnostic contains:",
+            "    Matcher allOfMatcher = Matchers.allOf(annotations(AT_LEAST_ONE, isType(\"foo\")));",
```
rickie: Let's rename them to `matcher1` and `matcher2` to be in the "same style" as the other variable names.

rickie (approved):
Added a commit.

```
Have `IdentityConversion` flag `com.google.errorprone.matchers.Matchers#{allOf,anyOf}` (#420)
```

Thanks for opening a PR @EvgheniiShipilov , it looks good :rocket:!

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/IdentityConversionTest.java
```diff
@@ -157,6 +163,11 @@ void identification() {
             "    Mono<Integer> m1 = Mono.from(Mono.just(1));",
             "    // BUG: Diagnostic contains:",
             "    Mono<Integer> m2 = Mono.fromDirect(Mono.just(1));",
+            "",
+            "    // BUG: Diagnostic contains:",
+            "    Matcher matcher1 = Matchers.allOf(annotations(AT_LEAST_ONE, isType(\"foo\")));",
+            "    // BUG: Diagnostic contains:",
+            "    Matcher matcher2 = Matchers.anyOf(annotations(AT_LEAST_ONE, isType(\"bar\")));",
```
Stephan202: We can also move these up a bit, in line with the main code order. (Edit: seems there are other inconsistencies. Will move some stuff around.)

Next to that:
1. We can use simpler expressions for this test.
2. We should do this more generally, but since it's cheap for this case: let's also add some non-matching cases.

Stephan202 (approved):
Rebased and added a commit. Expanding on @rickie's suggested commit message:
```
Have `IdentityConversion` flag `com.google.errorprone.matchers.Matchers#{allOf,anyOf}` (#420)

While there, sort and rename some (test) code.

Fixes #340.
```

rickie:
Changes LGTM @Stephan202!

PR #426, opened 2022-12-28
Have `RequestParamType` ignore parameter types with custom deserialization support

chamil-prabodha:
One would want to use classes such as `ImmutableList` with `@RequestParam` given that the custom support for the underlying conversion is provided. For such occasions we can specify the classes we wish to be included on top of what is already provided (by Spring) by using the flag `-XepOpt:RequestParamType:Includes=<class-name>`

### DIFF
--- a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/RequestParamTypeTest.java
+++ b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/RequestParamTypeTest.java
@@ -17,6 +17,9 @@ final class RequestParamTypeTest {
           .setArgs(
               ImmutableList.of(
                   "-XepOpt:RequestParamType:Includes=com.google.common.collect.ImmutableSet"));
+  private final CompilationTestHelper restrictedWithInvalidFlagCompilationTestHelper =
+      CompilationTestHelper.newInstance(RequestParamType.class, getClass())
+          .setArgs(ImmutableList.of("-XepOpt:RequestParamType:Includes="));
 
   @Test
   void identification() {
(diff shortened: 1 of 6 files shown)


Stephan202 (approved):
Rebased and added a commit. Thanks for this PR!

Suggested commit message:
```
Have `RequestParamType` ignore parameter types with custom deserialization support (#426)

While there, introduce and use a new `Flags` utility class; various checks'
list flags now better support empty lists.
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/RequestParamTypeTest.java
```diff
@@ -6,6 +6,10 @@
 final class RequestParamTypeTest {
   private final CompilationTestHelper compilationTestHelper =
       CompilationTestHelper.newInstance(RequestParamType.class, getClass());
+  private final CompilationTestHelper restrictedCompilationTestHelper =
```
rickie, line 9: In Error Prone there are cases for both situations; but it can be nice to inline the `restrictedCompilationTestHelper` IMO. Usually there is only one test with the restricted variant and would make it extra clear in the test method itself that we are using extra flags. Besides that, the flag would be closer to the test, which makes the intend clearer as well. 

WDYT?

Stephan202: Yes, I've pondered this a few times as well. I'm fine with that, though it sounds like we should do that in a separate PR.

rickie: I'll file a PR for that then!

rickie: Filed https://github.com/PicnicSupermarket/error-prone-support/pull/442.

PR #439, opened 2023-01-04
Introduce `MapGetOrDefault` Refaster rule

### LINKED ISSUE #431: Rewrite `requireNonNullElse(map.get(k), defaultValue)` -> `map.getOrDefault(k, defaultValue)`
### Problem

As @Stephan202 mentioned in https://github.com/PicnicSupermarket/error-prone-support/issues/364, we can create a Refaster rule that rewrites the following:

`requireNonNullElse(map.get(k), defaultValue)` -> `map.getOrDefault(k, defaultValue)`.  

- [ ] Improve performance.

```java
Map<Integer, Integer> map = Map.of(1, 2); 
Integer integer = requireNonNullElse(map.get(1), 3);
```

to:
```java
Map<Integer, Integer> map = Map.of(1, 2);
Integer integer = map.getOrDefault(1, 3);
```

benhalasi:
adds rules to fix #431.

Added Rules:
-  prefer `map.getOrDefault(k, defaultValue)` over `requireNonNullElse(map.get(k), defaultValue)`

Wrote test inputs and outputs.

Could be a problem that when `defaultValue` is null, `Objects#requireNonNullElse` throws a NPE while `Map#getOrDefault` does not.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
@@ -45,10 +45,10 @@ final class MapRules {
     }
   }
 
-  /**
-   * Prefer {@link Map#getOrDefault(Object, Object)} over more contrived alternatives. Note that
-   * this method may return null instead of throwing a NPE.
-   */
+  /** Prefer {@link Map#getOrDefault(Object, Object)} over more contrived alternatives. */
+  // Note that
+  // XXX: `requireNonNullElse` throws an NPE if the second argument is `null`, while the alternative
+  // does not.
   static final class MapGetOrDefault<K, V, T> {
     @BeforeTemplate
     V before(Map<K, V> map, T key, V defaultValue) {
@@ -56,7 +56,6 @@ final class MapRules {
     }
 
     @AfterTemplate
-    @Nullable
     V after(Map<K, V> map, T key, V defaultValue) {
       return map.getOrDefault(key, defaultValue);
     }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
@@ -46,9 +46,8 @@ final class MapRules {
   }
 
   /** Prefer {@link Map#getOrDefault(Object, Object)} over more contrived alternatives. */
-  // Note that
-  // XXX: `requireNonNullElse` throws an NPE if the second argument is `null`, while the alternative
-  // does not.
+  // XXX: Note that `requireNonNullElse` throws an NPE if the second argument is `null`, while the
+  // alternative does not.
   static final class MapGetOrDefault<K, V, T> {
     @BeforeTemplate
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
```diff
@@ -43,6 +45,23 @@ V after(Map<K, V> map, T key) {
     }
   }
 
+  /**
+   * Prefer {@link Map#getOrDefault(Object, Object)} over more contrived alternatives. Note that
+   * this method may return null instead of throwing a NPE.
```
Stephan202: Good point! (I'll move this to an `XXX` comment so that we can find it back more easily later; we have some plans around calling this sort of thing out more prominently.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
```diff
@@ -43,6 +45,23 @@ V after(Map<K, V> map, T key) {
...
+   */
+  static final class MapGetOrDefault<K, V, T> {
+    @BeforeTemplate
+    V before(Map<K, V> map, T key, V defaultValue) {
+      return requireNonNullElse(map.get(key), defaultValue);
+    }
+
+    @AfterTemplate
+    @Nullable
```
Stephan202: It's only `@Nullable` if the input is; in this case we didn't annotate the params as such.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/MapRulesTestInput.java
```diff
@@ -11,7 +13,7 @@
 final class MapRulesTest implements RefasterRuleCollectionTestCase {
   @Override
   public ImmutableSet<?> elidedTypesAndStaticImports() {
-    return ImmutableSet.of(HashMap.class);
+    return ImmutableSet.of(HashMap.class, requireNonNullElse("foo", "bar"));
```
Stephan202: I notice that in `elidedTypesAndStaticImports` we use `null` for non-primitive parameters (the code isn't executed anyway), so let's stick with that.

(That said, perhaps we _should_ find a way to run these files, and enforce more realistic test cases that way... something to ponder.)

Stephan202 (approved):
Tnx for your contribution!

Suggested commit message:
```
Introduce `MapGetOrDefault` Refaster rule (#439)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapRules.java
rickie: Hmm, I'm not completely agreeing with this though. Sometimes we do this in a separate `<p>` sentence in the Javadoc. Will leave it for now though, apart from a minor fix :).

rickie (approved):
Added a commit and rebased.
Made a tweak to the suggested commit message :).

I'm afraid we should wait a bit more with merging this one. I already started testing the current `master` on downstream repositories, so would probably be better to include this in next release :).

Thanks for _another_ contribution @benhalasi 🚀 !

PR #441, opened 2023-01-04
Have `apply-error-prone-suggestions.sh` download JitPack-hosted artifacts

Stephan202:
Suggested commit message:
```
Have `apply-error-prone-suggestions.sh` download JitPack-hosted artifacts (#441)

While there, tweak the usage message of both `apply-error-prone-suggestions.sh`
and `run-mutation-tests.sh`.
```

### DIFF
--- a/apply-error-prone-suggestions.sh
+++ b/apply-error-prone-suggestions.sh
@@ -9,13 +9,14 @@
 set -e -u -o pipefail
 
 if [ "${#}" -gt 1 ]; then
-  echo "Usage: ./$(basename "${0}") [PatchChecks]"
+  echo "Usage: ${0} [PatchChecks]"
   exit 1
 fi
 
 patchChecks=${1:-}
 
 mvn clean test-compile fmt:format \
+  -s "$(dirname "${0}")/settings.xml" \
   -T 1.0C \
   -Perror-prone \
   -Perror-prone-fork \
--- a/run-mutation-tests.sh
+++ b/run-mutation-tests.sh
@@ -7,7 +7,7 @@
 set -e -u -o pipefail
 
 if [ "${#}" -gt 1 ]; then
-  echo "Usage: ./$(basename "${0}") [TargetTests]"
+  echo "Usage: ${0} [TargetTests]"
   exit 1
 fi
 


apply-error-prone-suggestions.sh
```diff
@@ -9,13 +9,14 @@
 set -e -u -o pipefail
 
 if [ "${#}" -gt 1 ]; then
-  echo "Usage: ./$(basename "${0}") [PatchChecks]"
+  echo "Usage: ${0} [PatchChecks]"
```
Stephan202, line 12: We no longer assume that the script is executed from the directory in which it resides.

apply-error-prone-suggestions.sh
```diff
@@ -9,13 +9,14 @@
...
   exit 1
 fi
 
 patchChecks=${1:-}
 
 mvn clean test-compile fmt:format \
+  -s "$(dirname "${0}")/settings.xml" \
```
Stephan202, line 19: This fix has downsides of its own: it overrides the user's `settings.xml`, which may contain other useful or even required settings. For Picnic devs it means that our Nexus deployment is bypassed, triggering one extra round of downloads.

The alternative is to move the `settings.xml` contents into a profile in the top-level `pom.xml`, but I seem to recall that Maven Central doesn't like that. (Or perhaps is does allow it when part of a profile? I'd have to check.)

Stephan202 (commented):
So as indicated this is not a perfect fix, but it's the easiest thing I can do on short notice. (Modifying `pom.xml` isn't harder, but then there are more cases I'd have to reason about, including any possible impact on the next release to Maven Central.) As this is an improvement regardless, I think it's an acceptable change.

CC @benhalasi in case you want to try it :)

rickie (approved):
Thanks for the quick fix @Stephan202 ! 

Would indeed be nice if @benhalasi is up for checking if with these changes it'd work :smile:.

Verified that it still works for me :).

apply-error-prone-suggestions.sh
Badbond: Based on [Setting up multiple repositories](https://maven.apache.org/guides/mini/guide-multiple-repositories.html) (esp. the order) and [Using mirrors for repositories](https://maven.apache.org/guides/mini/guide-mirror-settings.html); as Picnic devs mirror everything anyway, could we instead just place `repositories` in the root of the pom (i.e. not in a profile)? Then we can also omit the `settings.xml` altogether.

To my understanding, this would behave the same as that the super pom [defines](https://maven.apache.org/ref/3.0.4/maven-model-builder/super-pom.html) a repository to Maven central, which is always included too.

Badbond (approved):
LGTM. An approvement nonetheless indeed.

benhalasi:
I've checked and works well for me too. @rickie

rickie:
> I've checked and works well for me too. @rickie

Thanks for verifying @benhalasi !

PR #443, opened 2023-01-04
Introduce BugPattern for removing duplicate `StepVerifier#expectNext` calls

SimonBaars:
Woowee! That was a freaking mess.

At first I thought I could easily solve this with a little Refaster pattern. Boy was I wrong. So many edge cases, some many little thingies. So I went down the Error Prone rabbit hole, and boi was it deep. Who knew simply traversing nodes in an AST could be such a mess? Hell, even internal functions of Error Prone (yes I'm talking about you `ASTHelpers.findEnclosingNode`) have infinite while loops in them. BWAAAAH!

Anyway, it's 1AM in the night and I think I have a fully working version. As in, I made all the tests I could think of and they pass. Now it's up to the Lvl. 200 mafia bosses @rickie and crew to poke some holes in this dumpster fire of a bugpattern.

To explain a little bit what is going on:
I do a little for-loop where I step to all the other method invocations. The smart logic here is `getChild()`, which jumps `n` steps over the next few nodes (child nodes) to find the next `expectNext` call. By jumping 2 steps at a time, we visit all the method calls. Other types of nodes that it may visit should be eliminated either by the matcher or by the position check.

Also I miss you 😢😢😢😢😢😢❤️

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StepVerifierDuplicateExpectNext.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StepVerifierDuplicateExpectNext.java
@@ -20,6 +20,8 @@ import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MemberSelectTree;
 import com.sun.source.tree.MethodInvocationTree;
 import com.sun.tools.javac.code.Type;
+
+import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -67,14 +69,17 @@ public final class StepVerifierDuplicateExpectNext extends BugChecker
       return Description.NO_MATCH;
     }
 
-    MemberSelectTree ms = (MemberSelectTree) tree.getMethodSelect();
-    MethodInvocationTree parent = (MethodInvocationTree) (ms).getExpression();
-    boolean matches = STEP_EXPECTNEXT.matches(parent, state);
-    if (!matches) {
+    MethodInvocationTree parent = tree;
+    List<ExpressionTree> args = new ArrayList<>();
+    while(STEP_EXPECTNEXT.matches(getParent(parent), state)){
+      parent = getParent(parent);
+      args.addAll(parent.getArguments());
+    }
+
+    if (args.isEmpty()) {
       return Description.NO_MATCH;
     }
 
-    List<? extends ExpressionTree> args = parent.getArguments();
     String newArgument =
         tree.getArguments().stream().map(Object::toString).collect(Collectors.joining(", "));
     SuggestedFix.Builder argumentsFix =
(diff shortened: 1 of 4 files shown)


SimonBaars:
Sometimes it helps to visually review the PR, so here's a little illustration of this PR's contents:
image

Stephan202:
Nice parting gift @SimonBaars! :smile:

W.r.t. the build failures: Running `./apply-error-prone-suggestions.sh` (with #441) should (largely) take care of those. :sparkles:

Shame that the various `expectNext` overloads aren't supported by Refaster's `@Repeated` annotation; we have a similar issue for `ImmutableList.of` and friends. Thinking of all the combinations that'd be necessary, I agree that a `BugChecker` is indeed the way to go.

I'll let Lvl. boss @rickie do the first review. :rofl:

rickie:
Nice job @SimonBaars! It was fun to pair up on this a few times! (Still sad to see you go man 😢) 

Currently review backlog is a bit full, so it'll probably be next week before I can do a proper deep-dive 😉.

SimonBaars:
> Currently review backlog is a bit full, so it'll probably be next week before I can do a proper deep-dive wink.

You're beginning to turn into a manager!

SimonBaars:
@rickie 
image

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StepVerifierDuplicateExpectNext.java
```diff
@@ -0,0 +1,115 @@
...
+      newArgs.addAll(child.getArguments());
+    }
+
+    if (newArgs.isEmpty()) {
+      return Description.NO_MATCH;
+    }
+
+    String newArgument = newArgs.stream().map(Object::toString).collect(joining(", "));
+    List<? extends ExpressionTree> myArgs = tree.getArguments();
+    SuggestedFix.Builder argumentsFix =
```
rickie, line 87: `matchMethodInvocation` method is maybe a bit too full. Already made a comment about the middle part. Maybe we can extract the `Description` related things to a method like we do in some other places. Look for the regex: `try.*\(` to see some methods that return either a `Fix` or `Description` for some examples.

rickie (changes requested):
Nice meme haha, I did an initial review where I first skimmed over the whole setup and provided feedback. Sorry for not getting back earlier to this. The Error Prone Support review backlog is quite big 😅. 

Couldn't apply all feedback myself so leaving some open points for you as well 😉. Hope you don't mind :). 

There are some interesting concepts in this check! Cool stuff!

SimonBaars:
@rickie Hey, sorry, I'm not following this thread very actively.
What would you still like me to take a look at?

rickie:
There are some open points that I haven't addressed yet. If you want, you can take a look there 😄.

PR #450, opened 2023-01-06
Introduce BugPattern for disallowing specified static imports

cernat-catalin:
Introduce a rule that flags _bad_ static imports (complements the existing `StaticImport` rule).

`BadStaticImport` might not be the best name but it was in line with `com.google.errorprone.bugpatterns.BadImport`.

**Approach**
Initially I tried matching static imports and then finding the relevant identifiers but that seemed too complicated 🧶 . After many rewrites matching method invocations and identifiers seemed to work (and was simpler than expected) though I'm not sure if this covers all cases or if there are false positives. I'm also not sure what (if any) is the performance penalty for matching on identifiers.

**Static imports and bad static imports**
Since these two rules need to coexist they need to be mutually exclusive (otherwise bad things could happen like the order in which are applied would matter). As far as I can tell, there are multiple ways to do this depending on how much flexibility we want to allow when defining exceptions (i.e. exempted members).

I've tried to allow the rules to be as flexible as possible so the following should be possible to specify:
1. All identifiers from type X except A,B,C... should be statically imported
2. All identifiers from type X except A,B,C... should not be statically imported

The mutual exclusion is guaranteed by the tests. Consider the following notation:
- `T.*` Types that should be statically imported (defined in `StaticImport`)
- `T.ID` Type identifier pairs that should be statically imported (defined in `StaticImport`)
- `T'.*` Types that should not be statically imported (defined in `BadStaticImport`)
- `T'.ID'` Type identifier that should not be statically imported (defined in `BadStaticImport`)
- `*.ID'` Identifiers that should not be statically imported (defined in `BadStaticImport`)

In `StaticImport` we allow all type identifier pairs from this set: `(T.* \ T'.ID' \ *.ID') U T.ID`
In `BadStaticImport` we allow all type identifier pairs from this set: `(T'.* \ T.ID) U T'.ID' U *.ID'`
In the tests we also enforce the following which should guarantee the mutual exclusion of the above defined sets (`^` is intersection):
- `T.* ^ T'.* = empty`
- `T.ID ^ T'.ID' = empty`
- `T.ID ^ *.ID' = empty`

There are still some comments to resolve plus candidates and tests to be added (will get to it in some subsequent commits).


rickie:
Hey @cernat-catalin, thanks for opening the PR 🚀 ! We have a GitHub Action that runs Pitest on the changes of the PR. This helps identifying edge cases that are not tested. Can you look at the comments the bot placed on the changes in the [Files changed](https://github.com/PicnicSupermarket/error-prone-support/pull/450/files) tab? If you need help interpreting the results, let me know!

cernat-catalin:
@rickie Pitest is lit! It's the first time I encounter it but it was really helpful.

There still are three mutations on some conditionals. I could remove them (and suppress a warning generated by nullaway) but I have a feeling that they are useful however I can't find any examples 😞 .

Stephan202:
I might wait for @rickie to do a first review round (I'm quite swamped atm), but let me say: _great PR description!_ :heart:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/BadStaticImport.java
```diff
@@ -0,0 +1,194 @@
...
+      return getDescription(tree, state);
+    }
+
+    return Description.NO_MATCH;
+  }
+
+  private Description getDescription(IdentifierTree tree, VisitorState state) {
+    Symbol symbol = ASTHelpers.getSymbol(tree);
+    SuggestedFix.Builder builder =
+        SuggestedFix.builder().removeStaticImport(getImportToRemove(symbol));
```
rickie: Hmm, I'm curious if we can always do this 🤔. Removing an import _can_ be dangerous. This is usually a task for the `RemoveUnusedImport` check in a subsequent run. 

However, I'm thinking that we can actually get away with this one. What if there are two identifiers statically imported that we update, is there a possibility that one of the two _cannot_ be imported 🤔. If there is a possibility that can happen, we cannot remove the static import.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/BadStaticImport.java
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  @SuppressWarnings("NullAway")
+  private static String getImportToRemove(Symbol symbol) {
+    return String.join(
+        ".", symbol.getEnclosingElement().getQualifiedName(), symbol.getSimpleName());
+  }
+
+  private static boolean isMatch(IdentifierTree tree, VisitorState state) {
+    Symbol symbol = ASTHelpers.getSymbol(tree);
```
rickie: We use this method here and in the `getDescription`. We can probably start with that as we are more interested in that than in the Tree itself. Saves us an extra if statement.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/BadStaticImport.java
```diff
@@ -0,0 +1,194 @@
...
+    return describeMatch(tree, builder.build());
+  }
+
+  @SuppressWarnings("NullAway")
+  private static String getImportToRemove(Symbol symbol) {
+    return String.join(
+        ".", symbol.getEnclosingElement().getQualifiedName(), symbol.getSimpleName());
+  }
+
+  private static boolean isMatch(IdentifierTree tree, VisitorState state) {
```
rickie: Hmm `isMatch` should have a more descriptive name. Will check this out later. Or feel free to propose something.

rickie (changes requested):
Added two commits. 

Till now I have mostly placed comments and improvements on the style. I needed to do a round of that first before I can dive in a little more. I took some time to get a full picture of the check. 

So will have to do another round some other time. Already left some comments. Let me know if you have any questions :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
```diff
@@ -0,0 +1,194 @@
...
+        builder
+            .removeStaticImport(getImportToRemove(symbol))
+            .prefixWith(tree, replacement)
+            .build());
+  }
+
+  private static String getImportToRemove(Symbol symbol) {
+    return String.join(
+        ".",
+        requireNonNull(symbol.getEnclosingElement()).getQualifiedName(),
```
rickie, line 139:
```suggestion
       symbol.owner.getQualifiedName(),
```
Now that I think of it, we can probably just use `owner` directly. That also makes me wonder if we should call it "enclosingSymbol" further down, that's probably not the correct term 🤔.

rickie (commented):
Another thought.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/BadStaticImportTest.java
```diff
@@ -0,0 +1,198 @@
...
+            "    Object o1 = empty();",
+            "",
+            "    Object l1 = copyOf(ImmutableList.of());",
+            "    Object l2 = ImmutableList.copyOf(ImmutableList.of());",
+            "",
+            "    Locale lo1 = ROOT;",
+            "    Locale lo2 = Locale.ROOT;",
+            "",
+            "    Object c1 = MIN;",
+            "    Object c2 = MAX;",
```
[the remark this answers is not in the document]
cernat-catalin: There was no particular reason to be honest (just easier 😄 )

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/BadStaticImport.java
```diff
@@ -0,0 +1,194 @@
...
+import tech.picnic.errorprone.bugpatterns.util.SourceCode;
+
+/** A {@link BugChecker} that flags methods and constants that should not be statically imported. */
+// XXX: Also introduce checks that disallows the following candidates:
+// - `ZoneOffset.ofHours` and other `ofXXX`-style methods.
+// - `java.time.Clock`.
+// - Several other `java.time` classes.
+// - Likely any of `*.{ZERO, ONE, MIX, MAX, MIN_VALUE, MAX_VALUE}`.
+@AutoService(BugChecker.class)
+@BugPattern(
```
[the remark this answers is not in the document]
cernat-catalin: I think it's fine! As mentioned, I was really unsure about the name. But I agree that `Bad` does make it sound a bit too harsh.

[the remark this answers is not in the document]
cernat-catalin: Now I'm thinking if we should also rename `BAD_STATIC_IMPORT_CANDIDATE_MEMBERS`, etc.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/BadStaticImport.java
```diff
@@ -0,0 +1,194 @@
...
+    summary = "Identifier should not be statically imported",
+    link = BUG_PATTERNS_BASE_URL + "BadStaticImport",
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = SIMPLIFICATION)
```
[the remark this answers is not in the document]
cernat-catalin: Agree! But this makes me think, doesn't the same logic follow for `StaticImport`?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/BadStaticImport.java
cernat-catalin: Good question. I'm thinking the following:

Say we have an identifier `X` that with two occurrences (or more) in the same file. IIUC, the situation we want to avoid is the `isMatch` method returning true in one place and false in another.

However, for that to happen, one of the following must return different results for the occurrences:
- `symbol.owner != TYP` Can't have different owners
- `isDefinedInThisFile(...)` They can't be defined in different files
- `isIdentifierStaticallyImported(...)` Since it is the same identifier we will have the same result
- `isExempted` and `isCandidate` Will return the same results

I guess it boils down to the fact that we never use the surrounding context of the identifier to make a decision. WDYT?

cernat-catalin:
Rebased, added some more identifiers and 🔫 all of the surviving mutants (with a somewhat weird example).

cernat-catalin:
@rickie @Stephan202 Is this something that we still want?

rickie:
> @rickie @Stephan202 Is this something that we still want?

Hi @cernat-catalin, yes **definitely**. Indeed, there is not a lot of activity from our side on this PR lately. There are a few (relatively) big PRs open in EPS that require quite some time + deep focus to review. The fact we didn't review recently does _not_ mean the PR is not good.

We are aware of the PR andit  will be reviewed, given our schedules it's hard to give an estimation. All I can say is that, this still a really nice PR and we definitely didn't forget about it.

cernat-catalin:
Thanks for the answer @rickie ! No rush from my side and I understand there is a lot of work to be done and not that many people that can do it. Glad to see that it is still on your radar!

cernat-catalin:
Thanks for the answer @rickie ! No rush from my side and I understand there is a lot of work to be done and not that many people that can do it. Glad to see that it is still on your radar!

rickie:
Rebased :).

rickie (commented):
Had another look at the code 😄, it's not an easy PR, has a lot of edge cases to consider. 

Cool stuff 🚀 !

Stephan202:

Stephan202 (commented):
Rebased and added a commit. The new approach is quite different from the previous approach, so please carefully review.

From my side there's one key open point: reviewing the test code. Hope to have some time for that "soon".

Stephan202 (approved):
Rebased and added one more commit.

Suggested commit message:
```
Introduce `NonStaticImport` check (#450)

This check complements the existing `StaticImport` check. We ensure that the 
two checks do not suggest mutually incompatible approaches to static imports. 
```

Thanks for your patience with this one, @cernat-catalin!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
```diff
@@ -0,0 +1,212 @@
...
+  private static boolean shouldNotBeStaticallyImported(String type, String member) {
+    return (NON_STATIC_IMPORT_CANDIDATE_TYPES.contains(type)
+            && !STATIC_IMPORT_CANDIDATE_MEMBERS.containsEntry(type, member))
+        || NON_STATIC_IMPORT_CANDIDATE_MEMBERS.containsEntry(type, member)
+        || NON_STATIC_IMPORT_CANDIDATE_IDENTIFIERS.contains(member);
+  }
+
+  private static void replaceUndesiredStaticImportUsages(
+      CompilationUnitTree tree,
+      VisitorState state,
```
rickie, line 185: `VisitorState` should always be at the end.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
```diff
@@ -0,0 +1,212 @@
...
+  private static ImmutableTable<String, String, UndesiredStaticImport> getUndesiredStaticImports(
+      CompilationUnitTree tree, VisitorState state) {
+    ImmutableTable.Builder<String, String, UndesiredStaticImport> imports =
+        ImmutableTable.builder();
+    for (ImportTree importTree : tree.getImports()) {
+      Tree qualifiedIdentifier = importTree.getQualifiedIdentifier();
+      if (importTree.isStatic() && qualifiedIdentifier instanceof MemberSelectTree) {
+        MemberSelectTree memberSelectTree = (MemberSelectTree) qualifiedIdentifier;
+        String type = SourceCode.treeToString(memberSelectTree.getExpression(), state);
+        String member = memberSelectTree.getIdentifier().toString();
```
rickie, line 162: Maybe `memberIdent` or `memberIdentifier`? That makes it a tiny bit clearer?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
```diff
@@ -0,0 +1,212 @@
...
+              "fill",
+              "list",
+              "max",
+              "min",
+              "nCopies",
+              "rotate",
+              "sort",
+              "swap")
+          .put("java.util.Locale", "ROOT")
+          .put("java.util.Optional", "empty")
```
rickie, line 95: We drop this one and the `Filters#empty` and add an entry to `NON_STATIC_IMPORT_CANDIDATE_IDENTIFIERS` with `empty`.

rickie (approved):
Code looks really nice! Reporting on the import statements is actually quite cool 🚀 !

Nice that we can get this one across the finish line 😄.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
Stephan202: It's a `member` of the aforementioned `type`. That it's an identifier is apparent from the `String` type. This also matches the `shouldNotBeStaticallyImported` parameter names. No _very_ strong opinion, but I think that `member` was a better name than `memberIdentifier`.

Stephan202:
(Other changes LGTM :).)

cernat-catalin:
Nice to see this one being pushed! I kept postponing to look over it when I had more time.

Thanks @Stephan202 and @rickie for the changes!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
rickie: Will revert.

rickie:
Rebased and added a commit with the revert, will merge once 🟢 !

PR #454, opened 2023-01-07
Introduce `MockitoMockClassReference` check

Badbond:
~~_Opening PR as draft already to get some feedback on the best approach for the unimplemented test cases regarding type arguments and generics (see corresponding test)._~~

---

Suggested commit message:

```
Introduce `MockitoMockClassReference` check (#454)

Flags Mockito mock and spy creation expressions that explicitly specify the
type of mock or spy to create, while this information can also be inferred from
context.
```

---

### Description

Adds a bug check that flags usages of `Mockito#mock(class)` and `Mockito#spy(Class)` which can be replaced with the no-arg overload, as introduced in Mockito 4.10.0  (https://github.com/mockito/mockito/pull/2779). This is possible when the type this mock is assigned to is known and equal to the class passed to this method. 

Some simple cases:

```diff
-    Foo foo = mock(Foo.class);
+    Foo foo = mock(); 
```

```diff
  Foo getFoo() {
-    return mock(Foo.class);
+    return mock(); 
  }
```

### Edge-cases

**Subtypes**

We don't want to replace cases where the class passed is a subtype of the class that the mock is assigned to. For example (where `Foo extends|implements Bar`):

```java
Bar bar = mock(Foo.class);
assertThat(bar).isInstanceOf(Foo.class);
```

Changing this would otherwise make the assertion fail.

**Classes with generic type arguments**

It is safe to mock classes with generic type parameters (e.g. `List<E>`) as the mock should be programmed either way based on the type of the type parameter. Therefore these are currently rewritten. For example:

```diff
-    List<Foo> foos = mock(List.class);
-    List bars = mock(List.class);
+    List<Foo> foos = mock(); 
+    List bars = mock();
```

**Variably typed mocks**

When the type of the mock is variable, especially when encapsulated in a method parameter, there can be problems when removing the argument. Therefore, currently, these cases are not supported. A simple case that _could_ be rewritten is:

```diff
-Class<BigInteger> clazz = BigInteger.class;
-BigInteger mock = mock(clazz);
+BigInteger mock = mock();
}
```

However, when the variable is dynamic, it could lead to problems. For example:

```diff
-<T> T getMock(class<T> clazz) {
-  return mock(clazz);
+<T> T getMock() {
+  return mock();
}


@Test
void mockType() {
-  assertThat(getMock(BigInteger.class)).isInstanceOf(BigInteger.class);
+  assertThat((BigInteger) getMock()).isInstanceOf(BigInteger.class);
}
```

**Non-explicitly typed mocks**

When the mock is assigned to a non-explicitly typed variable (i.e. when using `var`), we can not simply drop the mock argument as otherwise Mockito could not derive its type. Thus we will not rewrite:

```java
var foo = mock(Foo.class);
```


error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReferenceTest.java
```diff
@@ -0,0 +1,139 @@
...
+            "    List nonGenericallyTypedSpy = spy(List.class);",
+            "    var dynamicallyTypedSpy = spy(Integer.class);",
+            "    Object objectSpy = spy(new Object());",
+            "  }",
+            "}")
+        .doTest();
+  }
+
+  @Test
+  void unimplementedCases() {
```
Badbond, line 53: I'd like some input as to how we wish to behave in these edge-cases and what would be the ideal way to detect those cases.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReferenceTest.java
```diff
@@ -0,0 +1,139 @@
...
+            "    // The real type of `variableTypedMock` is `BigInteger`.",
+            "    // But when replaced, it will be `Number`.",
+            "    Class<? extends Number> variableType = BigInteger.class;",
+            "    Number variableTypedMock = mock(variableType);",
+            "  }",
+            "",
+            "  <T extends Number> T getGenericMock(Class<T> clazz) {",
+            "    // No idea how to detect type T as part of generics usage.",
+            "    return mock(clazz);",
+            "  }",
```
Badbond, line 79: Coming to think of it, if it were not such a complex case where `T extends Number`, but where it was just `T`, then we could likely rewrite it:

```diff
  <T> T getGenericMock(Class<T> clazz) {
-    return mock(clazz);
+    return mock();
  }
```

Perhaps this case is just similar as `variableTypedMock` above but instead with a generic result type. So we might be able to simplify these test cases a bit and indeed follow the `ASTHelpers.isSameType` approach where it compared the `erasure` of both types.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReference.java
```diff
@@ -0,0 +1,76 @@
...
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<MethodInvocationTree> SUPPORTED_MOCKITO_METHOD_INVOCATION =
+      allOf(
+          argumentCount(1),
+          argument(0, isSameType(Class.class.getName())),
+          staticMethod().onClass("org.mockito.Mockito").namedAnyOf("mock", "spy"));
+  // XXX: Replace `var` usage with explicit type instead.
+  private static final Matcher<VariableTree> INCOMPATIBLE_VARIABLE_TREE =
+      anyOf(ASTHelpers::hasNoExplicitType, MockitoMockClassReference::hasTypeDifference);
```
Badbond, line 51: As discussed with @Stephan202 offline. It might be cool to also rewrite the following:

```diff
-    var foo = mock(Foo.class);
+    Foo foo = mock();
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReferenceTest.java
```diff
@@ -0,0 +1,139 @@
...
+            "import static org.mockito.Mockito.mock;",
+            "",
+            "import java.math.BigInteger;",
+            "import java.util.List;",
+            "",
+            "class A {",
+            "  void m() {",
+            "    // This case is arguable as it is unsafe in any case.",
+            "    // This currently is not identified as the `erasure` is the same.",
+            "    List<String> genericallyTypedMock = mock(List.class);",
```
Badbond, line 67: As pointed out, this is an unsafe cast in both before and after case. As such, at least in our own codebases, we sometimes have casts with `SuppressWarnings`. If we are to rewrite these and we can circumvent these warnings, it might be nice to also remove existing suppressions.

Stephan202: Hmm, I would have expected dropping `List.class` to be safe here. And if we can drop a suppression in this case: great!

Stephan202 (commented):
Very nice @Badbond! I'll try to find some time to play with the code for proper advice; not sure I'll have time for that this weekend, though. But let's see :crossed_fingers:

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReferenceTest.java
rickie: I don't think we should rewrite this case, if we would, it'd require a suppression as you mention. That is not an improvement IMO, so I wouldn't bother for now.

rickie: The same holds for the `getGenericMock` so I'm not sure if we would gain anything from that. Also, if we apply that change, then the parameter would be unused 🤔. So all in all, not sure if it would be an improvement.

rickie (commented):
Added a commit with some suggestions. Also some opinions on open stuff :).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReferenceTest.java
Stephan202: I filed mockito/mockito#2866; with those changes dropping `List.class` would be fine.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReference.java
```diff
@@ -45,11 +51,13 @@ public final class MockitoMockClassReference extends BugChecker
   private static final Matcher<MethodInvocationTree> MOCKITO_MOCK_OR_SPY =
       allOf(
           argumentCount(1),
-          argument(0, isSameType(Class.class.getName())),
+          argument(0, allOf(isSameType(Class.class.getName()), not(isVariable()))),
```
Badbond, line 54: There can be caveats when the passed argument is a variable that can lead into non-compilable/working-with-Mockito code, e.g. the following is odd and won't work:

```diff
-<T> T getMock(class<T> clazz) {
-  return mock(clazz);
+<T> T getMock() {
+  return mock();
}


@Test
void mockType() {
-  assertThat(getMock(BigInteger.class)).isInstanceOf(BigInteger.class);
+  assertThat((BigInteger) getMock()).isInstanceOf(BigInteger.class);
}
```

There might be cases where we could still replace the variable, but I couldn't find an elegant way for it without going into the caveats. For example the trivial:

```diff
-Class<BigInteger> clazz = BigInteger.class;
-BigInteger mock = mock(clazz);
+BigInteger mock = mock();
}
```

Feel free to play around and see if you find an elegant way.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReference.java
```diff
@@ -60,18 +68,28 @@ public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState
...
   }
 
   private static boolean hasTypeDifference(VariableTree tree, VisitorState state) {
-    return !ASTHelpers.isSameType(
-        ASTHelpers.getType(tree), ASTHelpers.getType(tree.getInitializer()), state);
+    return hasTypeDifference(tree, tree.getInitializer(), state);
+  }
+
+  private static boolean hasTypeDifference(ReturnTree tree, VisitorState state) {
+    Tree returnTypeTree = requireNonNull(state.findEnclosing(MethodTree.class)).getReturnType();
```
Badbond, line 88: This `requireNonNull` is safe because passed is a `ReturnTree`. Alternatively we could have passed the matching `MethodInvocationTree` (to prevent having to do `tree.getExpression` again), but then finding the enclosing `MethodTree` is less 'safe' in case this method gets reused. I found this to be cleaner and clearer given the similar method signatures.

Furthermore, inlining as a lambda expression instead was [flagged by ErrorProne](https://errorprone.info/bugpattern/UnnecessaryLambda), and indeed I do find this to read better.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReference.java
Badbond: Didn't have the brain juice to go after this as it consists of multiple fixes. Feel free to have a crack at it. If someone wants to pair up, I'm open to that too. Otherwise, I'd favor leaving the XXX and revisiting this at a later time (and creating a GH issue for it).

Badbond (commented):
I am happy with the current state of the PR to continue its review :+1:. Thanks for filing the Mockito PR upstream @Stephan202.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReference.java
```diff
@@ -0,0 +1,95 @@
...
+    Tree parent = state.getPath().getParentPath().getLeaf();
+    if (parent.getKind() == VARIABLE
+        && INCOMPATIBLE_VARIABLE.matches((VariableTree) parent, state)) {
+      return Description.NO_MATCH;
+    } else if (parent.getKind() == RETURN
+        && INCOMPATIBLE_RETURN.matches((ReturnTree) parent, state)) {
+      return Description.NO_MATCH;
+    }
+
+    return describeMatch(tree, SuggestedFix.delete(Iterables.getOnlyElement(tree.getArguments())));
```
Stephan202, line 80: The code now assumes a match in case the parent is not a `VARIABLE` or `RETURN` statement. So the following statement is also (incorrectly) flagged:
```java
Objects.hash(mock(Integer.class));
```

Stephan202 (commented):
Had a quick look; I don't mind proposing a fix, but @Badbond if you wan to have a crack at it yourself, lemme know :)

Badbond:
> Had a quick look; I don't mind proposing a fix, but @Badbond if you wan to have a crack at it yourself, lemme know :)

Feel free to!

Stephan202:
On the list it goes :smile:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReference.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  private static boolean hasTypeDifference(VariableTree tree, VisitorState state) {
```
rickie, line 83: After adding some random test cases I found some more examples / cases that should be covered. I'll push a change where this is addressed.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReference.java
rickie: After some rewriting and trying stuff out I found out we are strictly speaking too many cases. So the example in my previous comment is the `Kind.NEW_CLASS` edge case. We should only rewrite things if we are sure it will be correct, so swapping the default here.

While doing so, I changed the setup to not use `Matcher`s and have multiple `hasTypeDifference` methods. PTAL and let me know what you think.

rickie (commented):
Did a rewrite based on some (open) comments and edge cases that I spotted 😄. Let me know what you think. 

Added a commit and rebased.

Stephan202 (commented):
Rebased and added a commit. Didn't finish reviewing yet :)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReferenceTest.java
```diff
@@ -0,0 +1,165 @@
...
+            "",
+            "  Number getSubTypeMock() {",
+            "    return mock(Integer.class);",
+            "  }",
+            "}")
+        .doTest();
+  }
+
+  @Test
+  void replacement() {
```
Stephan202, line 98: We can simplify this test, since this fix isn't (so) context dependent.

Stephan202 (approved):
Added one more commit. Also tweaked the suggested commit messages, as we now cover more methods. Eager to see this check in use :smile:

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReferenceTest.java
```diff
@@ -13,17 +13,13 @@ void identification() {
             "A.java",
             "import static org.mockito.Mockito.mock;",
             "import static org.mockito.Mockito.spy;",
+            "import static org.mockito.Mockito.withSettings;",
```
rickie, line 16: Note to self: always check for more edge cases that we can test 😛.

rickie (approved):
Do we want to create an XXX to identify when we can drop a preceding suppression like [here](https://github.com/rickie/guava/pull/1/files#diff-efbfe355206c155fdfd4075918fc57349052466c6c84c6a40b60eff2db260c90R210)?

(🔝 not a blocker, so approving)

Haha, passing the `tree` instead of calling either the `getInitializer` or `getExpression` with cast is big brain move 😄. 

Nice improvements and very nice check @Badbond 🚀 !

Stephan202:
> Do we want to create an XXX to identify when we can drop a preceding suppression like [here](https://github.com/rickie/guava/pull/1/files#diff-efbfe355206c155fdfd4075918fc57349052466c6c84c6a40b60eff2db260c90R210)?

Good one. For this PR we should at least (manually?) clean up the impacted files. I see three options:
1. Extend this check to drop `@SuppressWarnigs("unchecked")` when making this change. Easiest, but rather specific.
2. Introduce a separate (ideally non-Mockito-specifc) check that identifies such unnecessary suppressions. Complexity TBD, but more in line with our "better to compose minimal improvements" philosophy.
3. Introduce an off-by-default shell script or `compilesWithFix` bug checker that flags an type of obsolete suppression. Complexity TBD.

> Haha, passing the `tree` instead of calling either the `getInitializer` or `getExpression` with cast is big brain move 😄. 

Tnx ;)

Stephan202:
> For this PR we should at least (manually?) clean up the impacted files.

Didn't notice that the diff referenced another repo :upside_down_face:. Anyway, options remain the same.

rickie:
Added a commit with the XXX. 

I'd like to propose the following. Create a GH issue for (1) and keep (2) for later. Although that is the best way to solve it, it sounds like quite an involved check that not _everyone_ is able to pick up.

rickie:
Added a commit with the XXX. 

I'd like to propose the following. Create a GH issue for (1) and keep (2) for later. Although that is the best way to solve it, it sounds like quite an involved check that not _everyone_ is able to pick up.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoMockClassReference.java
```diff
@@ -0,0 +1,90 @@
...
+ * A {@link BugChecker} that flags the use of {@link org.mockito.Mockito#mock(Class)} and {@link
+ * org.mockito.Mockito#spy(Class)} where instead the type to be mocked or spied can be derived from
+ * context.
+ */
+// XXX: This check currently does not flag method invocation arguments. When adding support for
+// this, consider that in some cases the type to be mocked or spied must be specified explicitly so
+// as to disambiguate between method overloads.
+// XXX: This check currently does not flag (implicit or explicit) lambda return expressions.
+// XXX: This check currently does not drop suppressions that become obsolete after the
+// suggested fix is applied, consider adding support for this.
```
Stephan202, line 41:
```suggestion
// suggested fix is applied; consider adding support for this.
```

Stephan202 (commented):
Having slept on it, I don't think (1) makes sense: an initial version of (2) could be one that only works for `mock()` calls.

rickie:
> Having slept on it, I don't think (1) makes sense: an initial version of (2) could be one that only works for mock() calls.

Let's discuss offline what the best approach would be. For now I'd propose to merge once 💚 ! 🚀

Badbond:
Thanks for the tweaks @Stephan202! PR LGTM for merge. Very nice that you flag additional cases of multi-arg mocking where we can still omit the class as well as the usage within lambda expressions. Amazing polish of tests, comments and casting too :heart_eyes:. 

> Having slept on it, I don't think (1) makes sense: an initial version of (2) could be one that only works for `mock()` calls.

I would also think (2) is the best option to also gain the most value. Complexity indeed to-be-seen (likely as part of implementation), but it sounds to me that this can be a generic check flagging safe casts (to subtype/to same type/raw-to-raw) annotated with the `unchecked` suppression.

PR #455, opened 2023-01-09
Upgrade Error Prone 2.17.0 -> 2.18.0

Stephan202:
Suggested commit message:
```
Upgrade Error Prone 2.17.0 -> 2.18.0 (#455)

See:
- https://github.com/google/error-prone/releases/tag/v2.18.0
- https://github.com/google/error-prone/compare/v2.17.0...v2.18.0
- https://github.com/PicnicSupermarket/error-prone/compare/v2.17.0-picnic-1...v2.18.0-picnic-1
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,7 +149,7 @@
         <version.auto-value>1.10.1</version.auto-value>
         <version.error-prone>${version.error-prone-orig}</version.error-prone>
         <version.error-prone-fork>v${version.error-prone-orig}-picnic-1</version.error-prone-fork>
-        <version.error-prone-orig>2.17.0</version.error-prone-orig>
+        <version.error-prone-orig>2.18.0</version.error-prone-orig>
         <version.error-prone-slf4j>0.1.17</version.error-prone-slf4j>
         <version.guava-beta-checker>1.0</version.guava-beta-checker>
         <version.jdk>11</version.jdk>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -1583,6 +1587,9 @@
                                     -Xep:StaticOrDefaultInterfaceMethod:OFF
                                     <!-- We generally discourage `var` use. -->
                                     -Xep:Varifier:OFF
+                                    <!-- Yoda conditions are not always more
+                                    readable than the alternative. -->
+                                    -Xep:YodaCondition:OFF
```
Stephan202, line 1592: To be sure: we can discuss enabling this, of course. Maybe I should just get used to it. Even so: the resultant changes also require careful `null` analysis.

rickie: I'm on the fence about this. Enabling this would result in better/easier readable code IMO. What concerns do you have with enabling this 👀? (Other than getting used to it 😉)

In any case, it would make sense to enable this in a separate PR (especially if we want to make the next internal PSM release).

Stephan202: > What concerns do you have with enabling this :eyes: (Other than getting used to it :wink: )

Well the question is whether it's just "getting used to", or that in some cases readability actually suffers. Separately from that it might force us to introduce additional explicit `null` checks, which would be plain awkward.

rickie: We could open a PR where we enable the check and go over the results and make a decision for EPS there? It is hard to reason about it now if we don't know how "bad" it'll be.

Stephan202: Draft PRs are free ;)

That said, I'd also like to see the arguments brought up during the internal discussion on this topic. :student:

rickie (approved):

chamil-prabodha (approved):

PR #465, opened 2023-01-16
Extend `MonoIdentity` Refaster rule

werli:
After applying EPS on a repository, I've noticed essentially the following interesting pattern:
```java
Mono.<Void>empty().flux().next()
```
- `Mono#flux()` converts the `Mono` to a `Flux` as-is.
- `Flux#next()` however simply takes the first element in the `Flux` or returns empty without any side-effects (contrary to `Flux#single()`).

So, this is practically a no-op.

### Suggested commit message
```
Extend `MonoIdentity` Refaster rule (#465)

By flagging expressions of the form `mono.flux().next()`.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -377,6 +377,11 @@ final class ReactorRules {
       return mono.then();
     }
 
+    @BeforeTemplate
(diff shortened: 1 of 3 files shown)


Stephan202 (approved):
Makes sense! Extended the suggested commit message.

rickie (approved):

rickie:
Nice @werli :rocket:!

werli:
@Stephan202 & @rickie _insane_ review speed, thanks!

PR #470, opened 2023-01-19
Extend `MonoIdentity` Refaster rule

werli:
Sorry for two bite size PRs, but I've just now noticed one more case which wasn't covered in #465.

`Mono#singleOrEmpty()` _has_ side-effects, but only if there is more than one element which by definition cannot occur after a `Mono#flux()` call.

I've checked again to see whether more expressions would qualify, however `Mono#single()` and `Mono#last()` don't qualify since they have side-effects when no element can be found.

### Suggested commit message
```
Extend `MonoIdentity` Refaster rule (#470)

By flagging expressions of the form `mono.flux().singleOrEmpty()`.
```


rickie (approved):
Looks nice! Small PRs are very welcome 😉. Easy to review as well ;).

Lekker bezig man!

Suggested commit message LGTM.

Stephan202 (approved):
Beautiful :heart:

PR #472, opened 2023-01-20
Introduce `FluxImplicitBlock` check

### LINKED ISSUE #468: Disallow implicit blocking operators of Project Reactor's Flux
### Problem

Excessive blocking is a known pitfall in reactive programming. Some of Reactor's blocking operators are quite explicit about their behavior, e.g. `Mono#block` and `Flux#blockFirst`, leaving no doubt that the operation is in fact blocking.

This does not apply to two of the convenience methods exposed by Reactor's `Flux`, [Flux#toStream](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#toStream--) and [Flux#toIterable](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#toIterable--). These operators are documented to block, but this is not apparent from the method signature. It is easy to miss that the operation is blocking despite any potential lazy consumption of the resulting `Stream` or `Iterable`.

- [ ] Improve performance.

We should rewrite the problematic operators in terms of explicitly blocking operators, highlighting actual behavior.

Focusing first on the stream, I would like to rewrite the following code:
```java
flux.toStream()
```

to:
```java
flux.collect(toList()).block().stream()
```

To be determined what a _general_ rewrite for `Flux#toIterable` would look like, or if this is actually desirable.

### Considerations

- The methods in question have overloads detailing their usage. We could consider allowing those, assuming that those operators are less likely to be used accidentally. Compare with the existing [FluxFlatMapUsage](https://error-prone.picnic.tech/bugpatterns/FluxFlatMapUsage/) rule.
- The problematic methods being convenience methods, the proposed solution is more verbose than the original.

### Participation

- [x] I am willing to submit a pull request to implement this improvement.

Willing, sure! Less sure on able. Would be happy to see anyone else submit a pull request to this effect. :D

benhalasi:
fixes #468

Flags implicitly blocking operators on Reactor's `Flux` as fragile code.

#### [Flux#toStream](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#toStream--)

Suggested alternatives for `flux.toStream()`:
1. SuppressWarnings
2. if guava is available, replace with: `flux.collect(toImmutableList()).block().stream();`
3. replace with: `flux.collect(toUnmodifiableList()).block().stream();`

#### [Flux#toIterable](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#toIterable--)
Suggested alternatives for `flux.toIterable()`:
1. SuppressWarnings
2. if guava is available, replace with: `flux.collect(toImmutableList()).block();`
3. replace with: `flux.collect(toUnmodifiableList()).block();`


Stephan202:
Thanks for filing this PR @benhalasi! The linked issue does suggest rewriting these methods, but also references [FluxFlatMapUsage](https://error-prone.picnic.tech/bugpatterns/FluxFlatMapUsage). I wonder whether we should go that same route here, and introduce a `BugChecker` rather than Refaster rules. My thinking is that, while the current approach is pragmatic and does cause `toStream()` and `toIterable()` to be flagged, the alternatives are considerably less efficient, and may even cause an `OutOfMemoryError` in case of very large or unbounded reactive streams.

The `BugChecker` could suggest several fixes, accompanied with a warning an `OutOfMemoryError` warning:
1. Adding `@SuppressWarnings` (like in [`IdentityConversion`](https://github.com/PicnicSupermarket/error-prone-support/blob/499f922328b6dfa69726aafc8be44b3fffd49e0b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IdentityConversion.java#L116)).
2. Adding `.collect(toImmutableList())`, [if Guava is available](https://github.com/PicnicSupermarket/error-prone-support/blob/499f922328b6dfa69726aafc8be44b3fffd49e0b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java#L26-L31).
3. Adding `.collect(toList())`.

(Given our coding style, Picnic users would likely select option (1) or (2). If I'd find this in code I was personally responsible for, I'd likely add `@SuppressWarnings` and file a cleanup ticket :thinking:.)

CC @Ernir for input as the original author of the issue.

benhalasi:
I see how I didn't get it right for the first time, will try to work it out and come back with something until Monday.

rickie:
No problem @benhalasi, it appears to be a bit more challenging than anticipated. In retrospect we maybe should've worked out this issue a bit further beforehand. 

If you have any questions, let us know 😄. We're happy to help.

Stephan202:
> I see how I didn't get it right for the first time, will try to work it out and come back with something until Monday.

No rush! :muscle:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,129 @@
...
+    link = BUG_PATTERNS_BASE_URL + "ImplicitBlockingFluxOperation",
+    linkType = CUSTOM,
+    severity = ERROR,
+    tags = LIKELY_ERROR)
+public class ImplicitBlockingFluxOperation extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Supplier<Type> FLUX =
+      Suppliers.typeFromString("reactor.core.publisher.Flux");
+  public static final MethodMatchers.MethodClassMatcher FLUX_METHOD =
```
rickie, line 48: IIUC this can be replaced with `Matcher<ExpressionTree>`  🙂.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperationTest.java
```diff
@@ -0,0 +1,141 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import com.google.errorprone.BugCheckerRefactoringTestHelper;
+import com.google.errorprone.BugCheckerRefactoringTestHelper.FixChoosers;
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+final class ImplicitBlockingFluxOperationTest {
+
```
rickie, line 9:
```suggestion
```
Checkstyle will complain about this :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
rickie: This probably doesn't need to be public, making it private will also make the error go away.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperationTest.java
```diff
@@ -0,0 +1,141 @@
...
+            "  void m() {",
+            "    flux().collect(toUnmodifiableList()).block();",
+            "    flux().collect(toUnmodifiableList()).block().stream();",
+            "  }",
+            "",
+            "  Flux<Integer> flux() {",
+            "    return Flux.just(1, 2, 3);",
+            "  }",
+            "}")
+        .doTest(BugCheckerRefactoringTestHelper.TestMode.TEXT_MATCH);
```
rickie, line 139:
```suggestion
        .doTest(TestMode.TEXT_MATCH);
```
This is a convention we have.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * A {@link BugChecker} that flags usages of such Flux methods that are documented to block, but
+ * that behaviour is not apparent from their signature.
+ *
+ * <p>The alternatives suggested are explicitly blocking operators, highlighting actual behavior.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary =
```
rickie, line 35: This explanation should go in the `explanation` field probably. A summary should be a short summary in (preferably and where possible) in on sentence.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,129 @@
...
+            + "please make sure that they are used with this in mind.",
+    link = BUG_PATTERNS_BASE_URL + "ImplicitBlockingFluxOperation",
+    linkType = CUSTOM,
+    severity = ERROR,
+    tags = LIKELY_ERROR)
+public class ImplicitBlockingFluxOperation extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Supplier<Type> FLUX =
+      Suppliers.typeFromString("reactor.core.publisher.Flux");
```
rickie, line 47: We can inline the string used in `FLUX` in the `FLUX_METHOD`.

rickie (changes requested):
Hey @benhalasi thanks for changing the implementation to this check! Will do a more thorough review probably tomorrow. Already left some comments to explain why the CI fails currently 😄.

benhalasi:
Thank you for the fast feedback.

---

Also here's the comment I forgot to attach:

I tried to stick to the coding style I saw in other files as much as possible, but I'm sure there's a lot of room to improve. I'm the least confident in the comments/javadocs and the tests. 

I'm looking forward to any suggestion or change request, and also to answers to some questions:

#### Code duplication between checkers

I copied matchers from e.g. `FluxFlatMapUsage`, is there a pattern to have these potentially shared matchers collected or the checkers considered to be fully independent from each-other and we don't care about these kinds of duplication?  

#### How to replace a method call with a method call chain.

In the `getCollectAndBlockFix` method I want to replace the matched method call with something like `flux.collect(..).block()...`.

For this, originally, I wanted to use `SuggestedFix.replace()`. This lost the flux symbol that we had the method calls on.

Then I tried to use one of the `SuggestedFixes.renameMethod*()` methods, but as I saw they expect a method name instead of a method call chain, they put a `()` after the replacement - it would be possible to use these methods, but it seems hacky.

As I saw the best solution is to have a 'replacer' that is similar to the existing rename methods, but instead of a method name it expects (potentially chained) method calls.

I didn't find such method and I wasn't sure enough about this to implement it without any feedback, so I went with a temporary - hacky - solution where I use the normal replace and prefix it with the part we would have lost.

benhalasi:
I committed the changes @rickie suggested.

I see some checks are still failing, I'm not able to reproduce them locally nor I want to spam further this PR so I leave it here until further feedback. The gust of the PR is already done I think.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,130 @@
...
+ * <p>The alternatives suggested are explicitly blocking operators, highlighting actual behavior,
+ * not adequate for automatic replacements, they need manual review.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    explanation =
+        "`Flux#toStream` and `Flux#toIterable` are documented to block, "
+            + "but this is not apparent from the method signature; "
+            + "please make sure that they are used with this in mind.",
+    summary = "Accidental blocking of `Flux` with convenience method.",
```
rickie:
```suggestion
    summary = "Accidental blocking of `Flux` with convenience method",
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,130 @@
...
+    link = BUG_PATTERNS_BASE_URL + "ImplicitBlockingFluxOperation",
+    linkType = CUSTOM,
+    severity = ERROR,
```
rickie: It is not per se an error, I think it's more of a suggestion :).

rickie: Same goes for `LIKELY_ERROR`, it is a preference of style. It is a good improvement to perform this rewrite but the original wasn't really "wrong".

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,130 @@
...
+    tags = LIKELY_ERROR)
+public class ImplicitBlockingFluxOperation extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<ExpressionTree> FLUX_TO_ITERABLE =
```
rickie: Will push a change for this one. Now the check I have at the bottom of the class is not so nice, will come back to that. Maybe indeed we need this setup but utilizing the `#namedAnyOf` makes a bit more sense to me. Maybe there is a easier way of checking whether we are making a `SuggestedFix` for the `toIterable` variant or not.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,130 @@
...
+    description.addFix(getSuppressWarningsFix(state));
+    if (ThirdPartyLibrary.GUAVA.isIntroductionAllowed(state)) {
+      description.addFix(getGuavaFix(tree, state));
+    }
+    description.addFix(getUnmodifiableListFix(tree, state));
+
+    return description.build();
+  }
+
+  private static SuggestedFix getSuppressWarningsFix(VisitorState state) {
```
rickie: I think we can inline this one.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,130 @@
...
+    SuggestedFix.Builder unmodifiableListFix = SuggestedFix.builder();
+    String toUnmodifiableList =
+        SuggestedFixes.qualifyStaticImport(
+            "java.util.stream.Collectors.toUnmodifiableList", unmodifiableListFix, state);
+    return getCollectAndBlockFix(tree, state, unmodifiableListFix, toUnmodifiableList + "()")
+        .build();
+  }
+
+  /**
+   * merges `flux.collect(...).block()...` fix into given fix with specified collector and postfix
```
rickie:
```suggestion
   * Merges `flux.collect(...).block()...` fix into given fix with specified collector and postfix
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,130 @@
...
+   * to match the original expression tree.
+   *
+   * @param collector expression
```
rickie: Note to self: this Javadoc is not really corresponding with the actual parameters 😉. Also, we add a `.` after every sentence.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,130 @@
...
+    // XXX: replace DIY string replace fix with something more resilient
+    String flux =
+        state.getSourceForNode(tree).substring(0, state.getSourceForNode(tree).indexOf("."));
+    String replacement = String.format("%s.collect(%s).block()%s", flux, collector, postfix);
+    fix.merge(SuggestedFix.replace(tree, replacement));
+    return fix;
+  }
+
+  /**
+   * finds the extension of `Flux.collect(...).block()` expression to match the original expression
```
rickie:
```suggestion
   * Finds the extension of `Flux.collect(...).block()` expression to match the original expression
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperationTest.java
rickie: Pushed this change :).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperationTest.java
```diff
@@ -0,0 +1,140 @@
...
+  @Test
+  void identification() {
+    CompilationTestHelper.newInstance(ImplicitBlockingFluxOperation.class, getClass())
+        .addSourceLines(
+            "A.java",
+            "import reactor.core.publisher.Flux;",
+            "",
+            "class A {",
+            "  void m() {",
+            "    // BUG: Diagnostic contains:",
```
rickie, line 19: We can simplify this testing setup like this:
```suggestion
            "    // BUG: Diagnostic contains:",
              "class A {",
            "  void m() {",
            "    Flux.just(1).toIterable();",
            "    Flux.just(2).toStream();",
            "  }")
```

Additionally it would be nice to add some other test cases where we use `toIterable` and `toStream` on other objects then `Flux` to show it doesn't flag those.

rickie (changes requested):
> I copied matchers from e.g. FluxFlatMapUsage, is there a pattern to have these potentially shared matchers collected or the checkers considered to be fully independent from each-other and we don't care about these kinds of duplication?

Good question. If some `Matchers` are used multiple times, we can consider adding them to `MoreMatchers.java`. Maybe we should have a Reactor specific matchers file in `error-prone-contrib`s `tech.picnic.errorprone.bugpatterns.util` package 🤔. 

All to say, yes we care about this kind of duplication and we want to avoid it 😄. 

> How to replace a method call with a method call chain.
In the getCollectAndBlockFix method I want to replace the matched method call with something like flux.collect(..).block()....

I think the `CollectorMutability` check has some nice similarities that might be interesting. Based on that check I think you are on the the right path, because [here](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CollectorMutability.java#L91) you can see an example that does the same: creating a string and replacing that completely. 

Didn't dive _too_ deep but it seems like the approach is right.

Oh and btw, the Pitest plugin is complaining because a test is missing. We need almost the same test as the `identificationWithoutGuavaOnClasspath` that is in `CollectorMutabilityTest.java`.  There needs to be a test that verifies that it _doesn't_ perform the Guava fix if the classpath is empty, that is the reason there is a `withClasspath()` method in the test I mention here. If you don't get around to adding that test, I'll do it when I do a proper review :).

Ahh there is now a failing test because of this line:
```
String flux = source.substring(0, source.indexOf("."));
```
I _think_ this is because I changed `flux().` with `Flux.just(1)` in the tests and is currently a limitation of the implementation. This is indeed where you asked specifically for help. I think there might be even a better option that we can use. See an example [here](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StringCaseLocaleUsage.java#L70-L75) in the `StringCaseLocaleUsage` check. It is also possible to specify specific positions where the fix should be applied. It might be smart to go this route. 

A lot of things to consider. Feel free to play with the options 😄. Thanks for the good questions and I hope my answers make some things clear. I'll soon dive into this again myself. However, for today I need to focus on some other stuff. If you have questions, let us know! 

I added a commit and applied some of the suggestions :).

benhalasi:
The answers and suggestions were very helpful, `ASTHelpers` and `ErrorProneTokens` are awesome.

I did a fairly large rewrite and I got some additional questions too :D, mostly about assumptions I made, whether they are valid. I marked these with `XXX` comments too.

#### `replaceMethodInvocationWithCollect` method

Main concerns:
- assumes that the replacement expression will evaluate to a `Mono<Collection<?>>` - which is valid for the current cases, but could be misleading later, when extending this rule.
- would return `emptyFix` if we a match an expression that evaluates to a type we don't handle.
  - if this is the way to go, how to test this behaviour?

#### `replaceMethodInvocation` method

It replaces `method(args)` in a matched `Clazz.optionally().chained().method(args)` expression.

First of all I'm not even sure if this is the correct name for it or not.
- in `CollectorMutability` there's method named the same that would just replace the whole match if I'm not mistaken.
- the `SuggestedFix.replace(tree.getMethodSelect(), replacement)` would replace `Clazz.optionally().chained().method` without the brackets or args.

I'm guessing the reason for this is that someone could just replace the method and then handle its args separately?

Has some assumptions that I'm not sure if it can have. If not, whether it should throw an error / warning or maybe should just return an `Optional`. Currently it will fail when:
- `ErrorProneTokens.getTokens(...)` returns empty list
- `ASTHelpers.getStartPosition(tree)` returns `Position.NOPOS`, which should only happen if `tree` is `null` (I think)
- token list doesn't contain the matched method.
- token list contains elements after the matched method invocation and its parenthesizes.

#### other 'helper' methods

I would guess the functionality of `isTokenTheInvokedMethod` and `isClassValidSubstituteFor` methods are already implemented and I just didn't find them, so these are naive implementations, may need some extra work if we go with them.

#### tests

I added test for identification without Guava on the classpath.

> Additionally it would be nice to add some other test cases where we use toIterable and toStream on other objects then Flux to show it doesn't flag those.

Didn't find any other class with `toStream` or `toIterable` methods, should I just create an empty class for this, if so, where should I implement it?

#### move flux matchers to `MoreMatchers`

I didn't work on this yet, it seems like a separate issue from this one. Am I wrong and should include it here?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,192 @@
...
+        .replace(methodInvocationStartPosition, methodInvocationEndPosition, replacement);
+  }
+
+  // XXX: Replace with prewritten solution (?) or rewrite it in a more resilient way.
+  private static boolean isTokenTheInvokedMethod(MethodInvocationTree tree, ErrorProneToken token) {
+    return token.hasName()
+        && Objects.equals(token.name().toString(), getSymbol(tree).getQualifiedName().toString());
+  }
+
+  // XXX: Replace with prewritten solution. (?)
```
rickie: So I addressed this XXX together with the next one. Indeed there is a solution provided out-of-the-box that we can use instead of these two methods. Error Prone doesn't have that good of a documentation that easily helps us identify such improvements. We can rely on the `Types#isSubType`  here, which allows us to do pretty cool stuff. 

Will push a suggestion please take a look at it 😄. Let me know if you have any questions. I combined it with a usage of our own beautiful mini "DSL", see `MoreTypes.java`. That allows us to greatly simplify the logic used here.

This might not be easy to get at first, so if you have any questions about it, let us know :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,192 @@
...
+  // XXX: Assumes that the specified tree is valid, has starting position and contains the matched
+  // method invocation.
+  // XXX: Assumes that the specified tree's end is the matched method invocation's end.
+  private static SuggestedFix.Builder replaceMethodInvocation(
+      MethodInvocationTree tree, String replacement, VisitorState state) {
+    ImmutableList<ErrorProneToken> tokens =
+        ErrorProneTokens.getTokens(SourceCode.treeToString(tree, state), state.context);
+
+    int treeStartPosition = getStartPosition(tree);
+    int methodInvocationStartPosition =
```
rickie, line 121: Nice improvements done here! 

I _think_ there might be a solution where we don't need to use the `ErrorProneToken`s. Usually we only go that route as "last resort" or if comments are involved. Now that I think of that, that _could_ also be the case here. Maybe we should add test cases for that. See the `StringCaseLocaleUsage` test cases for some examples. However, I'm not sure yet if that is the correct way to go about this either. I need to take some more time to dive into this before being able to give a good advice. For now I need to focus on some other things so will stop here :).

rickie (changes requested):
Thanks for applying the feedback @benhalasi . I see some really nice progress and a nice thought process :rocket:! 

I did a review but didn't yet get to the end. We are on the right path here! 

I addressed most of your questions in my comments. However as my review didn't get till the end I wasn't able to apply to all questions yet.

Added a few commits 😄.

---


Let me answer most of your other questions:

> assumes that the replacement expression will evaluate to a Mono<Collection<?>> - which is valid for the current cases, but could be misleading later, when extending this rule.

Good consideration; however, I think this is fine. In the case of `toStream()` the resulting type will be exactly the same right, as we end with `.stream()`. 

> if this is the way to go, how to test this behaviour?

No we generally use `Optional`s for this _if_ this is really necessary. However, in this exact case it probably wasn't necessary because we use `Matcher`s to match quite specific expressions.

> It replaces `method(args)` in a matched `Clazz.optionally().chained().method(args)` expression.

Yes this sounds good, it should only change the part you mention. I didn't thoroughly check this part of the logic yet so exact review TBD 😉. 

> `ASTHelpers.getStartPosition(tree)` returns `Position.NOPOS`

We should probably handle that case 😄. 

> I would guess the functionality of isTokenTheInvokedMethod and isClassValidSubstituteFor methods are already implemented and I just didn't find them, so these are naive implementations, may need some extra work if we go with them.

You are correct, worked on this and will push a suggestion!

> Didn't find any other class with toStream or toIterable methods, should I just create an empty class for this, if so, where should I implement it?

Ah indeed, it was harder than expected. At the very least we should indeed have a simple class with a method `.toStream()` and .toIterable()` to show it does nothing with those methods.

> I didn't work on this yet, it seems like a separate issue from this one. Am I wrong and should include it here?

Hmm yeah that's true. If we move it we would also need to add tests to the respective class. We can defer it.

benhalasi:
Thanks for the reviews, both the code and my understanding of it are a lot clearer now :smile:.

I made commits for some issues that have not been addressed yet (I think):

#### assumptions about given tree

I resolved `replaceMethodInvocation` method's assumptions by returning `Optional.empty()` when it encounters such a problem.

Following this, in `replaceMethodInvocationWithCollect` my usage of `Optional.map()` is questionable but it seemed better than an *if then early return*.

I took a look at `StringCaseLocaleUsage` to have an idea on how to test this, but this behavior is not tested there either.

#### test for (not) matching `toIterable` and `toStream` on other objects

I implemented `NotFlux` class for this, also made 'tests' for it. I don't have too much experience on writing tests so I might be absolutely wrong here, I tried anyways.

---

I yet to have a deep dive in some of the classes you referred and to fix the testing around guava.

benhalasi:
I made `identificationWithoutGuavaOnClasspath` test fail when guava gets introduced despite expected behavior. - I'm guessing this is not the best solution, it seemed to easy and a bit hacky, so I'll try to find alternatives based on utility classes.


The only example I found for using `ASTHelpers#getStartPosition` and testing if it returns `POSITION.NOPOS` is in `StringCaseLocaleUsage`. This case is not tested there either. My question is basically how to deal with this - I'd had things in mind, but I assume not testing this for `StringCaseLocaleUsage` is not an accident.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperationTest.java
```diff
@@ -0,0 +1,147 @@
...
+            "    Flux.just(1).toIterable();",
+            "    // BUG: Diagnostic contains:",
+            "    Flux.just(2).toStream();",
+            "    // BUG: Diagnostic contains:",
+            "    long count = Flux.just(3).toStream().count();",
+            "",
+            "    Flux.just(3).toStream(16);",
+            "    new NotFlux().toIterable();",
```
rickie, line 34: You put quite some effort in this solution, nice! Luckily we can make it a bit easier for ourselves and insert the class as subclass. Something else we could do is add another class on the same instance of the `CompilationTestHelper`. For an example see the `JUnitMethodDeclarationTest#identification` where we call `addSourceLines` twice, once for a `A.java` and once with `B.java`. 

I pushed a suggestion based on your changes 🚀 !

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxOperation.java
```diff
@@ -0,0 +1,140 @@
...
+  private static Optional<SuggestedFix> trySuggestFix(
+      String fullyQualifiedMethodInvocation, MethodInvocationTree tree, VisitorState state) {
+    SuggestedFix.Builder fix = SuggestedFix.builder();
+    String replacement =
+        SuggestedFixes.qualifyStaticImport(fullyQualifiedMethodInvocation, fix, state);
+
+    return replaceMethodInvocationWithCollect(tree, replacement + "()", fix.build(), state);
+  }
+
+  // XXX: Assumes that the generated `collect(...)` expression will evaluate to
```
rickie, line 92: This does make sense because we convert the method. The resulting type of the complete expression will not change, so the behavior is "preserved". So will drop the `XXX` 😄.

rickie (commented):
Wow, good job on the `identificationWithoutGuavaOnClasspath` test 🚀! I really like the idea. It is quite a specific case to test, so I actually would say this makes a lot of sense. I didn't think of that solution 😄. 

> The only example I found for using `ASTHelpers#getStartPosition` and testing if it returns `POSITION.NOPOS` is in `StringCaseLocaleUsage`. This case is not tested there either. My question is basically how to deal with this - I'd had things in mind, but I assume not testing this for `StringCaseLocaleUsage` is not an accident.

Ahh maybe I should've been clearer, I indeed meant that the way the `getStartPosition` is used there is interesting for this case. Indeed there is not really a test for that as we would need a way to have "fake code" or "code without an actual position". That is too much of a hassle to test, so we can leave it as is. So your assumption is correct 😉.

I added a commit with some of the changes I mentioned.

Again, thanks for taking this effort, it's appreciated. Pushing my changes for now but will circle back to this later today probably :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
```diff
@@ -0,0 +1,132 @@
...
+            : ".block()";
+    return replaceMethodInvocation(tree, collectMethodInvocation, state)
+        .map(fix -> fix.merge(additionalFix))
+        .map(fix -> fix.postfixWith(tree, postFix))
+        .map(SuggestedFix.Builder::build);
+  }
+
+  private static Optional<SuggestedFix.Builder> replaceMethodInvocation(
+      MethodInvocationTree tree, String replacement, VisitorState state) {
+    ImmutableList<ErrorProneToken> tokens =
```
rickie, line 108: I found a way to do this without the `ErrorProneToken`! I'll push the changes 😄.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
```diff
@@ -0,0 +1,132 @@
...
+import com.sun.tools.javac.util.Position;
+import java.util.Optional;
+import tech.picnic.errorprone.bugpatterns.util.SourceCode;
+import tech.picnic.errorprone.bugpatterns.util.ThirdPartyLibrary;
+
+/**
+ * A {@link BugChecker} that flags {@link reactor.core.publisher.Flux} operators that are implicitly
+ * blocking.
+ *
+ * <p>The alternatives suggested are explicitly blocking operators, highlighting actual behavior,
```
rickie, line 37: Compared the Javadoc with some other classes and I think we can omit the two paragraphs TBH :).

rickie (approved):
Now that I've made my way to the end of the class I saw that we can be a little stronger with our assumptions. So the changes I'll push seem a bit in conflict with what I explained before but let me try to explain it. I'm talking about the wrapping in `Optional`s here with e.g. for some of the return types. 

I compared the `Position.NOPOS` situation with some other examples and I think we can use a `checkState` here, because the end position should actually be there. Pitest will complain about this but we can leave that specific case.
So by adding a `checkState` can be certain that we have a Fix at that point, so we don't need to wrap everything in an `Optional`. This allows us to further cleanup code 😄.

I'll add a commit with the proposed changes.

(For an example of the `checkState` usage, check the `RefasterRuleCollection#L310`.)

I'm happy with the state of the PR and will pass on the baton to @Stephan202 to review the PR 😄. Again, thanks for the effort @benhalasi ! 🚀  See you on Friday 😉.

Let me know what you think of the current state of the PR and the latest changes.

Suggested commit message:
```
Introduce `FluxImplicitblock` check (#472)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
```diff
@@ -0,0 +1,106 @@
...
+          trySuggestFix("com.google.common.collect.ImmutableList.toImmutableList", tree, state));
+    }
+    description.addFix(trySuggestFix("java.util.stream.Collectors.toList", tree, state));
+
+    return description.build();
+  }
+
+  private static SuggestedFix trySuggestFix(
+      String fullyQualifiedMethodInvocation, MethodInvocationTree tree, VisitorState state) {
+    SuggestedFix.Builder fix = SuggestedFix.builder();
```
rickie: We can improve the name to make clear this is actually only containing relevant (static) import information.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
```diff
@@ -0,0 +1,106 @@
...
+
+    return replaceMethodInvocationWithCollect(tree, replacement + "()", fix.build(), state);
+  }
+
+  private static SuggestedFix replaceMethodInvocationWithCollect(
+      MethodInvocationTree tree,
+      String collectArgument,
+      SuggestedFix additionalFix,
+      VisitorState state) {
+    String collectMethodInvocation = String.format("collect(%s)", collectArgument);
```
rickie: We could name this `collect`, but not sure about it. Let's leave it open for other reviewers :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
```diff
@@ -0,0 +1,106 @@
...
+  private static SuggestedFix trySuggestFix(
+      String fullyQualifiedMethodInvocation, MethodInvocationTree tree, VisitorState state) {
+    SuggestedFix.Builder fix = SuggestedFix.builder();
+    String replacement =
+        SuggestedFixes.qualifyStaticImport(fullyQualifiedMethodInvocation, fix, state);
+
+    return replaceMethodInvocationWithCollect(tree, replacement + "()", fix.build(), state);
+  }
+
+  private static SuggestedFix replaceMethodInvocationWithCollect(
```
rickie: Have a suggestion for this method name: `replaceWithExplicitBlock`. Will push 😄.

rickie (commented):
Couldn't help myself and push a few minor name suggestions :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
```diff
@@ -0,0 +1,106 @@
...
+        SuggestedFixes.qualifyStaticImport(fullyQualifiedMethodInvocation, importFix, state);
+
+    return replaceWithExplicitBlock(tree, replacement + "()", importFix.build(), state);
+  }
+
+  private static SuggestedFix replaceWithExplicitBlock(
+      MethodInvocationTree tree,
+      String collectArgument,
+      SuggestedFix importFix,
+      VisitorState state) {
```
Stephan202: We can inline this one; the code will be just as clear or clearer, and it'll allow for tiny (irrelevant) performance improvements.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
```diff
@@ -0,0 +1,106 @@
...
+      description.addFix(
+          trySuggestFix("com.google.common.collect.ImmutableList.toImmutableList", tree, state));
+    }
+    description.addFix(trySuggestFix("java.util.stream.Collectors.toList", tree, state));
+
+    return description.build();
+  }
+
+  private static SuggestedFix trySuggestFix(
+      String fullyQualifiedMethodInvocation, MethodInvocationTree tree, VisitorState state) {
```
Stephan202: It's not a method invocation by itself:
```suggestion
      String fullyQualifiedCollectorMethod, MethodInvocationTree tree, VisitorState state) {
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
```diff
@@ -0,0 +1,106 @@
...
+
+  private static SuggestedFix replaceWithExplicitBlock(
+      MethodInvocationTree tree,
+      String collectArgument,
+      SuggestedFix importFix,
+      VisitorState state) {
+    String collect = String.format("collect(%s)", collectArgument);
+    Types types = state.getTypes();
+    String explicitBlock =
+        types.isSubtype(ASTHelpers.getResultType(tree), types.erasure(STREAM.get(state)))
```
Stephan202:
```suggestion
        ASTHelpers.isSubtype(ASTHelpers.getResultType(tree), STREAM.get(state), state)
```
(Also performs type erasure.)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxTest.java
```diff
@@ -0,0 +1,159 @@
...
+            "class A {",
+            "  void m() {",
+            "    // BUG: Diagnostic contains:",
+            "    Flux.just(1).toIterable();",
+            "    // BUG: Diagnostic contains:",
+            "    Flux.just(2).toStream();",
+            "    // BUG: Diagnostic contains:",
+            "    long count = Flux.just(3).toStream().count();",
+            "",
+            "    Flux.just(3).toStream(16);",
```
Stephan202: Let's also list the other overloads we don't wish to match.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxTest.java
```diff
@@ -0,0 +1,159 @@
...
+            "",
+            "  public final class Foo<T> {",
+            "    public Iterable<T> toIterable() {",
+            "      return ImmutableList.of();",
+            "    }",
+            "",
+            "    public Stream<T> toStream() {",
+            "      return Stream.empty();",
+            "    }",
+            "  }",
```
Stephan202: For these test cases we can omit redundant modifiers :)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxTest.java
```diff
@@ -0,0 +1,159 @@
...
+            "}")
+        .doTest();
+  }
+
+  @Test
+  void identificationWithoutGuavaOnClasspath() {
+    CompilationTestHelper.newInstance(ImplicitBlockingFlux.class, getClass())
+        .withClasspath(CorePublisher.class, Flux.class, Publisher.class)
+        .expectErrorMessage("X", not(containsPattern("toImmutableList")))
```
Stephan202, line 60: Nice test! For contrast we should then verify that the first test _does_ contain this string.

Stephan202 (approved):
Rebased and added a commit. Very nice work! :tada: 

I have one question. W.r.t. the name, alternatives could be:
- `ImplicitlyBlockingFlux`
- `FluxImplicitBlock`

WDYT of either of these?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
```diff
@@ -0,0 +1,97 @@
...
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Type;
+import com.sun.tools.javac.util.Position;
+import java.util.stream.Stream;
+import tech.picnic.errorprone.bugpatterns.util.ThirdPartyLibrary;
+
+/**
+ * A {@link BugChecker} that flags {@link reactor.core.publisher.Flux} operators that are implicitly
```
Stephan202, line 31:
```suggestion
 * A {@link BugChecker} that flags use of {@link reactor.core.publisher.Flux} operators that are implicitly
```

Stephan202 (commented):
One more thing I just noticed (don't have the branch checked out atm; can commit later.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
Stephan202: Committed a more elaborate version of this.

Stephan202 (commented):
So that leaves just the naming question.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxTest.java
benhalasi: I noticed that in the identification test the `containsPattern("SuppressWarnings.*toImmutableList.*toList")` will break when the order of suggestions is changed.
If I understand correctly, according to error-prone the suggestions should be ordered in decreasing preference, but this behavior should be tested through `replacement[First, Second, Third, ...]SuggestedFix` methods.

If this seems to be important, an alternative could be: 
``` java
  and(
      containsPattern("SuppressWarnings"),
      containsPattern("toImmutableList"),
      containsPattern("toList")))
```

Stephan202: Fair point! Will you add a commit? :smile:

benhalasi:
regarding the name, I think `FluxImplicitBlock` has better IDE support when it's full name is unknown, it will be easier to find in search prompts and will be grouped together with other Flux related classes. If these are common use-cases I'd vote for this one.

In the other hand, `ImplicitlyBlockingFlux` might be more 'English' and would be nicer to encounter in error-prone result.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFluxTest.java
benhalasi: Yes, in a minute

Stephan202:
Indeed, both are relevant considerations. Looking at the existing checks, I do see that we often choose the `FluxImplicitBlock` style, so I'm tending to that option. (Though I do also see a few rules that should then be renamed :eyes:.)

Let's see what @rickie thinks :smile:

Stephan202:
Changes LGTM :+1:

rickie:
> I have one question. W.r.t. the name, alternatives could be:
> 
> * `ImplicitlyBlockingFlux`
> * `FluxImplicitBlock`
> 
> WDYT of either of these?

I think it should be the `FluxImplicitBlock`. It's indeed more aligned with the other checks. When I went over this, I picked `ImplicitBlockingFlux` mostly because of `NonEmptyMono` check. That should probably be renamed to `MonoNonEmpty` then. However, that is unrelated to this PR. @Stephan202 we can go over the rest that might need a tweak this week.

W.r.t. the other changes I only did a brief look and saw some nice improvements. 

Will do another pass over the changes probably tomorrow 😄.

Stephan202:
Or `FluxImplicitBlocking`? No strong opinion. @rickie feel free to merge!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
```diff
@@ -0,0 +1,106 @@
...
+            : ".block()";
+    return replaceMethodInvocation(tree, collect, state)
+        .merge(importFix)
+        .postfixWith(tree, explicitBlock)
+        .build();
+  }
+
+  private static SuggestedFix.Builder replaceMethodInvocation(
+      MethodInvocationTree tree, String replacement, VisitorState state) {
+    String methodName = ASTHelpers.getSymbol(tree).getQualifiedName().toString();
```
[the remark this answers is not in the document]
rickie: So for `MethodSymbol`s we can have a rule that rewrites `getQualifiedName` to `getSimpleName` because there will be no dots, right?

rickie:
Improvements look really good! One open question for @Stephan202 about the Refaster rule. Will now rebase and merge once 🍏 !

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImplicitBlockingFlux.java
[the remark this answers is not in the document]
Stephan202: That's one option, yes, though note that this also applies to other `Symbol` subtypes. A more generic rule (hence "or perhaps a simple bug checker") would be to perform this rewrite for any strict subtype of `Symbol` other than `{Class,Delegated,Package}Symbol`.

But in either case the question is whether this is worth the effort right now. It can/should go on the list of "things pointed out during code review" that we discussed last week, though :smile:

PR #473, opened 2023-01-22
Document `MonoFlatMapToFlux` Refaster rule limitation

Stephan202:
Suggested commit message:
```
Document `MonoFlatMapToFlux` Refaster rule limitation
```

For those wishing to attach a debugger: the type against which Refaster matches is retrieved [here](https://github.com/google/error-prone/blob/66ec4fd3ee0a1f6ae97a4d5ad7570ea1c0741c5c/core/src/main/java/com/google/errorprone/refaster/Template.java#L168). I suppose that at the place where the binding is stored, we can inspect the lambda expression/method reference and perhaps construct a more accurate type, but that's a rabbit hole I'd rather not jump in right now.


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -474,6 +474,11 @@ Flux<S> after(
    * Flux}.
    */
   abstract static class MonoFlatMapToFlux<T, S> {
+    // XXX: I'd be more expressive if this `@Placeholder` were replaced with a `Function<? super T,
```
rickie, line 477:
```suggestion
    // XXX: It'd be more expressive if this `@Placeholder` were replaced with a `Function<? super T,
```

rickie: Maybe writing it out is also fine :).

rickie (approved):
That sounds like an interesting problem 😋. 

Added a commit.

Stephan202:
Change LGTM :)

PR #476, opened 2023-01-25
Introduce BugPattern for removing duplicate `Mockito.verifyNoInteractions()` calls

### LINKED ISSUE #475: Introduce BugPattern for removing duplicate `Mockito.verifyNoInteractions()` calls
### Problem

When verifying that mocks didn't have interactions in tests sometimes many calls to `Mockito.verifyNoInteractions()` are used
`verifyNoInteractions(mock1);`
`verifyNoInteractions(mock2);`
which unnecessarily pollutes the code, because this method actually accepts varargs.
This can be simplified to just:
`verifyNoInteractions(mock1, mock2);`

- [ ] Improve performance.

We should rewrite multiple calls to `verifyNoInteractions` to one call.

```java
verifyNoInteractions(mock1);
verifyNoInteractions(mock2);
...
verifyNoInteractions(mockN);
```

to:
```java
verifyNoInteractions(mock1, mock2, ..., mockN);
```

### Considerations

Tricky parts in my opinion are:
1. That there might be a variable number of mocks verified in this way.
2. That multiple individual calls might not be consequential.
3. That the scope we need to check this must be just inside one test method.

### Participation

- [x] I am willing to submit a pull request to implement this improvement.(I'm willing to _try_ to submit a pr, this will be my first encounter with error prone 🙂)

Kamil-Gabaydullin:
This pr adds a `BugPattern` to recognize multiple `verifyNoInteractions()` calls in test methods.

I'll leave comments throughout the pr to address questions and decisions.


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsage.java
```diff
@@ -0,0 +1,98 @@
...
+  private static final long serialVersionUID = 1L;
+
+  @Override
+  public Description matchMethod(MethodTree tree, VisitorState state) {
+    boolean isTestMethod = TEST_METHOD.matches(tree, state);
+    if (!isTestMethod) {
+      return Description.NO_MATCH;
+    }
+    var verifyNoInteractionsInvocations = getVerifyNoInteractionsInvocations(tree);
+    if (verifyNoInteractionsInvocations.size() < 2) {
```
Kamil-Gabaydullin, line 53: If there is only one call to `verifyNoInteractions()` we don't want to do anything too

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsage.java
```diff
@@ -0,0 +1,98 @@
...
+    SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
+    MethodInvocationTree lastInvocation =
+        verifyNoInteractionsInvocations.get(verifyNoInteractionsInvocations.size() - 1);
+    verifyNoInteractionsInvocations.forEach(
+        invocationTree -> {
+          if (!invocationTree.equals(lastInvocation)) {
+            fixBuilder.replace(
+                ASTHelpers.getStartPosition(invocationTree),
+                state.getEndPosition(invocationTree) + 1,
+                "");
```
Kamil-Gabaydullin, line 72: This replacement leaves an empty line behind, I couldn't figure out how to collapse lines 😞 or perhaps it should be taken care of by formatting and it's not a big deal?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsage.java
```diff
@@ -0,0 +1,98 @@
...
+          }
+        });
+    fixBuilder.replace(lastInvocation, "verifyNoInteractions(" + combinedArgument + ")");
```
Kamil-Gabaydullin, line 75: I've decided that adding it as a last line in a test might not be preferable and thus replacing the last invocation is better, since it preserved intended order and place of this verification.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsage.java
```diff
@@ -0,0 +1,98 @@
...
+  private static ImmutableList<MethodInvocationTree> getVerifyNoInteractionsInvocations(
+      MethodTree methodTree) {
+    ImmutableList.Builder<MethodInvocationTree> invocationTreeBuilder = ImmutableList.builder();
+
+    new TreeScanner<@Nullable Void, @Nullable Void>() {
+      @Override
+      public @Nullable Void visitMethodInvocation(
+          MethodInvocationTree node, @Nullable Void unused) {
+        Symbol symbol = ASTHelpers.getSymbol(node);
+        if (symbol.getSimpleName().toString().equals("verifyNoInteractions")) {
```
Kamil-Gabaydullin, line 89: Not sure if this is the right way to determine a call to `verifyNoInteractions`(my first approach was `node.getMethodSelect().toString()` but then I was quickly shut down by a compilation error `Tree#toString shouldn't be used` 😅)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsageTest.java
```diff
@@ -0,0 +1,193 @@
...
+            "import static org.mockito.Mockito.mock;",
+            "import static org.mockito.Mockito.verifyNoInteractions;",
+            "",
+            "import org.junit.jupiter.api.Test;",
+            "",
+            "class A {",
+            "  @Test",
+            "  void m() {",
+            "    Runnable runnable1 = mock(Runnable.class);",
+            "",
```
Kamil-Gabaydullin, line 181: To the point that there are empty lines left behind 😞

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsageTest.java
```diff
@@ -0,0 +1,193 @@
...
+            "  @Test",
+            "  // BUG: Diagnostic contains:",
+            "  void f() {",
+            "    Runnable runnable1 = mock(Runnable.class);",
+            "    verifyNoInteractions(runnable1);",
+            "    Runnable runnable2 = mock(Runnable.class);",
+            "    Runnable runnable3 = mock(Runnable.class);",
+            "    verifyNoInteractions(runnable2, runnable3);",
+            "    Runnable runnable4 = mock(Runnable.class);",
+            "    Runnable runnable5 = mock(Runnable.class);",
```
Kamil-Gabaydullin, line 62: Perhaps there is a less verbose way to test this 🤔

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsage.java
```diff
@@ -0,0 +1,98 @@
...
+          if (!invocationTree.equals(lastInvocation)) {
+            fixBuilder.replace(
+                ASTHelpers.getStartPosition(invocationTree),
+                state.getEndPosition(invocationTree) + 1,
+                "");
+          }
+        });
+    fixBuilder.replace(lastInvocation, "verifyNoInteractions(" + combinedArgument + ")");
+
+    return describeMatch(tree, fixBuilder.build());
```
Kamil-Gabaydullin, line 77: This implementation will not find `Mockito.verifyNoInteractions()` calls, only `verifyNoInteractions()` with a static import 😞

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsageTest.java
```diff
@@ -0,0 +1,193 @@
...
+            "A.java",
+            "import static org.mockito.Mockito.mock;",
+            "import static org.mockito.Mockito.verifyNoInteractions;",
+            "",
+            "import org.junit.jupiter.api.Test;",
+            "",
+            "class A {",
+            "  @Test",
+            "  void m() {",
+            "    Runnable runnable1 = mock(Runnable.class);",
```
rickie, line 83: Nice tests @Kamil-Gabaydullin ! 
 
You already added quite some cases which is awesome.
What could even further improve the tests is by having other test statements between all these mocks. 
In tests there are also other methods that are mocked and used, I'm curious how those cases would be handled in replacement tests.

So in other words, now it contains actual mock statements and the `verifyNoInteractions` invocations. It would be nice if we add some `when(...)` calls and a random `assertThat` call maybe. That would make it a little more realistic tests and tests also the non-happy path a bit more.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsage.java
```diff
@@ -0,0 +1,98 @@
...
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = SIMPLIFICATION)
+public final class MockitoVerifyNoInteractionsUsage extends BugChecker
+    implements MethodTreeMatcher {
+  private static final long serialVersionUID = 1L;
+
+  @Override
+  public Description matchMethod(MethodTree tree, VisitorState state) {
+    boolean isTestMethod = TEST_METHOD.matches(tree, state);
```
rickie, line 48: We can inline this :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsage.java
```diff
@@ -0,0 +1,98 @@
...
+    if (!isTestMethod) {
+      return Description.NO_MATCH;
+    }
+    var verifyNoInteractionsInvocations = getVerifyNoInteractionsInvocations(tree);
```
rickie, line 52: In this codebase we don't use `var` 😄.

rickie (changes requested):
Hey @Kamil-Gabaydullin , thanks for opening the PR! Looks cool 🚀 !

I left a few suggestions on the PR 😄. I have to do some expectation management. The are a lot of PRs in review so I'm not sure when one of us gets around to doing a in depth review...
We'll keep you up to date for sure.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsage.java
rickie: > but then I was quickly shut down by a compilation error `Tree#toString` shouldn't be used :sweat_smile:)

We have a nice method for that. One can use the `SourceCode#treeToString` for that 😄. However, in this case it might be more useful to create a `Matcher` that matches on this specific method invocations from `verifyNoInteractions` which are also of type `org.mockito.Mockito`. Now every method that is named `verifyNoInteractions` will be matched. We don't want that :). 

(We should probably also add a test for that case 😉.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsage.java
rickie: Ahh I addressed this in a comment above I think.

Kamil-Gabaydullin:
> * Surviving mutants in this change: 1
> * Killed mutants in this change: 19
> 
> class	surviving	killed
> 🧟`tech.picnic.errorprone.bugpatterns.MockitoVerifyNoInteractionsUsage$1`	1	2
> 🎉`tech.picnic.errorprone.bugpatterns.MockitoVerifyNoInteractionsUsage`	0	17
> Mutation testing report by [Pitest](https://pitest.org/). Review any surviving mutants by inspecting the line comments under [_Files changed_](476/files).

No idea if something can be done with this warning that `return super.visitMethodInvocation(node, unused);` can be replaced with `return null;` in `TreeScanner` and no tests would fail, I've replaced same calls with nulls in `LexicographicalAnnotationAttributeListing` `TreeScanner` implementation and no tests failed as well 😬

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MockitoVerifyNoInteractionsUsage.java
```diff
@@ -0,0 +1,111 @@
...
+    ImmutableList<MethodInvocationTree> verifyNoInteractionsInvocations =
+        getVerifyNoInteractionsInvocations(tree, state);
+    if (verifyNoInteractionsInvocations.size() < 2) {
+      return Description.NO_MATCH;
+    }
+    String combinedArgument =
+        verifyNoInteractionsInvocations.stream()
+            .map(MethodInvocationTree::getArguments)
+            .flatMap(List::stream)
+            .map(Object::toString)
```
Stephan202, line 68: :eyes: looks like we're converting a `Tree` to `String` here. For that we should use `SourceCode.treeToString`.

Stephan202 (commented):
Thanks for contributing to Error Prone Support, @Kamil-Gabaydullin!

I haven't deeply looked into the PR yet, but I'm seeing parallels with #443: in both cases we want to merge adjacent method calls. Other examples might include e.g. `addAll` builder methods. So I'm thinking that perhaps a more generic check is hiding in this code :eyes:

rickie:
Hi @Kamil-Gabaydullin, for a while there was no activity on this PR. Partly because @Stephan202 and I were swamped and reviewing PRs that contain a new BugPattern can take quite some time. Still, having to wait for a long time for reviews is not nice. This situation is far from ideal and we want to do better. Hence, we set up an (internal) review pool to have others help out with reviewing PRs. Concretely this means we want to continue where we left off and start reviewing this again.

Therefore, we'd like to discuss the state and direction of the PR. Stephan flagged in [this comment](https://github.com/PicnicSupermarket/error-prone-support/pull/476#pullrequestreview-1345973503) that the current state could lead to semantic changes and therefore we need to change some things. There are some things we need to consider to determine how we want to continue with this PR.

There are a few options:
1. Update the PR to only merge _adjacent_ calls of both `Mockito#{reset,verifyNoInteractions}`.
2. Make this PR more generic and merge any adjacent method invocations that can be merged. Another example would be the `addAll`, so it would focus on adjacent _statements_ that can be merged.
3. Generify even further and also allow for chained method calls to be merged. This means that the functionality of  #443 will also be part of this PR. As a result, we can solve the merging of two method invocations for once and for all :smile:.
 
So the question is, are you still up for applying changes to get it into a state where we can start reviewing again? If yes, which direction do you want to go in?

Kamil-Gabaydullin:
Yep, I've been thinking for a while about this one after Stephan's comments.
To me the best direction for this PR seems to be the third one you've described, since making it only about `Mockito` is clearly just small part of the problem when there is an apparent generalization that can be done.

For the implementation, I think we should run this `BugPattern` only for specific cases(i.e. explicitly specify `Mockito.verifyNoInteractions`, `Mockito.verifyNoMoreInteractions`, `ImmutableList.add`, `ImmutableList.addAll` etc.), so that we can ensure that there is a vararg alternative present and also type safety, and perhaps debugging is easier when the calls are chained for some - so we'd like to avoid squashing the calls to `vararg` method. What's the best approach of specifying these calls - I have no idea, looks like this `BugPattern` would be the first of a kind 😅 

This doesn't seem like a "good first issue" anymore though 😁 I'm okay to _try_ implementing this (again), but I have my doubts 😅

Kamil-Gabaydullin:
Sounds good, I'll have a look around and pick something when I have the time.
Should I close this PR (and the issue)? Or close this PR and make changes to the issue?

PR #491, opened 2023-02-05
Run `org.kordamp.maven:pomchecker-maven-plugin:1.10.0:check-maven-central`

Stephan202:
Suggested commit message:
```
Run `org.kordamp.maven:pomchecker-maven-plugin:1.10.0:check-maven-central` (#491)

This new Maven goal validates that the `pom.xml` files meet the criteria
for being deployed to Maven Central. 
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1375,6 +1375,23 @@
                     </configuration>
                 </plugin>
                 <plugin>
+                    <groupId>org.kordamp.maven</groupId>
+                    <artifactId>pomchecker-maven-plugin</artifactId>
+                    <version>1.7.0</version>
+                    <configuration>
+                        <release>false</release>
+                    </configuration>
+                    <executions>
+                        <execution>
+                            <id>check-maven-central-compliance</id>
+                            <goals>
+                                <goal>check-maven-central</goal>
+                            </goals>
+                            <phase>verify</phase>
+                        </execution>
+                    </executions>
+                </plugin>
+                <plugin>
                     <groupId>org.pitest</groupId>
                     <artifactId>pitest-maven</artifactId>
                     <version>1.15.3</version>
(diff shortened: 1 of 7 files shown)


Stephan202:
The plugin (unconditionally) emits warnings such as:
```
[INFO] --- pomchecker-maven-plugin:1.7.0:check-maven-central (check-maven-central-compliance) @ refaster-compiler ---
[WARN] <url> is not defined in POM. Will use computed value from parent: 
	https://error-prone.picnic.tech/refaster-compiler
```

This is fair. We should indeed either override/duplicate this property, and/or update the website to provide some documentation for each module. Suppressing the warning requires the former, but before I do so: @rickie any thoughts on using the same URL (i.e. the home page) everywhere, vs. creating sub-pages? Perhaps a page in which we describe the project structure could be helpful.

rickie (approved):
Cool stuff! Configuration looks good.

rickie:
I agree that having a page per module would be nice and should definitely be the end goal. 

For now we can duplicate the URL indeed and start working on an overview page after that?

chamil-prabodha (approved):
Awesome! 🚀

pom.xml
```diff
@@ -1280,6 +1280,24 @@
                     <artifactId>jacoco-maven-plugin</artifactId>
                     <version>0.8.8</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.kordamp.maven</groupId>
+                    <artifactId>pomchecker-maven-plugin</artifactId>
+                    <version>1.8.0-SNAPSHOT</version>
```
Stephan202, line 1286: I upgraded to this SNAPSHOT to test kordamp/pomchecker#9; works nicely! We can merge once 1.8.0 is released :)

Stephan202 (commented):
@rickie @chamil-prabodha the PR was simplified; PTAL. (The build fails because of the `<requirePluginVersions />` check; that's expected for now.)

Stephan202:
Rebased, added a commit and updated the suggested commit message. But the build now fails on:
```
The plugin org.kordamp.maven:pomchecker-maven-plugin:1.8.0 requires Maven version 3.9.0
```

So this PR is now blocked for the same reason that #434 is.

rickie:
Oke thanks for updating though, will add `pause renovate` for now 😄.

Stephan202:
By now version 1.9.0 is out, and it too requires Maven 3.9.0. So this upgrade is still blocked on #434.

Stephan202:
Bumped to version 1.10.0. This PR will be unblocked by #964.

PR #494, opened 2023-02-08
Introduce `BugPatternTestExtractor` with tests

rickie:
❗ This is draft PR on top of #428 ❗ 

_It looks like a huge PR but it consists mostly of tests and associated resources!_

There are some open points that we should discuss before we mark this as 100% ready for review:
- We should have a way to make our utils more widely available, instead of having them in `error-prone-contrib`. It would be nice to be able to use the `SourceCode` util in `documentation-support` such that we can de-duplicate code (as well as an `Matcher` that lives in our utils). Let's discuss how to approach this.
- The heuristics used for determining what a BugPattern test is are improved w.r.t. the initial implementation. However, we can likely discuss further improvements. Still, the idea of the (internal Jira) ticket for this was to get this working in a better way than we have currently on the `website` branch. If we want to further improve the heuristics, we can discuss that maybe in a follow up PR. I also added an XXX for this.
- The current extraction approach of the `BugPatternExtractor#extract` is not as optimal as it can be. However, again, the goal of this ticket is not to implement the most optimal solution _now_. I added some XXXs to indicate where we can further improve.

Let's discuss this and hopefully continue with this approach.


documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    ScanBugPatternTest scanBugPatternTest = new ScanBugPatternTest();
+    scanBugPatternTest.scan(tree, state);
+
+    String bugPatternName = className.substring(0, className.lastIndexOf("Test"));
+    return scanBugPatternTest.hasTestUsingClassInstance(bugPatternName);
+  }
+
+  private static final class ScanBugPatternTest extends TreeScanner<@Nullable Void, VisitorState> {
```
rickie, line 70: The name of this `Scanner` can probably be improved.... The same holds for the `encounteredClasses` and `hasTestUsingClassInstance`.... Idk why, but really struggled with these names 🤔. Definitely open to suggestions 😄.

documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
```diff
@@ -0,0 +1,180 @@
...
+    abstract String name();
+
+    abstract ImmutableList<String> identificationTests();
+
+    abstract ImmutableList<BugPatternReplacementTestDocumentation> replacementTests();
+  }
+
+  @AutoValue
+  abstract static class BugPatternReplacementTestDocumentation {
+    static BugPatternReplacementTestDocumentation create(String sourceLines, String outputLines) {
```
rickie, line 171: Couldn't find a way around this, the way we instantiate the `BugPatternTestDocumentation` earlier on wasn't possible with this class...

documentation-support/src/test/java/tech/picnic/errorprone/documentation/BugPatternTestExtractorTest.java
```diff
@@ -0,0 +1,465 @@
...
+  void minimalBugPatternTest(@TempDir Path outputDirectory) throws IOException {
+    JavacTaskCompilation.compile(
+        outputDirectory,
+        "IdentityConversionTest.java",
+        "import com.google.errorprone.BugCheckerRefactoringTestHelper;",
+        "import com.google.errorprone.bugpatterns.BugChecker;",
+        "import com.google.errorprone.CompilationTestHelper;",
+        "",
+        "final class IdentityConversionTest {",
+        "  private static class IdentityConversion extends BugChecker {}",
```
rickie, line 36: Came up with this way to make testing a little easier... It's hard to make sure that the `IdentityConversion` is available without having to do an import. I figured this was "closest" to our actual testing setup...

rickie (commented):
Some context.

Added a commit that further simplifies the tests and fixes some formatting. Our check doesn't like the extra level of meta-ness in the tests 😋.

rickie:
Applied the feedback @Stephan202 gave offline. It's ready for review 😄.

documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
```diff
@@ -0,0 +1,179 @@
...
+            BugPatternReplacementTestDocumentation.create(
+                getSourceLines(node), replacementOutputLines));
+      } else if (REPLACEMENT_OUTPUT.matches(node, state)) {
+        replacementOutputLines = getSourceLines(node);
+      }
+      return super.visitMethodInvocation(node, unused);
+    }
+
+    // XXX: Duplicate from `ErrorProneTestSourceFormat`, should we move this to `SourceCode` util?
+    private static String getSourceLines(MethodInvocationTree tree) {
```
rickie, line 142: > should we move this to `SourceCode` util? 

Didn't apply something like this yet, to not introduce a dependency on `error-prone-contrib`. I think we should have a more generic "utilities" module.

documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
```diff
@@ -0,0 +1,179 @@
...
+    private final List<String> encounteredClasses = new ArrayList<>();
+
+    boolean hasTestUsingClassInstance(String clazz) {
+      return encounteredClasses.contains(clazz);
+    }
+
+    @Override
+    public @Nullable Void visitMethodInvocation(MethodInvocationTree node, VisitorState state) {
+      if (BUG_PATTERN_TEST_METHOD.matches(node, state)) {
+        MemberSelectTree firstArgumentTree = (MemberSelectTree) node.getArguments().get(0);
```
Stephan202, line 93: While not likely in our code base, this could cause a `ClassCastException`. Can be fixed by augmenting the `Matcher`.

Stephan202 (commented):
Rebased and added a commit. This was a partial review; more at another moment. Nice feature!

Stephan202 (commented):
Rebased and added a commit. More later.

Stephan202 (commented):
Rebased and added a commit. Still need to review the new extractor and associated tests, but the framework code is now nicer, IMHO.

First stab at a suggested commit message:
```
Introduce `BugPatternTestExtractor` (#494)

This change extends the `documentation-support` module with an extractor of
`BugPattern` test source code snippets.

While there, simplify the `Extractor` interface, drop the `ExtractorType` enum 
and improve test coverage.
```

documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+
+  @Override
+  public String identifier() {
+    return "bugpattern-test";
+  }
+
+  // XXX: Improve support for correctly extracting multiple sources from a single
+  // `{BugCheckerRefactoring,Compilation}TestHelper` test.
+  @Override
+  public Optional<BugPatternTestDocumentation> tryExtract(ClassTree tree, VisitorState state) {
```
Stephan202, line 64: Before reviewing further, I'd like to propose a different extraction setup: rather than returning a single `BugPatternTestDocumentation`, we can return a list of test descriptions. This way we make no assumptions about class names etc. It may make processing the JSON a bit more involved, but likely not significantly so. Perhaps easier to discuss this topic offline.

rickie: The initial idea was to get these first set of improvements out while trying to minimize changes to the JSON output. After that when everything is nicely tested we can easily iterate over the setup and improve.  

Anyway; let's indeed discuss offline.

Badbond:
Rebased.

documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
Badbond: We still need to make assumptions with the classnames, but then in the Ruby code though. We need to match the bug pattern documentation with the test data, [which currently happens based on file names](https://github.com/PicnicSupermarket/error-prone-support/blob/66ad1190ab21a25390327ddfdeedc821f6c2ea5b/website/generate-docs.rb#L23-L24). It would be more useful if we could've already combined this at an extractor level, but that might be slightly trickier given needing to combine multiple classes over different packages.

documentation-support/src/test/resources/tech/picnic/errorprone/documentation/bugpattern-test-documentation-replacement-expect-unchanged.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "TestChecker",
+  "identificationTests": [],
+  "replacementTests": [
+    {
+      "inputLines": "class A {}\n",
+      "outputLines": ""
+    }
+  ]
```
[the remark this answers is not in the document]
rickie: My current thinking leans towards leaving the unchanged examples out.

rickie:
Rebased.

rickie:
So at least I have something _working_ right now. Changing the setup took me a bit too long. I think the breaks in between working on this and not taking the time to actually get to the bottom of this weren't helping. 

I'll push something for now, but did not look into optimizing the code now. If you have some time feel free to skim over it and give some pointers. I went for a solution where we are quite ready for future improvements (e.g. adding the `TestEntry` class as well). Didn't see the value of having only a string there if we wanted to have multiple classes and wanted to be clear about their paths / packages later on, LMKWYT 😄.  

Tomorrow we probably have a meeting and I'll ask you some things about it :).

We talked about the scanner, I wasn't sure whether we talked about not introducing another one or that we wanted to get rid of the _existing_ one as well... Sorry for that. Either way, this is not a final state but wanted to at least push this.

rickie:
Had a productive flight ✈️! Have a new proposal and I really like the structure of the JSON files now! Extending later will be easy 😄.

documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
```diff
@@ -0,0 +1,227 @@
...
+                if (JUNIT_TEST_METHOD.matches(m, state)) {
+                  scanner.scan(m, state);
+                }
+              }
+
+              return scanner.getTests().isEmpty()
+                  ? null
+                  : new AutoValue_BugPatternTestExtractor_TestCases(
+                      bugPatternName, scanner.getTests());
+            });
```
oxkitsune, line 68:
```suggestion
        .flatMap(
            bugPatternName -> {
              CollectBugPatternTests scanner = new CollectBugPatternTests();

              for (Tree m : tree.getMembers()) {
                if (JUNIT_TEST_METHOD.matches(m, state)) {
                  scanner.scan(m, state);
                }
              }
              
              return scanner.getTests().map(tests -> 
                new AutoValue_BugPatterTestExtractorTestCases(
                  bugPatternName, tests));
            });
```

oxkitsune: As mentioned offline, my suggestion sorta reduces readability! So I prefer the original I think :smile:

rickie:
Rebased :).

Stephan202 (commented):
We're getting close :). I rebased and added a commit. Didn't finish review of the tests yet; hence the new `XXX` comments.

documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
```diff
@@ -0,0 +1,230 @@
...
+                  List<TestEntry> entries = new ArrayList<>();
+                  if (isReplacementTest) {
+                    extractReplacementTestCases(node, entries, state);
+                  } else {
+                    extractIdentificationTestCases(node, entries, state);
+                  }
+
+                  collectedTestCases.add(
+                      new AutoValue_BugPatternTestExtractor_TestCase(
+                          classUnderTest, ImmutableList.copyOf(entries).reverse()));
```
Stephan202, line 109: We should validate here that `entries` is not empty.

Stephan202: Ah, I meant to start a new review. Have local changes for this already.

documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
```diff
@@ -0,0 +1,230 @@
...
+import java.util.Optional;
+import org.jspecify.annotations.Nullable;
+import tech.picnic.errorprone.documentation.BugPatternTestExtractor.TestCases;
+
+/**
+ * An {@link Extractor} that describes how to extract data from classes that test a {@code
+ * BugChecker}.
+ */
+// XXX: Consider whether to omit or handle differently identification tests without `// BUG:
+// Diagnostic (contains|matches)` markers.
```
Stephan202: I think this makes sense. Likewise for unchanged code. (Now that I say this, I vaguely recall that perhaps we discussed pushing this to the website generation script. But I already updated the code :grimacing:. If we'd like to keep this, then more elegant would be to introduce "this code is unchanged" output. That would require further work, though, as it'd require explaining things like [this](https://github.com/PicnicSupermarket/error-prone-support/blob/6d27e46baf64b51d428f75f67277cbff7e39964d/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CollectorMutabilityTest.java#L63C1-L78).)

Stephan202 (commented):
Rebased and added another commit with a bunch of changes. I think that the only point that remains from my side is to review the second half of the tests in `BugPatternTestExtractorTest`; likely those can be improved/condensed a bit.

Stephan202 (commented):
Added one more commit. Will later do a final pass an approve.

Stephan202 (approved):
Suggested commit message:
```
Introduce `BugPatternTestExtractor` for documentation generation (#494)

This new `Extractor` implementation collects identification and replacement 
source code from `BugChecker` unit tests.

While there:
- Refactor the existing `Extractor` setup such that instances are service 
  loaded and need to implement only a single method, thereby avoiding the need
  to align logic between multiple source code locations.
- Extend the validation performed by the `Compilation` test helper class.
- Extend the `ErrorProneTestHelperSourceFormat` check to support source code 
  passed to the `Compilation` test helper class.
```

Stephan202 (commented):
Rebased and added a commit; tnx @mohamedsamehsalah!

Before we merge this PR, ideally we prepare the `website` branch to be compatible with it. But besides that: LGTM!

Stephan202:
These changes trigger a warning that is new in JDK 21; will have a look.

Stephan202:
K, when I work around that warning the JDK build fails on [this exception](https://github.com/google/error-prone/blob/747227a8c8f6ee91c7a57859e4402e5aeca921c5/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L320C17-L320C41) with `java.lang.IllegalArgumentException: super()`. Turns that the symbol is `null` for that `MethodInvocationTree`. I don't see a fix yet.

Stephan202:
Pushed a commit to resolve [this warning](https://inside.java/2023/07/29/quality-heads-up/), but the `super()` symbol issue requires a deeper look; something for ~tomorrow.

Stephan202:
K, figured it out. Added one more commit with a fix and a comment explaining the fix. Also added some other compile parameters that may help with debuggability and/or reproducibility.

Stephan202:
Repeating a tweaked version of the [earlier](https://github.com/PicnicSupermarket/error-prone-support/pull/494#pullrequestreview-1554443574) suggested commit message, for easier copying later on:
```
Introduce `BugPatternTestExtractor` for documentation generation

This new `Extractor` implementation collects identification and
replacement source code from `BugChecker` unit tests.

While there:
- Refactor the existing `Extractor` setup such that instances are
  service-loaded and need to implement only a single method, thereby
  avoiding the need to align logic between multiple source code 
  locations.
- Extend the validation performed by the `Compilation` test helper
  class.
- Extend the `ErrorProneTestHelperSourceFormat` check to support source
  code passed to the `Compilation` test helper class.
```

PR #497, opened 2023-02-12
Introduce `StreamMapTo{Double,Int,Long}Sum` Refaster rules

giovannizotta:
I came across this pattern during a code review:
`.map(Stock::getQuantity).reduce(0, Integer::sum);`

This felt more readable to me:
`.mapToInt(Stock::getQuantity).sum();`

It can also be extended to `Long` and `Double`. Let me know what you think! 🙂 

Suggested commit message:
```
Introduce `StreamMapTo{Double,Int,Long}Sum` Refaster rules (#497)

As well as a new `IsLambdaExpressionOrMethodReference` matcher.
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -379,4 +380,16 @@ boolean after(Stream<T> stream) {
       return stream.allMatch(e -> test(e));
     }
   }
+
+  static final class StreamMapReduceIntSum<T> {
```
Stephan202: We try to name rules after the code in the `@AfterTemplate`, which in this case would yield:
```suggestion
  static final class StreamMapToIntSum<T> {
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -379,4 +380,16 @@ boolean after(Stream<T> stream) {
...
+    @BeforeTemplate
+    int before(Stream<T> stream, Function<? super T, Integer> mapper) {
+      return stream.map(mapper).reduce(0, Integer::sum);
+    }
+
+    @AfterTemplate
+    int after(Stream<T> stream, ToIntFunction<T> mapper) {
```
Stephan202, line 394: This rule assumes that the `mapper` matched by the `@BeforeTemplate` is a lambda expression or method reference, rather than a reference to an actual `Function<? super T, Integer>` instance. To guard against this we could introduce a new `Matcher` and pass it to a `@Matches` annotation. The alternative is to defer this work and add an `XXX` comment. Let us know if you're up for the former :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
Stephan202: (I was wondering whether this rule should be in the `IntStreamRules` class, but since `IntStream` is only an intermediate type, I suppose not.)

Stephan202 (commented):
Adding the `Double` and `Long` variants makes sense to me :).

I added a commit; suggested commit message:
```
Introduce `StreamMapTo{Double,Int,Long}Sum` Refaster rules (#497)
```

Stephan202:
Forgot to say: congrats and thanks for your first Error Prone Support contribution :tada:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
giovannizotta: I was wondering the same, and came to the same conclusion 🙂

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
giovannizotta: Sure, I'm up for it! I just need to take some time and figure out how to do it :)

Stephan202: Cool! A good place to start is to check out [these classes](https://github.com/PicnicSupermarket/error-prone-support/tree/master/refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers) :)

giovannizotta:
Thank you! 😊 

I added the `Matcher`, but I couldn't figure out how to make a test for it from [these tests](https://github.com/PicnicSupermarket/error-prone-support/tree/master/refaster-support/src/test/java/tech/picnic/errorprone/refaster/matchers). Do you have any pointers? 😅

rickie:
I found a related example that we used during my thesis project: [google/error-prone@`master...PicnicSupermarket:error-prone:rossendrijver/return_type`#diff-cd9d301320]([google/error-prone@`master...PicnicSupermarket:error-prone:rossendrijver/return_type`#diff-cd9d301320](https://github.com/google/error-prone/compare/master...PicnicSupermarket:error-prone:rossendrijver/return_type#diff-cd9d301320c36ceb0440830afe7652e10a516f6cb356f425fa0c1b32df3acf18R27-R)In the case63). The Matcher used there did some slightly different things and was "hacked together" a little 😬. 

I would write the Matcher to be in the same setup/style as the other matchers. Having a Matcher DELEGATE like this:
```
  private static final Matcher<ExpressionTree> DELEGATE =
      anyOf(isSameType(CHAR_TYPE), isSameType(Character.class));
```

The `DELEGATE` can also invoke a method that you would have in `IsLambdaExpressionOrMethodReference` where you can do the analysis. 

The `IsLambdaExpressionOrMethodReferenceTest` would have the following things;
- Test code just like the other Matcher tests.
- A TestMatcher that uses the `AbstractTestMatcher` to validate against the `DELEGATE`.

giovannizotta:
Thank you for your time Rick! That's definitely a nice pointer, I'll take a look at it when I have some time :)

giovannizotta:
I'm having fun! 😄 
[This](https://gist.github.com/giovannizotta/5157d499b28c03d092a8e8e8f9e9ab13#file-islamdaexpressionormethodreference-java) is the current state of the matcher, which I think resembles what you had in mind. 

However, I am having a bit of trouble testing it in the same way as the other `Matcher`s (like `IsCharacter`) are tested. Those are testing simple one-liners, while in this case, if I want to test that actual instances of `Function<T, {Double, Integer, Long}>` are matched, I need a multi-line test.

If I try something like [this](https://gist.github.com/giovannizotta/5157d499b28c03d092a8e8e8f9e9ab13#file-islambdaexpressionormethodreferencetest-java), it doesn't work because the matcher searches for matches in the whole test method, not only the `.map(function).reduce(0, Integer::sum)` part. Indeed, you can [see](https://gist.github.com/giovannizotta/5157d499b28c03d092a8e8e8f9e9ab13#file-log) that of course the matcher matches lambda expressions and method references, but not those that we are interested in. 

Do you have any idea on how to proceed? I bet there's a way to be able to handle this by taking a look at, for example, `ErrorProneTestHelperSourceFormatTest`, but I do not completely understand what's going on. I pushed a commit with the test disabled. Thank you in advance for your time! 🙂

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StreamRulesTestInput.java
```diff
@@ -138,4 +139,28 @@ ImmutableSet<Boolean> testStreamAllMatch() {
   boolean testStreamAllMatch2() {
     return Stream.of("foo").noneMatch(s -> !s.isBlank());
   }
+
+  ImmutableSet<Integer> testStreamMapToIntSum() {
+    Function<String, Integer> parseIntFunction = (String s) -> Integer.parseInt(s);
```
Stephan202: Just because it's shorter:
```suggestion
    Function<String, Integer> parseIntFunction = Integer::parseInt;
```

refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsLambdaExpressionOrMethodReference.java
```diff
@@ -0,0 +1,39 @@
...
+  /** Returns a matcher that matches lambda expressions. */
+  public static <T extends Tree> Matcher<T> isLambdaExpression() {
+    return (tree, state) -> tree instanceof LambdaExpressionTree;
+  }
+
+  /** Returns a matcher that matches method references. */
+  public static <T extends Tree> Matcher<T> isMethodReference() {
+    return (tree, state) ->
+        tree instanceof MemberReferenceTree
+            && ((MemberReferenceTree) tree).getMode().equals(INVOKE);
```
Stephan202: We can also allow `::new`, so the mode check can be omitted :)

refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsLambdaExpressionOrMethodReference.java
```diff
@@ -0,0 +1,39 @@
...
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.LambdaExpressionTree;
+import com.sun.source.tree.MemberReferenceTree;
+import com.sun.source.tree.Tree;
+
+/** A matcher of lambda expressions or method references. */
+public final class IsLambdaExpressionOrMethodReference implements Matcher<ExpressionTree> {
+  private static final long serialVersionUID = 1L;
+
+  private static final Matcher<ExpressionTree> DELEGATE =
```
Stephan202: While indeed _usually_ we introduce such a `DELEGATE` constant for performance reasons, in this case there's no efficiency gain over inlining the logic.

Stephan202 (commented):
Good question about how to test this stuff. The trick is so split the code over multiple lines. That said, this does indeed highlight a weakness of this kind of test: it's easy to over-match.

I rebased and added a commit with some improvements. Not yet done reviewing, but thought it'd be nice to show this intermediate state :)

(Functionally I think this now looks very good; will look a bit into minimizing the test without impacting coverage.)

refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsLambdaExpressionOrMethodReference.java
giovannizotta: Makes sense. I was thinking about the specific usage of this PR, but extending it to `::new` it makes it more reusable indeed :)

Stephan202 (approved):
Added one more commit. Very nice contribution @giovannizotta! :muscle:

```
Introduce `StreamMapTo{Double,Int,Long}Sum` Refaster rules (#497)

As well as a new `IsLambdaExpressionOrMethodReference` matcher.
```

giovannizotta:
Thank you @Stephan202 for the cleanup! It was the first time I dove into these concepts but I really enjoyed it, even though you ended up doing most of the work :smile: 

Also nice to see the intermediate state to know how to solve this kind of issue the next time :)

Stephan202:
Rest assured: this workflow is how it goes every time :smile:. It's in the nature of the project: tricky concepts, a large and unfamiliar API, and some rather opinionated maintainers ;)

If you're up for it, don't hesitate to contribute other changes if/when you observe them during code review :rocket:.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StreamRulesTestOutput.java
```diff
@@ -137,4 +138,28 @@ ImmutableSet<Boolean> testStreamAllMatch() {
...
+        Stream.of(1).mapToInt(i -> i * 2).sum(),
+        Stream.of("1").mapToInt(Integer::parseInt).sum(),
+        Stream.of("1").map(parseIntFunction).reduce(0, Integer::sum));
+  }
+
+  ImmutableSet<Double> testStreamMapToDoubleSum() {
+    Function<String, Double> parseDoubleFunction = Double::parseDouble;
+    return ImmutableSet.of(
+        Stream.of(1).mapToDouble(i -> i * 2.0).sum(),
+        Stream.of("1").mapToDouble(Double::parseDouble).sum(),
```
rickie: We generally use a different number or string per test case, see also further up in this file :). So let's change this.

rickie (approved):
@Stephan202, I'm curious, any particular reason to not add Javadoc for the Refaster rules? 

Added a commit with some suggestions.

Nice contribution @giovannizotta, cool stuff 🚀 !

Stephan202:
> @Stephan202, I'm curious, any particular reason to not add Javadoc for the Refaster rules?

Nope, just that the preceding rules also don't have any. :)

PR #501, opened 2023-02-16
Introduce ResultQueryFetchStream Refaster rule

maxsumrall:
When using the jOOQ SQL library for querying a database, you can either fetch the result set of your query into memory and proceed with application code, or you can return a lazy stream from the database and process the results as a stream. [See here for examples and explanations of fetching with jOOQ](https://blog.jooq.org/the-many-different-ways-to-fetch-data-in-jooq/). A common issue with the `.stream()` approach is that it is usually used unintentionally. This is dangerous as the stream returned here is not managed and not auto-closed, leaving applications potentially at risk to resource leaking. 

A less opinionated version of this rule may be to only make this replacement when the `.stream()` is invoked _outside of a try_with_resources_ block.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JooqRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JooqRules.java
@@ -1,0 +1,29 @@
+package tech.picnic.errorprone.refasterrules;
+
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.stream.Stream;
+import org.jooq.Record;
+import org.jooq.ResultQuery;
+import tech.picnic.errorprone.refaster.annotation.Description;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
+
+@OnlineDocumentation
+final class JooqRules {
+  private JooqRules() {}
+
+  /** Prefer eagerly fetching data over (lazy) streaming to avoid potentially leaking resources. */
+  @Description(
+      "Prefer eagerly fetching data over (lazy) streaming to avoid potentially leaking resources.")
+  static final class ResultQueryFetchStream {
+    @BeforeTemplate
+    Stream<?> before(ResultQuery<? extends Record> resultQuery) {
+      return resultQuery.stream();
+    }
+
+    @AfterTemplate
+    Stream<?> after(ResultQuery<? extends Record> resultQuery) {
+      return resultQuery.fetch().stream();
+    }
+  }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -369,6 +369,11 @@
                 <version>2.9.3</version>
             </dependency>
             <dependency>
+                <groupId>org.jooq</groupId>
+                <artifactId>jooq</artifactId>
+                <version>3.16.14</version>
+            </dependency>
+            <dependency>
                 <groupId>org.jspecify</groupId>
                 <artifactId>jspecify</artifactId>
                 <version>0.3.0</version>
@@ -1178,6 +1183,8 @@
                                 | BSD 3-clause
                                 | BSD License 3
                                 | Eclipse Distribution License (New BSD License)
+                                | Eclipse Distribution License - v 1.0
+                                | EDL 1.0
                                 | New BSD License
                             </licenseMerge>
                             <licenseMerge>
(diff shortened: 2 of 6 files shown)


maxsumrall:
rebased on `master`

pom.xml
```diff
@@ -368,6 +368,11 @@
                 <artifactId>value-annotations</artifactId>
                 <version>2.9.3</version>
             </dependency>
+            <dependency>
+                <groupId>org.jooq</groupId>
+                <artifactId>jooq</artifactId>
+                <version>3.16.14</version>
```
maxsumrall, line 374: This is the latest version of jOOQ supporting Java 11. From 3.17.x, it is not supported.

pom.xml
```diff
@@ -1178,6 +1183,8 @@
                                 | BSD 3-clause
                                 | BSD License 3
                                 | Eclipse Distribution License (New BSD License)
+                                | Eclipse Distribution License - v 1.0
```
maxsumrall, line 1186: Source: 
```[WARNING] License: 'Eclipse Distribution License - v 1.0' used by 1 dependencies:
 -Jakarta XML Binding API (jakarta.xml.bind:jakarta.xml.bind-api:3.0.0 - https://github.com/eclipse-ee4j/jaxb-api/jakarta.xml.bind-api)
```

pom.xml
```diff
@@ -1178,6 +1183,8 @@
...
+                                | EDL 1.0
```
maxsumrall, line 1187: Source:
```
[WARNING] License: 'EDL 1.0' used by 1 dependencies:
 -Jakarta Activation (com.sun.activation:jakarta.activation:2.0.0 - https://github.com/eclipse-ee4j/jaf/jakarta.activation)
```

ferdinand-swoboda (commented):
Very useful rule!

We can apply the same to the synonym `fetchStream()` and almost the same for the equally dangerous `fetchStreamInto(...)` variants.

> A less opinionated version of this rule may be to only make this replacement when the .stream() is invoked outside of a try_with_resources block.

That would certainly reduce false positives 👍🏼

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JooqRules.java
```diff
@@ -0,0 +1,29 @@
...
+import tech.picnic.errorprone.refaster.annotation.Description;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
+
+@OnlineDocumentation
+final class JooqRules {
+  private JooqRules() {}
+
+  /** Prefer eagerly fetching data over (lazy) streaming to avoid potentially leaking resources. */
+  @Description(
+      "Prefer eagerly fetching data over (lazy) streaming to avoid potentially leaking resources.")
```
ferdinand-swoboda, line 17: Would it be useful to link to the Jooq article here?

rickie:
> > A less opinionated version of this rule may be to only make this replacement when the .stream() is invoked outside of a try_with_resources block.
> 
> That would certainly reduce false positives 👍🏼

If that reduces the false positive rate I think it'd be smart to introduce a `Matcher` for that. The `Matcher` can literally check for the `try_with_resources` block. There are some examples [here](https://github.com/PicnicSupermarket/error-prone-support/tree/master/refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers). And of course, we can give extra pointers or discuss it if you're up for it.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JooqRules.java
maxsumrall: While it would be useful, it might not be wise as the blogpost might move URLs one day, and then the code is not correct. I think that would not be prudent in a library such as this. Perhaps linking to it in this PR is enough for curious folks to find one day if they go looking for the motivations for this rule.

maxsumrall:
> > > A less opinionated version of this rule may be to only make this replacement when the .stream() is invoked outside of a try_with_resources block.
> > 
> > 
> > That would certainly reduce false positives 👍🏼
> 
> If that reduces the false positive rate I think it'd be smart to introduce a `Matcher` for that. The `Matcher` can literally check for the `try_with_resources` block. There are some examples [here](https://github.com/PicnicSupermarket/error-prone-support/tree/master/refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers). And of course, we can give extra pointers or discuss it if you're up for it.

Thanks for the tip! I'll give it a try!

PR #513, opened 2023-02-26
Introduce `DirectReturn` check

Stephan202:
This PR replaces #1. I decided to pick this task up because the topic was raised after @rickie's recent Amsterdam JUG presentation. CC @SimonBaars.

Suggested commit message:
```
Introduce `DirectReturn` check (#513)
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
```diff
@@ -0,0 +1,101 @@
...
+ * statement.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Variable assignment is redundant; value can be returned directly",
+    link = BUG_PATTERNS_BASE_URL + "DirectReturn",
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = SIMPLIFICATION)
+public final class DirectReturn extends BugChecker implements BlockTreeMatcher {
```
Stephan202, line 43: Naming-wise, "something something unnecessary assignment" may be nicer. But (a) we're not performing a flow analysis that flags _all_ unnecessary variable assignments and (b) a name that accurately describes the subset covered would be quite unwieldy. Unless I'm just not creative enough, of course :upside_down_face:. Suggestions welcome.

Stephan202:
Just realized: ideally we do not suggest inlining if (a) the variable assigned to is of a different type than the method's return type and (b) this difference may influence the return value. Prime example here is Mockito's nullary `mock()` method (see also #454). I'll come back to this PR to see how we can best cover this case.

Stephan202:
Rebased and added a commit to cover this case. Did make it Mockito-specifc, because attempts at generalization became too contrived.

Stephan202 (commented):
Added one more commit.

rickie (commented):
Flushing one comment I had. Will continue later today.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Tells whether inlining the given expression to the associated return statement can be done
+   * safely.
+   *
+   * <p>Inlining is generally safe, but in rare cases the operation may have a functional impact.
+   * The sole case considered here is the inlining of a Mockito mock or spy construction without an
+   * explicit type. In such a case the type created depends on context, such as the method's return
+   * type.
```
werli, line 122: Great to have added this exclusion. Are we sure the exclusion is exhaustive however? My guess is: probably never _fully_, but this is a best-effort solution that should handle the majority of EPS users.

werli (approved):
The code itself and the added test cases were rather straightforward, nice work @Stephan202 :+1:

One more point about the exhaustiveness of the mockito exclusion, but I wouldn't expect more changes here as part of this PR. Hence I'm approving.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
Stephan202: Nope, this is not exhaustive, unfortunately. I did play with some more generic code, but it became way more complex than I could justify for this feature. So in the end gave up on that idea.

Stephan202 (commented):
Tnx for the reviews!

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
```diff
@@ -0,0 +1,164 @@
...
+            "    return variable;",
+            "  }",
+            "",
+            "  Object redundantSpyWithExplicitTypeVariableDeclaration() {",
+            "    // BUG: Diagnostic contains:",
+            "    String variable = spy(String.class);",
+            "    return variable;",
+            "  }",
+            "",
+            "  Object salienSpyTypeVariableDeclaration() {",
```
rickie:
```suggestion
            "  Object salientSpyTypeVariableDeclaration() {",
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreASTHelpers.java
```diff
@@ -46,4 +50,33 @@ public static ImmutableList<MethodTree> findMethods(CharSequence methodName, Vis
   public static boolean methodExistsInEnclosingClass(CharSequence methodName, VisitorState state) {
     return !findMethods(methodName, state).isEmpty();
   }
+
+  /**
+   * Returns the {@link MethodTree} from which control flow would exit if there would be a {@code
```
rickie:
```suggestion
   * Returns the {@link MethodTree} from which the control flow would exit if there would be a {@code
```
Sounds a bit better to me 🤔.

rickie (approved):
Really cool check 🚀.

Have some small comments and an extra test case or two. Let me know what you think 😄.

rickie:
Aii, misclicked and merged instead of rebased 😅.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreASTHelpers.java
```diff
@@ -46,4 +50,33 @@ public static ImmutableList<MethodTree> findMethods(CharSequence methodName, Vis
   public static boolean methodExistsInEnclosingClass(CharSequence methodName, VisitorState state) {
     return !findMethods(methodName, state).isEmpty();
   }
+
+  /**
+   * Returns the {@link MethodTree} from which the control flow would exit if there would be a
```
Stephan202, line 55: I think we should omit "the" here. (Cause this is not about a "definite" control flow.)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
```diff
@@ -0,0 +1,174 @@
...
+            "    if (true) {",
+            "      return variable;",
+            "    } else {",
+            "      // BUG: Diagnostic contains:",
+            "      variable = \"foo\";",
+            "      return variable;",
+            "    }",
+            "  }",
+            "",
+            "  Supplier<String> redundantAssignmentInLambda() {",
```
Stephan202, line 132:
```suggestion
            "  Supplier<String> redundantAssignmentInsideLambda() {",
```
Also matches the method above.

Stephan202 (commented):
Tnx; changes LGTM otherwise

rickie:
Applied and rebased. Will merge once 🟢, thanks for taking a look so quickly.

Stephan202:
Let's gooooo :rocket:

PR #523, opened 2023-03-07
Extend `null` check Refaster rules

### LINKED ISSUE #437: Rewrite `check{Argument,State}(object != null)` to `requireNonNull(object)`
### Problem

This is based on https://github.com/PicnicSupermarket/error-prone-support/pull/428#discussion_r1060430056. 

Let's rewrite expressions of `checkState` and `checkArgument` with a null check to a `requireNonNull` equivalent. 
There are multiple overloads for both methods that we should consider. 

- [ ] Improve performance.

```java
checkState(annotation != null);
checkState(annotation != null, "Annotation must be present"); 
checkArgument(uri.getHost() != null);
checkArgument(uri.getHost() != null, "Host should be present");
```

to:
```java
requireNonNull(annotation);
requireNonNull(annotation, "Annotation must be present"); 
requireNonNull(uri.getHost());
requireNonNull(uri.getHost(), "Host should be present");
```

### Considerations

Question:
What do we want with the `checkNotNull`? Do we prefer `requireNonNull` in all cases because it is in the JDK? 

In that case I'll extend the examples.

Ptijohn:
Follow up of this [issue](https://github.com/PicnicSupermarket/error-prone-support/issues/437).

Suggested commit message:
```
Extend `null` check Refaster rules (#523)

Summary of changes:
- Replace `CheckNotNull` with `RequireNonNull{,WithMessage}{,Statement}`.
- Extend `Is{,Not}Null`.

Fixes #437.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreASTHelpers.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreASTHelpers.java
@@ -2,6 +2,7 @@ package tech.picnic.errorprone.bugpatterns.util;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.Objects.requireNonNull;
 
 import com.google.common.collect.ImmutableList;
 import com.google.errorprone.VisitorState;
@@ -27,7 +28,7 @@ public final class MoreASTHelpers {
    */
   public static ImmutableList<MethodTree> findMethods(CharSequence methodName, VisitorState state) {
     ClassTree clazz = state.findEnclosing(ClassTree.class);
-    checkArgument(clazz != null, "Visited node is not enclosed by a class");
+    requireNonNull(clazz, "Visited node is not enclosed by a class");
     return clazz.getMembers().stream()
         .filter(MethodTree.class::isInstance)
         .map(MethodTree.class::cast)
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreASTHelpers.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreASTHelpers.java
@@ -1,7 +1,7 @@
 package tech.picnic.errorprone.bugpatterns.util;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
-import static java.util.Objects.requireNonNull;
 
 import com.google.common.collect.ImmutableList;
 import com.google.errorprone.VisitorState;
@@ -27,7 +27,7 @@ public final class MoreASTHelpers {
    */
   public static ImmutableList<MethodTree> findMethods(CharSequence methodName, VisitorState state) {
     ClassTree clazz = state.findEnclosing(ClassTree.class);
-    requireNonNull(clazz, "Visited node is not enclosed by a class");
+    checkArgument(clazz != null, "Visited node is not enclosed by a class");
     return clazz.getMembers().stream()
         .filter(MethodTree.class::isInstance)
         .map(MethodTree.class::cast)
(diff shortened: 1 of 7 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PreconditionsRules.java
```diff
@@ -78,7 +79,7 @@ void after(int index, int size, String message) {
   static final class RequireNonNull<T> {
     @BeforeTemplate
     void before(T object) {
-      if (object == null) {
+      if (Refaster.anyOf(object == null, null == object)) {
```
rickie: Instead of having to add the inverse check, we can add another Refaster rule that rewrites `null == object` to `object == null`. That allows us to simplify the code in many other places.

rickie (approved):
Added a commit.

Changes LGTM 😄. Added an extra template to make our lives easier. We can make more assumptions about the code and we need to account for less cases.

rickie:
Rebased and tweaked suggested commit message.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/NullRules.java
```diff
@@ -25,6 +25,11 @@ private NullRules() {}
   static final class IsNull {
     @BeforeTemplate
     boolean before(@Nullable Object object) {
+      return null == object;
+    }
+
+    @BeforeTemplate
+    boolean before2(@Nullable Object object) {
       return Objects.isNull(object);
     }
```
Stephan202, line 34: Here and below: we can use `Refaster.anyOf`.

Stephan202 (commented):
Rebased and added a commit. Open point is about the `IllegalArgumentException` changes; IMHO those should be omitted.

rickie (commented):
Added a commit and answered the open points :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PreconditionsRules.java
```diff
@@ -72,35 +74,73 @@ void after(int index, int size, String message) {
...
+    }
+
+    @AfterTemplate
+    @UseImportPolicy(STATIC_IMPORT_ALWAYS)
+    T after(T object) {
+      return requireNonNull(object);
+    }
+  }
+
+  /** Prefer {@link Objects#requireNonNull(Object)} over non-JDK alternatives. */
```
Stephan202, line 91:
```suggestion
  /** Prefer {@link Objects#requireNonNull(Object)} over more verbose or non-JDK alternatives. */
```
But, assuming we drop the `checkArgument` check:
```suggestion
  /** Prefer {@link Objects#requireNonNull(Object)} over more verbose alternatives. */
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreASTHelpers.java
```diff
@@ -27,7 +27,7 @@ private MoreASTHelpers() {}
    */
   public static ImmutableList<MethodTree> findMethods(CharSequence methodName, VisitorState state) {
     ClassTree clazz = state.findEnclosing(ClassTree.class);
-    checkArgument(clazz != null, "Visited node is not enclosed by a class");
+    requireNonNull(clazz, "Visited node is not enclosed by a class");
```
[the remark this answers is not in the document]
Stephan202: Dropping is the safe way forward :+1:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/NullRules.java
Stephan202: Indeed we should ;)

rickie (commented):
Added a commit and rebased.

rickie:
One open question is how to address the "Statement" part of the Refaster rules in the suggested commit message. I'm not sure if it adds value to put it there. Any suggestions @Stephan202 ?

PR #526, opened 2023-03-10
Introduce `BugCheckerRule` refaster rules

Ptijohn:
Suggested commit message:
```
Introduce `BugCheckerRule` refaster rules (#XXX)
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BugCheckerRules.java
```diff
@@ -0,0 +1,30 @@
...
+import com.google.errorprone.BugCheckerRefactoringTestHelper;
+import com.google.errorprone.BugCheckerRefactoringTestHelper.FixChooser;
+import com.google.errorprone.BugCheckerRefactoringTestHelper.FixChoosers;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
+
+/** Refaster rules related to BugChecker classes. */
+@OnlineDocumentation
+final class BugCheckerRules {
```
Ptijohn, line 12: Honestly not sure about the reach of this class 🤔 
I could also limit it to `BugCheckerRefactoringTestHelper`, but then it's also quite closed.
That would also impact docs and all.

Stephan202 (approved):
```
Introduce `BugCheckerRules` Refaster rule collection (#526)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BugCheckerRules.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  /**
+   * Prefer {@link BugCheckerRefactoringTestHelper.ExpectOutput#expectUnchanged()} over repeating
+   * the input.
+   */
+  // XXX: This rule assumes that the full source code is specified as a single string, e.g. using a
+  // text block. Support for multi-line source code input would require a `BugChecker`
+  // implementation instead.
```
Stephan202, line 39: Right now this rule won't match, but after #198 it'll be relevant.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BugCheckerRules.java
Ptijohn: Hahaha, yeah I saw this class and hesitated, but felt like doing this was cleaner ;) 
Thanks for the added rule 💪

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BugCheckerRules.java
Ptijohn: Thanks @Stephan202 💪

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BugCheckerRules.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
+
+/** Refaster rules related to {@link com.google.errorprone.bugpatterns.BugChecker} classes. */
+@OnlineDocumentation
+final class BugCheckerRules {
+  private BugCheckerRules() {}
+
+  /**
+   * Avoid calling {@link BugCheckerRefactoringTestHelper#setFixChooser(FixChooser)} with the
```
rickie: We should update this to also account for `setImportOrder`. Will propose something :).

rickie (approved):
Added a commit with Javadoc tweak and rebased.

Maybe we can consider adding something for `BugCheckerRefactoringTestHelper#setArgs` for the case where we pass an `ImmutableList` and have only one string. Could be nicer to drop `immutableList` in that case. If the `setArgs` is called with multiple strings we can opt for the `ImmutableList` variant. WDYT? (out of scope in any case though)

Apart from that, it looks really good 👍🏻, nice to see that it drops a few of these occurrences 😄. 

Thanks for picking up @Ptijohn 🚀 !

Stephan202:
Changes LGTM :+1:

W.r.t. `setArgs`: yes, that one is on my list of things to do. That case is not really Refaster-compatible (using `@Repeated` won't work); instead I'm thinking of introducing a bug checker that simplifies unary methods calls with an `ImmutableList.of`/`Arrays.asList` (or similar) argument, for which a varargs overload exists.

PR #527, opened 2023-03-10
Rewrite `Optional#orElse` to `Optional#orElseGet` for non-constant expressions

mlrprananta:
❗~~This is now on top of #599~~  ❗ 

Following an internal GH thread it felt fitting to write a rule to cover this case. 

This PR proposes a Refaster rule to rewrite `Optional#orElse` to `Optional#orElseGet` for method invocation expressions, (with 2 or more args), with performance as the main benefit.

```
Introduce `OptionalOrElseGet` Refaster rule (#527)

While there, introduce `IsLikelyTrivialComputation` Matcher.
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -220,6 +222,22 @@ T after(Optional<T> o1, Optional<T> o2) {
     }
   }
 
+  /**
+   * Prefer {@link Optional#orElseGet(Supplier)} over {@link Optional#orElse(Object)} if the given
+   * value is not a compile-time constant.
+   */
+  abstract static class OrElseToOrElseGet<T> {
```
Stephan202:
```suggestion
  static final class OrElseToOrElseGet<T> {
```
(This one is also suggested when running `./apply-error-prone-suggestions.sh`.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -220,6 +222,22 @@ T after(Optional<T> o1, Optional<T> o2) {
...
+    @BeforeTemplate
+    T before(Optional<T> o, @NotMatches(CompileTimeConstantExpressionMatcher.class) T value) {
```
Stephan202, line 231: So it seems that perhaps `CompileTimeConstantExpressionMatcher` is a bit too strict. When we run `apply-error-prone-suggestions.sh` this diff is also generated:
```diff
diff --git a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/ConflictDetectionTest.java b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/ConflictDetectionTest.java
index 70282dd4..5b430448 100644
--- a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/ConflictDetectionTest.java
+++ b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/ConflictDetectionTest.java
@@ -67,7 +67,7 @@ final class ConflictDetectionTest {
       return ConflictDetection.findMethodRenameBlocker(
               ASTHelpers.getSymbol(tree), tree.getName() + "t", state)
           .map(blocker -> buildDescription(tree).setMessage(blocker).build())
-          .orElse(Description.NO_MATCH);
+          .orElseGet(() -> Description.NO_MATCH);
     }
   }
 }
```

Similarly, as-is the rule triggers the following violation:
```
[WARN] /home/sschroevers/workspace/picnic/error-prone-support/refaster-support/src/main/java/tech/picnic/errorprone/refaster/AnnotatedCompositeCodeTransformer.java:[124,80] [Refaster Rule] OptionalRules.OrElseToOrElseGet: Refactoring opportunity
    (see https://error-prone.picnic.tech/refasterrules/OptionalRules#OrElseToOrElseGet)
  Did you mean 'return getAnnotationValue(Severity.class, Severity::value, delegate).orElseGet(()->SUGGESTION);'?
```

We might need a separate `Matcher` that allows any constant and field dereference :thinking:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -220,6 +222,22 @@ T after(Optional<T> o1, Optional<T> o2) {
     }
   }
 
+  /**
+   * Prefer {@link Optional#orElseGet(Supplier)} over {@link Optional#orElse(Object)} if the given
+   * value is not a compile-time constant.
```
Stephan202, line 227: I'll need to think a bit about phrasing, but likely we should call out that this rule can, in rare circumstances, change application behavior. (This can happen if `value` matches a non-pure expression.)

Stephan202 (commented):
Clearly we need to do a bit more work here. To complicate matters further, this rule will introduce quite a lot of lambda expressions that we'd like to write as a method reference instead. So I wonder whether we should first finalize the (very hard) task of getting [`MethodReferenceUsage`](https://github.com/PicnicSupermarket/error-prone-support/blob/c3cd535b16967d5ebb8e3cd8c953ede3c50cd7e9/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MethodReferenceUsage.java) production ready. (I have a [branch](https://github.com/PicnicSupermarket/error-prone-support/tree/sschroevers/MethodReferenceUsageCheck-improvements) for that, but realistically it'll be months (or longer...) before I'll get back to that).

Added a small commit.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
mlrprananta: Agreed, after running the script I did notice a lot of changes and some seemed unintended, in the end we're looking to prevent potentially expensive method invocations used with `orElse`, perhaps it would be a better to match based on method invocations instead?

mlrprananta:
Hey @rickie, between those 2 options I would prefer the second route for now, as I don't feel comfortable not enabling it internally. Second route would be a easier to cover as well initially and as you mentioned we can improve it later on!

rickie:
@mlrprananta sounds good to me! Let's go for that option 😄.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -220,6 +223,22 @@ T after(Optional<T> o1, Optional<T> o2) {
...
+   */
+  static final class OrElseToOrElseGet<T> {
+    @BeforeTemplate
+    T before(Optional<T> o, @Matches(IsMethodInvocationWithTwoOrMoreArgs.class) T value) {
```
mlrprananta, line 232: TODO: add `XXX`

refaster-support/src/test/java/tech/picnic/errorprone/refaster/matchers/IsMethodInvocationWithTwoOrMoreArgsTest.java
```diff
@@ -0,0 +1,73 @@
...
+            "",
+            "  String positive2() {",
+            "    // BUG: Diagnostic contains:",
+            "    return String.format(\"%s\", \"foo\");",
+            "  }",
+            "}")
+        .doTest();
+  }
+
+  /** A {@link BugChecker} that simply delegates to {@link IsCharacter}. */
```
mlrprananta, line 59:
```suggestion
  /** A {@link BugChecker} that simply delegates to {@link IsMethodInvocationWithTwoOrMoreArgs}. */
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -158,7 +160,8 @@ Optional<T> after(T input) {
   static final class MapOptionalToBoolean<T> {
     @BeforeTemplate
     boolean before(Optional<T> optional, Function<? super T, Boolean> predicate) {
-      return optional.map(predicate).orElse(Refaster.anyOf(false, Boolean.FALSE));
+      return Refaster.anyOf(
+          optional.map(predicate).orElse(false), optional.map(predicate).orElse(Boolean.FALSE));
```
mlrprananta, line 164: Had to rewrite this as this was causing a "false" positive, `orElse(Refaster.anyOf(false, Boolean.FALSE))` got rewritten to `orElseGet(() -> Refaster.anyOf(false, Boolean.FALSE))`

refaster-support/src/test/java/tech/picnic/errorprone/refaster/matchers/IsMethodInvocationWithTwoOrMoreArgsTest.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.CompilationTestHelper;
+import com.google.errorprone.bugpatterns.BugChecker;
+import org.junit.jupiter.api.Test;
+
+final class IsMethodInvocationWithTwoOrMoreArgsTest {
+  @Test
+  void matches() {
+    CompilationTestHelper.newInstance(
+            IsMethodInvocationWithTwoOrMoreArgsTest.MatcherTestChecker.class, getClass())
```
rickie:
```suggestion
            MatcherTestChecker.class, getClass())
```
We can drop the qualifier.

refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsMethodInvocationWithTwoOrMoreArgs.java
```diff
@@ -0,0 +1,27 @@
...
+import com.google.errorprone.matchers.Matcher;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+
+/** A matcher for method invocations with two or more arguments. */
+public final class IsMethodInvocationWithTwoOrMoreArgs implements Matcher<ExpressionTree> {
+  private static final long serialVersionUID = 1L;
+
+  /** Instantiates a new {@link IsMethodInvocationWithTwoOrMoreArgs} instance. */
+  public IsMethodInvocationWithTwoOrMoreArgs() {}
```
rickie, line 13: Interesting, [here](url) we discussed that we want to check for two MethodInvocations instead of one method invocation wit two arguments. This is not incorrect though. I think it is nice to combine the two things into one `Matcher` and call it something like: `IsChainedMethodCallOrHasAtLeastTwoArgs`. I tried implementing this but had some troubles with the `VisitorState` in the `matches` method. 

In the `AbstractMatcherTestChecker` we only pass the `VisitorState` from the `CompilationUnit` which is not the state that we want. So we need to probably change that setup a bit. For now I'll flush the changes that I have.

refaster-support/src/test/java/tech/picnic/errorprone/refaster/matchers/IsMethodInvocationWithTwoOrMoreArgsTest.java
```diff
@@ -0,0 +1,73 @@
...
+            "  }",
+            "",
+            "  String positive1() {",
+            "    // BUG: Diagnostic contains:",
+            "    return foo(\"foo\", \"bar\");",
+            "  }",
+            "",
+            "  String positive2() {",
+            "    // BUG: Diagnostic contains:",
+            "    return String.format(\"%s\", \"foo\");",
```
rickie, line 40: We use the word "or more" for the Matcher, so we should add a test case :).

Will push some tweaks to simplify a little.

refaster-support/src/test/java/tech/picnic/errorprone/refaster/matchers/IsMethodInvocationWithTwoOrMoreArgsTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+final class IsMethodInvocationWithTwoOrMoreArgsTest {
+  @Test
+  void matches() {
+    CompilationTestHelper.newInstance(
+            IsMethodInvocationWithTwoOrMoreArgsTest.MatcherTestChecker.class, getClass())
+        .addSourceLines(
+            "A.java",
+            "class A {",
+            "  String foo(String foo, String bar) {",
```
rickie: Let's move these to the bottom :).

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/OptionalRulesTestInput.java
```diff
@@ -119,4 +119,12 @@ ImmutableSet<Optional<String>> testOptionalFilter() {
   Optional<String> testOptionalMap() {
     return Optional.of(1).stream().map(String::valueOf).findAny();
   }
+
+  ImmutableSet<String> testOptionalOrElseGet() {
+    return ImmutableSet.of(
+        Optional.of("foo").orElse("bar"),
```
rickie, line 125: We make sure to use distinct values for every statement :).

rickie (commented):
Rebased and added a commit. There is the generalization of the Matcher left to also account for multiple method invocations :). 

Cool stuff 🚀 !

refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsMethodInvocationWithTwoOrMoreArgs.java
mlrprananta: > Interesting, [here](https://github.com/PicnicSupermarket/error-prone-support/pull/url) we discussed that we want to check for two MethodInvocations instead of one method invocation wit two arguments.

Ah, that was a misunderstanding on my part, but indeed both could apply for this situation. I can look into it a bit more to implement a (combined) matcher for chained method calls as well

mlrprananta: Hey @rickie, few questions:
1. Could you elaborate on the `VisitorState` issue you mentioned? Which state should we be using instead?
2. Could nested method invocations also apply for this situation? So something like:
```java
m1(m2("param"))
```

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/OptionalRulesTestInput.java
mlrprananta: Does it need to be distinct for mutation testing?

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
Stephan202: Ah, ideally the new matcher ignores `Refaster.anyOf` calls, since those aren't _true_ method invocations.

refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsMethodInvocationWithTwoOrMoreArgs.java
Stephan202: Indeed, the matcher shouldn't require >= 2 arguments: `foo().bar()` doesn't have any arguments, yet should be matched. One option is to use a `TreeScanner` (see examples in the code base) to count the number of `MethodInvocationTree`s.

IIUC the idea is to match non-trivial expressions that can't (or shouldn't) be replaced with a method reference. In this case we know that we're rewriting the expression to a `Supplier<T>`. So perhaps we're looking for a matcher of expressions which:
1. Contain more than one method invocation; or
2. Contain at least one method invocation with _one_ or more arguments.

(Invocations of a single nullary method such as `foo()`, `x.foo()` or `x.y.foo()` should not be matched for now, since in those case a method reference may be preferable.)

**That said**, expressions of the form `42 * foo()` should also be deferred.... So perhaps we should instead introduce an `IsLikelyTrivialComputation` matcher that  _does_ match `true`, `x`, `foo()`, `x.foo()`, `x.y.foo()` etc., and then have the Refaster rule use `@NotMatches(IsLikelyTrivialComputation.class)`. Then later once `MethodReferenceUsage` works properly, we can replace this matcher with a simpler `RequiresComputation` matcher and then use `@Matches(RequiresComputation.class)`.

Stephan202: > In the `AbstractMatcherTestChecker` we only pass the `VisitorState` from the `CompilationUnit` which is not the state that we want. So we need to probably change that setup a bit. For now I'll flush the changes that I have.

I have a fix for this on a branch somewhere; let me try to open a PR for this ~soon.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/OptionalRulesTestInput.java
Stephan202: Nope, it's about making sure that Refaster properly moves things around (which is easier to see if all values are distinct). Unfortunately for technical reasons our Refaster tests aren't exercised by Pitest.

Stephan202 (commented):
Interesting discussions!

rickie:
Nice answers @Stephan202 😄, curious to see the PR 🚀 !

refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsMethodInvocationWithTwoOrMoreArgs.java
Stephan202: I filed #599.

rickie: Nice examples @mlrprananta ! 

I checked in IntelliJ and saw that it mentions this optimization once you press `alt+enter` on the specific method. However, it doesn't highlight the method invocation to indicate you _should_ rewrite it to a method reference, which for simple cases IntelliJ does. Because for the more difficult ones it doesn't do that, I would say it is fine for now to consider these things as a non trivial computation and we can rewrite them. Probably the last three examples will not occur that often. 

So I think the fist three examples are on the "edge" of being a "trivial" computation... I'm not sure about those. The others can definitely be rewritten in the context of this PR.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/OptionalRulesTestInput.java
```diff
@@ -119,4 +119,13 @@ ImmutableSet<Optional<String>> testOptionalFilter() {
...
+        Optional.of("baz").orElse(toString()),
+        Optional.of("qux").orElse(String.valueOf(true)),
+        Optional.of("quux").orElse(String.format("%s", "quuz")),
+        Optional.of("corge").orElse(String.format("%s", "grault", "garply")));
```
Stephan202: I guess we can do with fewer test cases :)

refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsMethodInvocationWithTwoOrMoreArgs.java
Stephan202: Alright, so I had a close look at this, and pushed a proposal :upside_down_face:

Stephan202 (approved):
Added a commit; PTAL :)

(I'd be nice to get some reviews of the base PR :upside_down_face:)

refaster-support/src/test/java/tech/picnic/errorprone/refaster/matchers/IsLikelyTrivialComputationTest.java
```diff
@@ -0,0 +1,147 @@
...
+            "    return \"foo\".getBytes()[0];",
+            "  }",
+            "",
+            "  int negative5() {",
+            "    int[] arr = new int[0];",
+            "    return arr[hashCode()];",
+            "  }",
+            "",
+            "  int negative6() {",
+            "    return 1 * 2;",
```
rickie, line 42: For the `BinaryTree` we can check if both left and right are an `LiteralTree` or `IdentifierTree` as start? Also fine with leaving this as is.

rickie (approved):
One consideration, we already discussed that we don't want too many edge cases for this Matcher, so maybe it's already fine like this.

Anyway, really nice improvement! Curious to see how this performs on our codebase. 🚀 !

refaster-support/src/test/java/tech/picnic/errorprone/refaster/matchers/IsLikelyTrivialComputationTest.java
Stephan202: Yeah, one could argue that this _is_ a trivial computation. But then maybe so is `4 * 60 * 60` ("four hours"), which still wouldn't be flagged. And since the matcher is already quite complex, I'm not sure how far to go for diminishing returns. Even this case I don't expect to be triggered too frequently (at least in our internal code base).

Stephan202 (approved):
Tnx for the review @rickie. @mlrprananta we need one more review of #599 before we can merge this one. Are you up for that?

rickie:
Fair point --> Rebased --> Will merge.

rickie:
Updated the suggested commit message.

Stephan202:
The second like seems now less important than perhaps it is, but I lack creativity for a better proposal. Let's go :rocket:

rickie:
Yeah you are right, I tried combining them but the name is too long 😅.

PR #541, opened 2023-03-22
Extend `StepVerifierStepIdentity` Refaster rule

mohamedsamehsalah:
Following [this suggestion](https://github.com/PicnicSupermarket/picnic-platform/pull/9982#discussion_r1133960563).

Suggested commit message
```
Drop usage of redundant `expectNextCount(0)` checks
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1232,6 +1232,24 @@ final class ReactorRules {
     }
   }
 
+  /**
+   * Prefer {@link StepVerifier.LastStep#verifyComplete()} over {@code
+   * StepVerifier.Step#expectNextCount(0).verifyComplete()} or its alternatives.
+   */
+  static final class StepVerifierStepExpectNextCountZeroVerifyComplete<T> {
+    @BeforeTemplate
+    Duration before(StepVerifier.Step<T> step) {
+      return Refaster.anyOf(
+          step.expectNextCount(0L).expectComplete().verify(),
+          step.expectNextCount(0L).verifyComplete());
+    }
+
+    @AfterTemplate
+    Duration after(StepVerifier.LastStep step) {
+      return step.verifyComplete();
+    }
+  }
+
   /** Prefer {@link StepVerifier.LastStep#verifyComplete()} over more verbose alternatives. */
   static final class StepVerifierLastStepVerifyComplete {
     @BeforeTemplate
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1205,11 +1205,13 @@ final class ReactorRules {
   }
 
   /** Don't unnecessarily have {@link StepVerifier.Step} expect no elements. */
+  // XXX: Given an `IsEmpty` matcher that identifies a wide range of guaranteed-empty `Iterable`
+  // expressions, consider also simplifying `step.expectNextSequence(someEmptyIterable)`.
   static final class StepVerifierStepIdentity<T> {
     @BeforeTemplate
     @SuppressWarnings("unchecked")
     StepVerifier.Step<T> before(StepVerifier.Step<T> step) {
-      return Refaster.anyOf(step.expectNext(), step.expectNextCount(0L));
+      return Refaster.anyOf(step.expectNext(), step.expectNextCount(0));
     }
 
     @AfterTemplate
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1204,12 +1204,15 @@ final class ReactorRules {
     }
   }
 
-  /** Don't unnecessarily call {@link StepVerifier.Step#expectNext(Object[])}. */
+  /**
+   * Don't unnecessarily call {@link StepVerifier.Step#expectNext(Object[])} or {@code
+   * StepVerifier.Step#expectNextCount(0L)}.
+   */
   static final class StepVerifierStepExpectNextEmpty<T> {
     @BeforeTemplate
     @SuppressWarnings("unchecked")
     StepVerifier.Step<T> before(StepVerifier.Step<T> step) {
-      return step.expectNext();
+      return Refaster.anyOf(step.expectNext(), step.expectNextCount(0L));
     }
 
     @AfterTemplate
@@ -1229,19 +1232,6 @@ final class ReactorRules {
     @AfterTemplate
     StepVerifier.Step<T> after(StepVerifier.Step<T> step, T object) {
       return step.expectNext(object);
-    }
-  }
-
-  /** Remove redundant {@code StepVerifier.Step#expectNextCount(0)} checks. */
-  static final class StepVerifierStepExpectNextCountZero<T> {
-    @BeforeTemplate
-    StepVerifier.Step<T> before(StepVerifier.Step<T> step) {
-      return step.expectNextCount(0L);
-    }
-
-    @AfterTemplate
-    StepVerifier.Step<T> after(StepVerifier.Step<T> step) {
-      return step;
     }
   }
 
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1204,10 +1204,7 @@ final class ReactorRules {
     }
   }
 
-  /**
-   * Don't unnecessarily call {@link StepVerifier.Step#expectNext(Object[])} or {@code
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1232,6 +1232,24 @@ StepVerifier.Step<T> after(StepVerifier.Step<T> step, T object) {
...
 
+  /**
+   * Prefer {@link StepVerifier.LastStep#verifyComplete()} over {@code
+   * StepVerifier.Step#expectNextCount(0).verifyComplete()} or its alternatives.
+   */
+  static final class StepVerifierStepExpectNextCountZeroVerifyComplete<T> {
+    @BeforeTemplate
+    Duration before(StepVerifier.Step<T> step) {
+      return Refaster.anyOf(
+          step.expectNextCount(0L).expectComplete().verify(),
```
mohamedsamehsalah, line 1243: Do we need to check for `...expectComplete().verify()` as it is already checked in `StepVerifierLastStepVerifyComplete` class ?

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1232,6 +1232,24 @@ StepVerifier.Step<T> after(StepVerifier.Step<T> step, T object) {
...
+          step.expectNextCount(0L).verifyComplete());
```
werli, line 1244: I think we're not yet generic enough: Ideally, we rewrite:
```java
          step.expectNextCount(0L)
```
to just 
```java
          step
```

After all, it's not guaranteed that the next call will be `verifyComplete()`. :eyes:

werli (commented):
Nice first contribution. :muscle: 

I've left two comments.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
mohamedsamehsalah: Perfect ❤️ 
Thanks for heads-up 🕺

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
mohamedsamehsalah: Build error gave the same [answer](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/4489690922/jobs/7895839277#step:6:2231) :trollface:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1232,6 +1232,19 @@ StepVerifier.Step<T> after(StepVerifier.Step<T> step, T object) {
...
+    @BeforeTemplate
+    StepVerifier.Step<T> before(StepVerifier.Step<T> step) {
+      return step.expectNextCount(0L);
+    }
+
+    @AfterTemplate
+    StepVerifier.Step<T> after(StepVerifier.Step<T> step) {
+      return step;
+    }
+  }
```
werli, line 1246: Sorry, I just noticed: This rule and `StepVerifierStepExpectNextEmpty` have the same `after` step. We should join those into one.
Usually, these rules are named by the resulting code, see other examples in this class. Since there is "no" resulting code here, I'd then name this one `StepVerifierStepIdentity`. :+1:

werli (commented):
:pinching_hand:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
werli: @mohamedsamehsalah Let's also rename?

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1204,12 +1204,15 @@ StepVerifier.FirstStep<? extends T> after(Flux<T> flux) {
     }
   }
 
-  /** Don't unnecessarily call {@link StepVerifier.Step#expectNext(Object[])}. */
-  static final class StepVerifierStepExpectNextEmpty<T> {
+  /**
+   * Don't unnecessarily call {@link StepVerifier.Step#expectNext(Object[])} or {@code
+   * StepVerifier.Step#expectNextCount(0L)}.
+   */
```
werli, line 1210: Since we'll most likely extend this in the future:
```suggestion
  /** Don't unnecessarily have {@link StepVerifier.Step} expect no elements. */
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1204,12 +1204,15 @@ StepVerifier.FirstStep<? extends T> after(Flux<T> flux) {
...
+  static final class StepVerifierStepIdentity<T> {
     @BeforeTemplate
     @SuppressWarnings("unchecked")
     StepVerifier.Step<T> before(StepVerifier.Step<T> step) {
-      return step.expectNext();
+      return Refaster.anyOf(step.expectNext(), step.expectNextCount(0L));
```
werli, line 1215: Technically, empty `Iterable`s in a call to `StepVerifier.Step#expectNextSequence(Iterable<T>)` would also qualify, see suggestion.

```suggestion
      return Refaster.anyOf(
          step.expectNext(), step.expectNextCount(0L), step.expectNextSequence(ImmutableSet.of()));
```

However, IIUC we'd need to have a different `before` step which matches empty collections. Let's not apply this as part of this PR and keep it bite-sized. :wink:

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -389,8 +389,10 @@ StepVerifier.FirstStep<Integer> testStepVerifierFromFlux() {
     return StepVerifier.create(Flux.just(1));
   }
 
-  StepVerifier.Step<Integer> testStepVerifierStepExpectNextEmpty() {
-    return StepVerifier.create(Mono.just(0)).expectNext();
+  ImmutableSet<StepVerifier.Step<Integer>> testStepVerifierStepIdentity() {
+    return ImmutableSet.of(
+        StepVerifier.create(Mono.just(0)).expectNext(),
+        StepVerifier.create(Mono.just(0)).expectNextCount(0L));
   }
```
werli, line 396: (a) All the other examples contain / start with `Mono.just(1)`. Let's simply also apply this here. :+1:
(b) Cases in which there are multiple input cases, they'll be represented by different values. More concretely, I suggest this:
```suggestion
  ImmutableSet<StepVerifier.Step<Integer>> testStepVerifierStepIdentity() {
    return ImmutableSet.of(
        StepVerifier.create(Mono.just(1)).expectNext(),
        StepVerifier.create(Mono.just(2)).expectNextCount(0L));
  }
```

werli (approved):
I've pushed my suggestions. :+1: Thanks for picking this up @mohamedsamehsalah :muscle:

Suggested commit message to reflect actual changes and stay in line with project best practices:
```
Extend `StepVerifierStepIdentity` Refaster rule (#541)

By flagging expressions of the form `step.expectNextCount(0L)`.
```

PS: For the future: Best to avoid merge commits. Rebasing it it. :rocket:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1204,12 +1204,12 @@ StepVerifier.FirstStep<? extends T> after(Flux<T> flux) {
...
 
-  /** Don't unnecessarily call {@link StepVerifier.Step#expectNext(Object[])}. */
-  static final class StepVerifierStepExpectNextEmpty<T> {
+  /** Don't unnecessarily have {@link StepVerifier.Step} expect no elements. */
+  static final class StepVerifierStepIdentity<T> {
     @BeforeTemplate
     @SuppressWarnings("unchecked")
     StepVerifier.Step<T> before(StepVerifier.Step<T> step) {
-      return step.expectNext();
+      return Refaster.anyOf(step.expectNext(), step.expectNextCount(0L));
```
Stephan202: I see that both `0` and `0L` cover both `int`s and `long`s. We can use the `int` here, and then test against the `long`.
```suggestion
      return Refaster.anyOf(step.expectNext(), step.expectNextCount(0));
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: Nice one. This relies on an `IsEmpty` `Matcher` I've been meaning to finalize (see also a comment elsewhere in this class). Will add a note :)

Stephan202 (approved):
Added one more commit. Thanks @mohamedsamehsalah for opening this PR and thanks @werli for the thorough review!

rickie (approved):
Changes LGTM! Thanks for the review @werli 🚀 

Nice first contribution @mohamedsamehsalah 🥳, that sticker is coming your way 😉!!!

PR #568, opened 2023-04-06
Have `DirectReturn` check consider `finally` blocks

Stephan202:
Resolves an issue reported by @francisco-f-silva.

Suggested commit message:
```
Have `DirectReturn` check consider `finally` blocks (#568)
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
```diff
@@ -113,13 +122,13 @@ private static Optional<ExpressionTree> tryMatchAssignment(Symbol targetSymbol,
...
-   * <p>Inlining is generally safe, but in rare cases the operation may have a functional impact.
-   * The sole case considered here is the inlining of a Mockito mock or spy construction without an
-   * explicit type. In such a case the type created depends on context, such as the method's return
-   * type.
+   * <p>Inlining an expression generally does not change its return type, but in rare cases the
+   * operation may have a functional impact. The sole case considered here is the inlining of a
+   * Mockito mock or spy construction without an explicit type. In such a case the type created
+   * depends on context, such as the method's return type.
    */
   private static boolean canInlineToReturnStatement(
```
Stephan202, line 133: Should now find a better name for this method. Suggestions welcome.

rickie (approved):
Thanks for fixing the issue so fast after it popped up :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
Stephan202: Hmm, my concern here is more that `canInline` is actually too strong, since now this method only tests are required, but not sufficient set of conditions to be met. I guess we can just  merge as-is.

Stephan202 (commented):
Added a commit. Tnx for the review!

rickie:
Changed this to a bug as it was reported as such :).

PR #570, opened 2023-04-07
Introduce `FluxCollectToImmutableList` Refaster rule

mlrprananta:
Based on internal inquiry

Suggested commit message:
```
Introduce `FluxCollectToImmutableList` Refaster rule (#570)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1170,6 +1170,23 @@ final class ReactorRules {
   }
 
   /**
+   * Prefer {@link Flux#collect(Collector)} with {@link ImmutableList#toImmutableList} over {@link
+   * Flux#collectList()}.
+   */
+  static final class FluxCollectListToCollectImmutableList<T> {
+    @BeforeTemplate
+    Mono<List<T>> before(Flux<T> flux) {
+      return flux.collectList();
+    }
+
+    @AfterTemplate
+    @UseImportPolicy(STATIC_IMPORT_ALWAYS)
+    Mono<List<T>> after(Flux<T> flux) {
+      return flux.collect(toImmutableList());
+    }
+  }
+
+  /**
    * Prefer {@link Flux#collect(Collector)} with {@link ImmutableSet#toImmutableSet()} over more
    * contrived alternatives.
    */
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1176,12 +1176,8 @@ final class ReactorRules {
   static final class FluxCollectToImmutableList<T> {
     @BeforeTemplate
     Mono<List<T>> before(Flux<T> flux) {
-      return flux.collectList();
-    }
-
-    @BeforeTemplate
-    Mono<List<T>> before2(Flux<T> flux) {
-      return flux.collect(toImmutableList()).map(ImmutableList::copyOf);
+      return Refaster.anyOf(
+          flux.collectList(), flux.collect(toImmutableList()).map(ImmutableList::copyOf));
     }
 
     @AfterTemplate
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1170,13 +1170,18 @@ final class ReactorRules {
   }
 
   /**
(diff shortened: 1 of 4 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1166,6 +1167,23 @@ Flux<T> after(Flux<T> flux, Predicate<? super T> predicate, Comparator<? super T
     }
   }
 
+  /**
+   * Prefer {@link Flux#collect(Collector)} with {@link ImmutableList#toImmutableList} over {@link
+   * Flux#collectList()}.
+   */
+  static final class FluxCollectListToCollectImmutableList<T> {
```
mlrprananta, line 1174: Name might be a bit verbose, open to suggestions

mlrprananta: Maybe just `FluxCollectToImmutableList`?

rickie: Yes that is the perfect suggestions. We base the name on the content of the after template :). Will push the suggestion.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1166,6 +1167,23 @@ Flux<T> after(Flux<T> flux, Predicate<? super T> predicate, Comparator<? super T
...
+    @BeforeTemplate
+    Mono<List<T>> before(Flux<T> flux) {
```
rickie, line 1176: We can add: `flux.collect(toImmutableList()).map(ImmutableList::copyOf);`.

rickie (approved):
Added a commit with an extra rule :). Changes LGTM!

Slightly tweaked the suggested commit message.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1168,15 +1168,20 @@ Flux<T> after(Flux<T> flux, Predicate<? super T> predicate, Comparator<? super T
...
-  static final class FluxCollectListToCollectImmutableList<T> {
+  static final class FluxCollectToImmutableList<T> {
     @BeforeTemplate
     Mono<List<T>> before(Flux<T> flux) {
       return flux.collectList();
     }
 
+    @BeforeTemplate
+    Mono<List<T>> before2(Flux<T> flux) {
+      return flux.collect(toImmutableList()).map(ImmutableList::copyOf);
```
mlrprananta, line 1182: Can we use `Refaster.anyOf` in `before` instead?

rickie: Ah yeah, same return type so we definitely can 👍🏻.

mlrprananta: Will push

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1166,6 +1167,24 @@ Flux<T> after(Flux<T> flux, Predicate<? super T> predicate, Comparator<? super T
...
+    @BeforeTemplate
+    Mono<List<T>> before(Flux<T> flux) {
+      return Refaster.anyOf(
+          flux.collectList(), flux.collect(toImmutableList()).map(ImmutableList::copyOf));
+    }
+
+    @AfterTemplate
+    @UseImportPolicy(STATIC_IMPORT_ALWAYS)
+    Mono<List<T>> after(Flux<T> flux) {
+      return flux.collect(toImmutableList());
```
Stephan202: Let's make explicit that _strictly speaking_ this expression has a slightly different return type.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: Hmm, the `((Mono<ImmutableList<T>>) flux).map(ImmutableList::copyOf)` simplification should be in a rule of its own. (And is in fact a special case of a large range of rules.)

Stephan202 (approved):
Rebased and added a commit.

Suggested commit message:
```
Introduce `FluxCollectToImmutableList` Refaster rule (#570)

And extend `MonoIdentity` to simplify `mono.map(ImmutableList::copyOf)`
expressions where possible, as the new rule may introduce such cases.
```

rickie:
Changes LGTM! Will merge once 🟢.

PR #571, opened 2023-04-07
Introduce `FluxCollectToImmutableSet` Refaster rule

mlrprananta:
Idea is based on an internal GH thread.

Suggested commit message:
```
Introduce `FluxCollectToImmutableSet` Refaster rule (#571)
```

### DIFF
--- a/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
+++ b/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
@@ -378,7 +378,7 @@ final class ReactorRulesTest implements RefasterRuleCollectionTestCase {
   }
 
   Mono<ImmutableSet<Integer>> testFluxCollectToImmutableSet() {
-    return Flux.just(1, 4, 3, 2).collect(toImmutableList()).map(ImmutableSet::copyOf);
+    return Flux.just(1, 2).collect(toImmutableList()).map(ImmutableSet::copyOf);
   }
 
   ImmutableSet<Context> testContextEmpty() {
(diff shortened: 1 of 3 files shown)


rickie (approved):
Thanks for the PR @mlrprananta 🚀 !

In our OSS repo we try to avoid linking to "internal" things or repositories, so updated the description :).

Made small tweaks to the suggested commit message 😄.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -377,6 +377,10 @@ Flux<Integer> testFluxFilterSortWithComparator() {
     return Flux.just(1, 4, 3, 2).sort(reverseOrder()).filter(i -> i % 2 == 0);
   }
 
+  Mono<ImmutableSet<Integer>> testFluxCollectToImmutableSet() {
+    return Flux.just(1, 4, 3, 2).collect(toImmutableList()).map(ImmutableSet::copyOf);
```
rickie: On a second thought, let's simplify the numbers here. I think it makes sense to have `1, 2` as we are collecting and "usually" there will be multiple values to collect. The order used in the test above is important to the method that is used.

rickie (commented):
Added a commit to simplify a test 😬.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
mlrprananta: Should we also simplify for https://github.com/PicnicSupermarket/error-prone-support/pull/570?

rickie: Yes, will push a commit :).

Stephan202 (approved):
Rebased and added a tiny tweak.

rickie:
I hesitated with the "more" as not all were really "more contrived" :smile:.

PR #574, opened 2023-04-08
Introduce OpenSSF Scorecard GitHub action

Stephan202:
Suggested commit message:
```
Introduce OpenSSF Scorecard GitHub action (#574)

And resolve some of the issues it identified.

See https://securityscorecards.dev
```

For the output, see [this view](https://github.com/PicnicSupermarket/error-prone-support/security/code-scanning?query=is%3Aopen+pr%3A574). Only warnings about non-pinned dependencies remain. Those can be resolved using [Renovate](https://docs.renovatebot.com/modules/manager/github-actions/#additional-information), IIUC, so I propose we address that separately.

### DIFF
--- a/.github/workflows/openssf-scorecard.yml
+++ b/.github/workflows/openssf-scorecard.yml
@@ -18,7 +18,6 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v3.1.0
-        # XXX: Also apply elsewhere?
         with:
           persist-credentials: false
       - name: Run OpenSSF Scorecard analysis
--- a/.github/workflows/openssf-scorecard.yml
+++ b/.github/workflows/openssf-scorecard.yml
@@ -1,0 +1,32 @@
+name: OpenSSF Scorecard update
+on:
+  # XXX: Drop PR builds after testing.
+  pull_request:
+  push:
+    branches: [ master ]
+  schedule:
+    - cron: '0 4 * * 1'
+permissions:
+  contents: read
+  id-token: write
+  security-events: write
+jobs:
+  analyze:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3.1.0
+        # XXX: Also apply elsewhere?
+        with:
+          persist-credentials: false
+      - name: Run OpenSSF Scorecard analysis
+        uses: ossf/scorecard-action@v2.1.3
+        with:
+          results_file: results.sarif
+          results_format: sarif
+          publish_results: true
+      - name: Update GitHub's code scanning dashboard
+        uses: github/codeql-action/upload-sarif@v2.2.11
+        with:
+          sarif_file: results.sarif
+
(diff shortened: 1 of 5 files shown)


.github/workflows/pitest-update-pr.yml
```diff
@@ -9,16 +9,20 @@
...
   actions: read
-  checks: write
-  contents: read
-  pull-requests: write
 jobs:
   update-pr:
     if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    permissions:
+      actions: read
+      checks: write
```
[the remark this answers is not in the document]
Stephan202: This code is executed from `master`, thus safe IIUC.

.github/workflows/openssf-scorecard.yml
```diff
@@ -0,0 +1,32 @@
...
+  analyze:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3.1.0
+        # XXX: Also apply elsewhere?
+        with:
+          persist-credentials: false
+      - name: Run OpenSSF Scorecard analysis
+        uses: ossf/scorecard-action@v2.1.3
```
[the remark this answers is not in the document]
oxkitsune:
```suggestion
        uses: ossf/scorecard-action@80e868c13c90f172d68d1f4501dee99e2479f7af # v2.1.3
```
It's complaining about not using a pinned hash, any reason to not do this?
Imo if we keep a "human readable" comment with the actual version we should be good!
[Dependabot even has support](https://github.com/dependabot/dependabot-core/issues/4691) to update this comment!

.github/workflows/openssf-scorecard.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3.1.0
+        with:
+          persist-credentials: false
+      - name: Run OpenSSF Scorecard analysis
+        uses: ossf/scorecard-action@v2.1.3
+        with:
+          results_file: results.sarif
+          results_format: sarif
```
rickie, line 31: Cool stuff! This is what was also used by the EP PR for collecting diagnotics.

.github/workflows/openssf-scorecard.yml
```diff
@@ -0,0 +1,37 @@
...
+          publish_results: ${{ github.ref == 'refs/heads/master' }}
+      - name: Update GitHub's code scanning dashboard
+        uses: github/codeql-action/upload-sarif@v2.2.11
+        with:
+          sarif_file: results.sarif
+
```
rickie: One extra line here 👀.

rickie (approved):
This looks so interesting! Didn't dive too deep into all the checks this has, but it sounds promising.

Nice setup, could spot only 1 tiny thing, so added a commit.

.github/workflows/openssf-scorecard.yml
Stephan202: Nice eye for detail! :smile:

rickie:
Oh btw, Monday we'll do some big upgrades to Renovate. Let's see if it picks up on the GHA dependencies again. Otherwise, we'll look into that. Pending that we can setup the pinning that it now flags 😄. So to be continued.

.github/workflows/openssf-scorecard.yml
rickie: Consistency for the win 😉.

PR #576, opened 2023-04-10
Introduce `AssertThatMapContainsOnlyKeys` Refaster rule

mohamedsamehsalah:
Original suggestion [here](https://github.com/PicnicSupermarket/picnic-platform/pull/10073#discussion_r1155822141).

Suggested commit message:
```
Introduce `AssertThatMapContainsOnlyKeys` Refaster rule (#576)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
@@ -16,6 +16,7 @@ import com.google.errorprone.refaster.annotation.AfterTemplate;
 import com.google.errorprone.refaster.annotation.BeforeTemplate;
 import com.google.errorprone.refaster.annotation.UseImportPolicy;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
@@ -26,6 +27,7 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 import org.assertj.core.api.AbstractAssert;
 import org.assertj.core.api.AbstractBooleanAssert;
+import org.assertj.core.api.AbstractCollectionAssert;
 import org.assertj.core.api.AbstractMapAssert;
 import org.assertj.core.api.MapAssert;
 import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
```diff
@@ -220,6 +221,19 @@ MapAssert<K, V> after(Map<K, V> map, K key) {
...
+    @BeforeTemplate
+    AbstractAssert<?, ?> before(Map<K, V> map, Set<K> keys) {
+      return assertThat(map.keySet()).hasSameElementsAs(keys);
+    }
+
+    @AfterTemplate
+    @UseImportPolicy(STATIC_IMPORT_ALWAYS)
+    AbstractAssert<?, ?> after(Map<K, V> map, Set<K> keys) {
+      return assertThat(map).containsOnlyKeys(keys);
+    }
```
Stephan202: Let's change the return types to make clear that what we're doing here (like with many of the other rules) potentially breaks some code.

Stephan202 (approved):
Added a small commit and tweaked the suggested commit message. Nice:)

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/AssertJMapRulesTestInput.java
```diff
@@ -131,7 +131,7 @@ MapAssert<Integer, Integer> testAbstractMapAssertHasSameSizeAs() {
   }
 
   AbstractAssert<?, ?> testAssertThatMapContainsOnlyKeys() {
-    return assertThat(ImmutableMap.of(1, 2).keySet()).hasSameElementsAs(ImmutableSet.of(1));
+    return assertThat(ImmutableMap.of(1, 2).keySet()).hasSameElementsAs(ImmutableSet.of(3));
```
mohamedsamehsalah, line 134: Thanks 🚀 
I have seen this pattern everywhere but I cant fully understand the reasning behind this.
Why arent we using correct expected values ? 🤔

Stephan202: Good question. I wouldn't say we're fully consistent in this respect. But in new code I try to use distinct values, so that we properly validate that Refaster correctly "shuffles things around". (And since the code isn't executed anyway... :upside_down_face:.)

rickie (approved):
Changes LGTM! Nice, another contribution @mohamedsamehsalah :rocket:!

@werli do you want to take a look as well 😄?

werli (approved):

PR #592, opened 2023-04-23
Assign a proper build number for GitHub Actions builds

Stephan202:
Suggested commit message:
```
Assign a proper build number for GitHub Actions builds (#592)

This prevents the `Implementation-Version` field in
`META-INF/MANIFEST.MF` from claiming that the JARs were produced by a
local build.
```

How to test:
```sh
mvn clean install -DskipTests -Dverification.skip > /dev/null
unzip -c error-prone-contrib/target/error-prone-contrib-*-SNAPSHOT.jar META-INF/MANIFEST.MF | grep Implementation-Version:
GITHUB_RUN_ID=xxx mvn clean install -DskipTests -Dverification.skip > /dev/null
unzip -c error-prone-contrib/target/error-prone-contrib-*-SNAPSHOT.jar META-INF/MANIFEST.MF | grep Implementation-Version:
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,7 +53,7 @@ jobs:
       - name: Remove installed project artifacts
         run: mvn build-helper:remove-project-artifact
       - name: TEST
-        run: echo "Run ID: $GITHUB_RUN_ID"
+        run: 'echo "Run ID: $GITHUB_RUN_ID"'
 
 # XXX: Enable Codecov once we "go public".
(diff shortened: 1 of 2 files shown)


Stephan202:
Okay, so it seems that `GITHUB_RUN_ID` is not set unless explicitly exposed. We can do that, but it would mostly defeat the purpose of the Maven profile. And anyway this only really becomes relevant once we create release artifacts through GitHub Actions. I'll close this PR for now.

PR #599, opened 2023-04-26
Improve `AbstractMatcherTestChecker`

Stephan202:
I extracted these changes from [this branch](https://github.com/PicnicSupermarket/error-prone-support/compare/sschroevers/introduce-IsNullableMatcher), following the discussion in [this thread](https://github.com/PicnicSupermarket/error-prone-support/pull/527#discussion_r1173558090). There are no tests for this test code. My proposal would be to finalize #527 on top of this PR as proof that it works. (Over time we'll introduce other `Matcher`s that will rely on (and thus exercise) this code. IIRC I have one or more other branches that would benefit from these changes. I might try to open PRs for those.)

Suggested commit message:
```
Improve `AbstractMatcherTestChecker` (#599)

These changes enable testing of a wider range of `Matcher` implementations. In
a nutshell:
- The `Matcher` under test is now passed `VisitorState` instances with an
  accurate `TreePath`.
- The `Matcher` under test is no longer consulted for method select and cast
  type expressions, mirroring Refaster behavior.
```


rickie:
Rebased.

rickie (approved):
Nice improvement! 🚀

mlrprananta (approved):
LGTM!

PR #603, opened 2023-04-27
Require JDK 17 rather than JDK 11

Stephan202:
:exclamation: ~~This PR is on top of #604~~. :exclamation: This now targets `master`.

This PR is spun off from #198. The idea is that it is merged first; the scope of #198 is then reduced to text block support for `CompilationTestHelper` and `BugCheckerRefactoringTestHelper`. 

:exclamation: A lot of discussion in this PR was around use of [Jabel](https://github.com/bsideup/jabel). This turned out to cause flaky GitHub Actions builds, so eventually we decided to simply raise the project's baseline to JDK 17. This was the *now obsolete* suggested commit message:
```
Use Jabel to enable use of JDK >11 constructs while still targeting JDK 11 (#603)

The build now effectively requires JDK 17, though continued compatibility with
JDK 11 is validated by means of Surefire tests that use Maven Toolchains 
support to run JDK 11. 

Through Jabel, only a subset of new language features can be made compatible
with earlier versions of Java, so not all JDK 17 functionality can be used yet.
Similarly, APIs introduced after JDK 11 can also not be used yet.

Constructs that can now be used include text blocks, switch expressions and 
`instanceof` pattern matching. The code has been updated to make use of these
new constructs.

While there, make sure that the `sonar` Maven profile is picked up during
SonarCloud analysis.

See:
- https://github.com/bsideup/jabel
- https://maven.apache.org/guides/mini/guide-using-toolchains.html
```

*New* suggested  commit message:
```
Require JDK 17 rather than JDK 11 (#603)

By raising this baseline the project can now use Java 17 language features such
as text blocks, switch expressions and `instanceof` pattern matching. The code
has been updated to make use of these constructs.

Note that the project can still be used by builds that target an older version
of Java, as long as those builds are executed using JDK 17+.
```


Stephan202:
Open points (will circle back to this PR over the coming days):
1. A closer look at the JDK 20 compilation error; likely something due to Jabel.
2. Addressing the SonarCloud-reported issues.
3. Reviewing whether the Pitest-reported missing coverage is accurate/matches coverage on the target branch.

Stephan202:
I filed bsideup/jabel#177 for the JDK 20 build error. (We should also update the GitHub Actions workflow to use the latest JDK 20 release, but I'll open a separate PR for that.)

Stephan202:
> but I'll open a separate PR for that.

Filed #604.

Stephan202:
Rebased PR on top of #604.

Stephan202:
Updated the branch. The SonarCloud alerts have been resolved. I verified that all surviving mutations reported by Pitest are also present on the target branch; resolving those is out of scope.

The only open point is the Jabel JDK 20 incompatibility; let's wait for a few days to see whether we can get a new release for that. Alternatively we can temporarily disable the JDK 20 build.

Stephan202:
Rebase and resolved conflict.

rickie (approved):
Only had some minor tweaks. 

This PR is really cool 🚀 ! Epic setup @Stephan202 😄. It was a pleasure reviewing this.

Stephan202 (commented):
Rebased and added a commit with some new cleanup.

Stephan202:

Stephan202:
Rebased and added a commit. Also upgraded bsideup/jabel#177 to support JDK 21.

Stephan202:

Stephan202:
With #494 merged this PR now contains some not-completely-trivial conflicts. Before attempting to resolve those, I'll assess whether the impacted code is expected to change more anyway (likely yes), and if so prioritize those changes first.

Stephan202:

Stephan202:
Once again rebased and resolved conflicts, also modernized newly introduced code.

Stephan202:
Rebased and resolved conflict; the `pom.xml` setup is now slightly simpler.

Stephan202:
To decide: do we want to use [a fork](https://github.com/bsideup/jabel/issues/182#issuecomment-1898900307) in the interim? I rebased and pushed a commit with a proposal.

rickie (approved):
Nice that they produced a fork. Can't wait to use these constructs in EPS 😄.

Still really epic that you got this to work 🚀 !

pom.xml
```diff
@@ -935,9 +950,9 @@
                             <arg>10000</arg>
                         </compilerArgs>
                         <parameters>true</parameters>
-                        <source>${version.jdk}</source>
-                        <target>${version.jdk}</target>
                         <showWarnings>false</showWarnings>
+                        <source>${version.jdk.target}</source>
+                        <target>${version.jdk.target}</target>
```
werli, line 955: Is this intended? If so, can we add a comment explaining why source and target are both target? I have a hunch this is correct, yet it still looks super misleading. :eyes:

pom.xml
```diff
@@ -208,7 +208,10 @@
         <version.error-prone-orig>2.24.1</version.error-prone-orig>
         <version.error-prone-slf4j>0.1.22</version.error-prone-slf4j>
         <version.guava-beta-checker>1.0</version.guava-beta-checker>
-        <version.jdk>11</version.jdk>
+        <!-- Though the project targets JDK 11, the code base does use more
+        recent language constructs, thanks to the Jabel compiler plugin. -->
+        <version.jdk.source>17</version.jdk.source>
```
werli, line 213: I see https://github.com/bsideup/jabel/pull/177 is resolved. Why can't we have our source use Java 21? :eyes:

Stephan202: Ah, that ticket is about supporting JDK 20 and 21 for building, not for extending support for additional language features. Right now Jabel doesn't unlock support for any JDK 18+ language features, so there's no value in targeting e.g. Java 21.

---

I did have a quick look to see whether adding `PATTERN_SWITCH` [here](https://github.com/bsideup/jabel/blob/ad06a0be7d6f3bb41fdba149df21b4963036fe6c/jabel-javac-plugin/src/main/java/com/github/bsideup/jabel/JabelCompilerPlugin.java#L149-L165) would "just work", but the answer is "no".

Test code:
```diff
$ git stash show -p
diff --git a/example/src/main/java/com/example/JabelExample.java b/example/src/main/java/com/example/JabelExample.java
index 49fc126..d465164 100644
--- a/example/src/main/java/com/example/JabelExample.java
+++ b/example/src/main/java/com/example/JabelExample.java
@@ -23,6 +23,10 @@ public class JabelExample {
             case 2, 3 -> "two or three";
             default -> new JabelExample().new Inner().innerPublic();
         };
+        result = switch ((Object) args[0]) {
+            case String s -> s;
+            default -> "";
+        };
         return result;
     }
 
diff --git a/jabel-javac-plugin/src/main/java/com/github/bsideup/jabel/JabelCompilerPlugin.java b/jabel-javac-plugin/src/main/java/com/github/bsideup/jabel/JabelCompilerPlugin.java
index 0a5aeb1..aa0a15d 100644
--- a/jabel-javac-plugin/src/main/java/com/github/bsideup/jabel/JabelCompilerPlugin.java
+++ b/jabel-javac-plugin/src/main/java/com/github/bsideup/jabel/JabelCompilerPlugin.java
@@ -157,6 +157,7 @@ public class JabelCompilerPlugin implements Plugin {
                 case "EFFECTIVELY_FINAL_VARIABLES_IN_TRY_WITH_RESOURCES":
                 case "TEXT_BLOCKS":
                 case "PATTERN_MATCHING_IN_INSTANCEOF":
+                case "PATTERN_SWITCH":
                 case "REIFIABLE_TYPES_INSTANCEOF":
                 case "RECORDS":
                     //noinspection UnusedAssignment
```

Running `./gradlew test` this gives:
```
An exception has occurred in the compiler (21.0.1). Please file a bug against the Java compiler via the Java bug reporting page (https://bugreport.java.com) after checking the Bug Database (https://bugs.java.com) for duplicates. Include your program, the following diagnostic, and the parameters passed to the Java compiler in your report. Thank you.
com.sun.tools.javac.code.Symbol$CompletionFailure: class file for java.lang.runtime.SwitchBootstraps not found
```

The error is awkward, but I validated that if one targets JDK 21, then the `.class` file will contain a reference to [java.lang.runtime.SwitchBootstraps](https://github.com/openjdk/jdk/blob/437342b93e9e66340ac57bd1c6fdc948b3302db0/src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java), which means that indeed won't work with JDK 11 (as there that class isn't available).

pom.xml
Stephan202: It is intended indeed. `-source` must always be less than or equal to `-target`. `-target` specifies the minimum Java version targeted; that's still 11 as before. Will add a commit.

Stephan202 (commented):
Added a commit. Tnx for the reviews!

Stephan202:
One thing to keep an eye on: [the JDK 21 build failed once](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/7587784617/job/20669216392?pr=603), with an error indicating that Jabel didn't kick in (or didn't kick in fast enough).

I don't think this is necessarily a blocker, but it _is_ a bit of a risk. I might push a few empty commits to get a feel for how often this happens.

Stephan202:
Now [a different failure](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/7588287595/job/20670477806?pr=603): `ClassNotFoundException: com.sun.jna.Library`. :thinking:

Stephan202:
K, [same failure as before](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/7588649210/job/20671640858?pr=603): `pattern matching in instanceof is not supported in -source 11`. The error is preceded by a thread dump that shows two builds threads (including for the one for the `documentation-support` module, for which compilation fails) are blocked in `BasicJavacTask.initPlugins`, waiting for ByteBuddy to be installed. This looks like a classic race condition; requires more investigation.

Stephan202:
I'm trying to reproduce the issue locally; no luck so far

Stephan202:
Ran the build locally many times in a loop with JDK 17 and 21; never failed. Inspecting the printed stack traces, it's also not clear to me how/when the source code can be inspected before `BasicJavacTask#initPlugins` completes.

Stephan202:
I've now been doing a bunch more builds, and the incidence of build failures went down, to about 1 in 10 runs. There were [a](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/7602273825/job/20702626801?pr=603) [few](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/7602927513/job/20704044604?pr=603) more with `java.lang.ClassNotFoundException: com.sun.jna.Library`, but none of the other variaty. Just like the previous one of this kind, it impacts builds that use cached Maven artifacts. Not sure what to make of this, but it seems unlikely to be a Jabel-specific bug.

Anyway, we now have a decision to make. We could:
1. Not merge this PR. This would also block #198, which is a shame IMHO.
2. Merge the PR as-is. This will make our build more flaky, which may very well get annoying.
3. Simplify the PR by getting rid of Jabel and simply raising our baseline to JDK 17. This may impact adoption of the project by others, but I'm not sure how much of an issue that is<sup>&dagger;</sup>.

I'm tending to option (3). Note that this would not make the project incompatible with builds that target JDK 11, it would just require those builds to be executed with a more recent JDK. That doesn't sound too unreasonable to me, especially considering that Error Prone Support users presumably care about a modern build setup.

Interested to hear your thoughts @rickie @werli.

<sup>&dagger;</sup>: For the OpenRewrite integration it may be nice to keep JDK 11 support. But if we care enough about that, we could look into moving the Refaster rules to a separate JDK 11-targeting module.

Stephan202:
Discussed offline: we'll go for option 3: raise the baseline to JDK 17. Supporting text blocks is a major quality-of-life improvement for us as maintainers, and will make for a better workshop experience. There's also precedent in the industry, with e.g. new Spring and jOOQ releases depending on Java 17. We will document this new constraint in the README, and point out that while the project will require JDK 17, it can still be used in builds that target older versions of Java.

But before we do this, we'll release version 0.15.0 without this PR, so that there will be one more JDK 11-compatible release.

Stephan202:
I rebased, dropped all the empty test commits, and added a commit to raise the baseline to JDK 17. This entailed just reverting code (plus bumping the relevant property). I split one change off to #986. I reverted the README changes because [IDEA-288052](https://youtrack.jetbrains.com/issue/IDEA-288052) is still unresolved, and in new tests the `idea` profile did not work around this. (Which makes sense, actually.)

werli (approved):
I agree with going for option 3 by first releasing 0.15.0 and then raising the baseline. 👍 

> We will document this new constraint in the README, and point out that while the project will require JDK 17, it can still be used in builds that target older versions of Java

Is this still planned for this?

Stephan202:
@werli yes, I had a draft comment about that open in my browser :wink:

---

Updated the suggested commit message. Open points:
1. Document the build vs. target difference in the README.
2. Apply any other small changes that are unlocked by the JDK 17 baseline.

(Will do ~later today.)

Stephan202:
Alright, added one more commit. More improvements are possible, but those relate to functional improvements (new Refaster rules), and so belong to a separate PR. Will file one on top of this branch.

oxkitsune (approved):

refaster-test-support/src/main/java/tech/picnic/errorprone/refaster/test/RefasterRuleCollection.java
```diff
@@ -276,9 +276,10 @@ private void reportUnexpectedMatches(
             unexpectedMatchesByLineNumber.entries().stream()
                 .map(
                     e ->
-                        String.format(
-                            "Rule `%s` matches on line %s, while it should match in a method named `test%s`.",
-                            e.getValue(), e.getKey(), e.getValue()))
+                        """
+                        Rule `%s` matches on line %s, while it should match in a method named \
+                        `test%s`."""
+                            .formatted(e.getValue(), e.getKey(), e.getValue()))
```
Stephan202, line 282: We don't use `.formatted` elsewhere, but i.c.w. text blocks it makes sense. Let's review whether we want to switch all `String.format` usages.

Stephan202 (commented):
I rebased and added two more small commits, introducing a few more text blocks (and thus resolving some more SonarCloud violations).

I'll merge once built, so that I can start rebasing other PRs. (@rickie I realize that this skips a review opportunity; happy to open follow-up PRs.)

Stephan202:
In the context of this PR I looked into Maven Toolchains support. That code was eventually not merged, but part of that logic is now used by #1686.

PR #605, opened 2023-04-29
Extend `StreamRules` Refaster rule collection

Stephan202:
Suggested commit message:
```
Extend `StreamRules` Refaster rule collection (#605)

All changes suggested by SonarCloud's `java:s4034` rule, as well as the
examples mentioned in openrewrite/rewrite#2984 are now covered. (In a
number of cases through composition of more generic rules.)

See https://rules.sonarsource.com/java/RSPEC-4034
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
@@ -246,6 +246,7 @@ final class StreamRules {
   }
 
   /** In order to test whether a stream has any element, simply try to find one. */
+  // XXX: This this rule assumes that any matched `Collector` does not perform any filtering.
   static final class StreamIsEmpty<T> {
     @BeforeTemplate
     boolean before(Stream<T> stream, Collector<? super T, ?, ? extends Collection<?>> collector) {
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
@@ -246,7 +246,7 @@ final class StreamRules {
   }
 
   /** In order to test whether a stream has any element, simply try to find one. */
-  // XXX: This this rule assumes that any matched `Collector` does not perform any filtering.
+  // XXX: This rule assumes that any matched `Collector` does not perform any filtering.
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -247,12 +248,13 @@ Optional<S> after(Stream<T> stream, Function<? super T, S> function) {
   /** In order to test whether a stream has any element, simply try to find one. */
   static final class StreamIsEmpty<T> {
     @BeforeTemplate
-    boolean before(Stream<T> stream) {
+    boolean before(Stream<T> stream, Collector<? super T, ?, ? extends Collection<?>> collector) {
```
rickie, line 251: Isn't it nicer to have a separate before template for this one?

rickie (approved):
Changes LGTM!

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
Stephan202: No strong opinion. (Something we could enforce one way or another with automation...)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -247,12 +248,13 @@ Optional<S> after(Stream<T> stream, Function<? super T, S> function) {
...
       return Refaster.anyOf(
           stream.count() == 0,
           stream.count() <= 0,
           stream.count() < 1,
-          stream.findFirst().isEmpty());
+          stream.findFirst().isEmpty(),
+          stream.collect(collector).isEmpty());
```
Venorcis, line 257: couldn't you theoretically use a filtering collector though?

Stephan202: Yes! We should add a comment calling out this caveat. I'll add a commit later this morning :+1:

Stephan202: Added two more comments. Since this kind of is rare (and since we rewrite top-level `filtering` collectors to a regular `filter` step), I think this is an acceptable limitation for now.

Venorcis (approved):

Stephan202:
:+1:

PR #606, opened 2023-05-01
Fix Gradle installation guide link in README

Stephan202:
Suggested commit message:
```
Fix Gradle installation guide link in README (#606)
```

Without this the rendering is [incorrect](https://github.com/PicnicSupermarket/error-prone-support#gradle). Wrapped after 79 chars as usual.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,8 +109,9 @@ definition. -->
 
 #### Gradle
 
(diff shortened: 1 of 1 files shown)


ghokun (approved):
😅

rickie (approved):

PR #608, opened 2023-05-01
Extend `PrimitiveRules` Refaster rule collection

Stephan202:
Suggested commit message:
```
Extend `PrimitiveRules` Refaster rule collection (#608)

Assorted methods and constants exposed by Guava's
`com.google.common.primitives.*` types are now replaced with their JDK
equivalents.

While there, also update the parameter types of some existing `@AfterTemplate`
methods, resolving an inconsistency flagged by @knutwannheden.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PrimitiveRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PrimitiveRules.java
@@ -1,6 +1,13 @@
 package tech.picnic.errorprone.refasterrules;
 
+import com.google.common.primitives.Booleans;
+import com.google.common.primitives.Bytes;
+import com.google.common.primitives.Chars;
+import com.google.common.primitives.Doubles;
+import com.google.common.primitives.Floats;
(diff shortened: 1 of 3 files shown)


rickie (approved):
Added a commit with a think I spotted.

Nice improvements, funny that they have literally the same methods there.

Stephan202:
Tweak LGTM!

Venorcis (approved):
LGTM 👍🏻 

Could consider looking at all the `to/fromByteBuffer` functions in the future as well, as they state themselves `Arguably, it's preferable to use ByteBuffer`, but that's less straightforward 😄

Stephan202:
Ah, nice catch w.r.t. [this](https://guava.dev/releases/23.0/api/docs/com/google/common/primitives/Longs.html#fromByteArray-byte:A-) indeed. I looked only for explicit references to "Java" and "JDK". One thing that I'm not so sure about in this case, is that for the [dual](https://guava.dev/releases/23.0/api/docs/com/google/common/primitives/Longs.html#toByteArray-long-) there is no "acceptable" alternative, which might make the rewrite slightly awkward :thinking:.

PR #609, opened 2023-05-01
Extend `AssertJThrowingCallableRules` Refaster rule collection

rickie:
While applying the Refaster rules internally I found another case that we are missing. This was intended as this list is not meant to be exhaustive. However, I feel it makes sense to be consistent and for that reason I added some of the missing cases for the `assertThatNullPointerException`, `assertThatIOException`, and `assertThatExceptionOfType`. 

Now we have a `StartsWith`, `MessageContaining`, and `MessageNotContaining` for the ones that already had `WithParams` :smile:.

Suggested commit message:
```
Extend `AssertJThrowingCallableRules` Refaster rule collection (#609)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRules.java
@@ -319,6 +319,44 @@ final class AssertJThrowingCallableRules {
     }
   }
 
+  static final class AssertThatThrownByNullPointerExceptionHasMessageContaining {
+    @BeforeTemplate
+    @SuppressWarnings(
+        "AssertThatThrownByNullPointerException" /* This is a more specific template. */)
+    AbstractObjectAssert<?, ?> before(ThrowingCallable throwingCallable, String message) {
+      return assertThatNullPointerException()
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRules.java
```diff
@@ -319,6 +319,44 @@ static final class AssertThatThrownByNullPointerExceptionHasMessageStartingWith
...
   }
 
+  static final class AssertThatThrownByNullPointerExceptionHasMessageContaining {
+    @BeforeTemplate
+    @SuppressWarnings(
+        "AssertThatThrownByNullPointerException" /* This is a more specific template. */)
+    AbstractObjectAssert<?, ?> before(ThrowingCallable throwingCallable, String message) {
+      return assertThatNullPointerException()
+          .isThrownBy(throwingCallable)
+          .withMessageContaining(message);
```
Stephan202, line 329: There are also format string overloads :grimacing:. (But at we're fast reaching the point where a bug checker becomes lighter.)

Stephan202 (approved):
Rebased; LGTM!

Venorcis (approved):
LGTM 😄

PR #617, opened 2023-05-07
Introduce `VerifyOnlyElementInFlux` Refaster rule

giovannizotta:
When verifying a `Flux` only has one element, avoid collecting to list.

Notes:

1. The `@BeforeTemplate` is quite big, can it be a problem? I don't know if/how it can be reduced in size;
2. I am pretty sure the name of the rule is not the best, please suggest one that follows existing standards if you have better ideas;
3. It could be extended to `toImmutableSet` as well (see comment)
4. I used `flux.collect(toImmutableList())` instead of `flux.collectList()` because of the rule `FluxCollectToImmutableList`; 
5. How does it work if there is a match for this rule, but with `flux.collectList()`? Will error-prone perform "two passes" and first move `flux.collectList()` to `flux.collect(toImmutableList())` and then apply my rule?

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1332,20 +1332,13 @@ final class ReactorRules {
     }
   }
 
-  /** Avoid collecting when verifying only the given element is in the given {@link Flux}. */
+  /** Avoid value collection when verifying that a {@link Flux} emits exactly one value. */
   static final class FluxAsStepVerifierExpectNext<T> {
     @BeforeTemplate
     StepVerifier.Step<ImmutableList<T>> before(Flux<T> flux, T object) {
       return flux.collect(toImmutableList())
           .as(StepVerifier::create)
           .assertNext(list -> assertThat(list).containsExactly(object));
-    }
-
-    @BeforeTemplate
-    StepVerifier.Step<ImmutableSet<T>> before2(Flux<T> flux, T object) {
-      return flux.collect(toImmutableSet())
-          .as(StepVerifier::create)
-          .assertNext(set -> assertThat(set).containsExactly(object));
     }
 
     @AfterTemplate
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1434,4 +1434,20 @@ final class ReactorRules {
       return step.verifyTimeout(duration);
     }
   }
+
+  /** Avoid collecting when verifying only the given element is in the given {@link Flux}. */
+  static final class VerifyOnlyElementInFlux<T> {
+    @BeforeTemplate
+    Duration before(Flux<T> flux, T object) {
+      return flux.collect(toImmutableList())
+          .as(StepVerifier::create)
+          .assertNext(list -> assertThat(list).containsExactly(object))
+          .verifyComplete();
+    }
+
+    @AfterTemplate
+    Duration after(Flux<T> flux, T object) {
+      return flux.as(StepVerifier::create).expectNext(object).verifyComplete();
+    }
+  }
 }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1332,11 +1332,13 @@ final class ReactorRules {
     }
   }
 
-  /** Avoid value collection when verifying that a {@link Flux} emits exactly one value. */
-  static final class FluxAsStepVerifierExpectNext<T> {
+  /** Avoid list collection when verifying that a {@link Flux} emits exactly one value. */
+  // XXX: This rule assumes that the matched collector does not drop elements. Consider introducing
+  // a `@Matches(DoesNotDropElements.class)` or `@NotMatches(MayDropElements.class)` guard.
+  static final class FluxAsStepVerifierExpectNext<T, L extends List<T>> {
     @BeforeTemplate
-    StepVerifier.Step<ImmutableList<T>> before(Flux<T> flux, T object) {
-      return flux.collect(toImmutableList())
+    StepVerifier.Step<L> before(Flux<T> flux, Collector<? super T, ?, L> listCollector, T object) {
+      return flux.collect(listCollector)
           .as(StepVerifier::create)
           .assertNext(list -> assertThat(list).containsExactly(object));
     }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1332,6 +1332,28 @@ final class ReactorRules {
     }
   }
 
+  /** Avoid collecting when verifying only the given element is in the given {@link Flux}. */
+  static final class FluxAsStepVerifierExpectNext<T> {
+    @BeforeTemplate
+    StepVerifier.Step<ImmutableList<T>> before(Flux<T> flux, T object) {
+      return flux.collect(toImmutableList())
+          .as(StepVerifier::create)
+          .assertNext(list -> assertThat(list).containsExactly(object));
+    }
+
+    @BeforeTemplate
+    StepVerifier.Step<ImmutableSet<T>> before2(Flux<T> flux, T object) {
+      return flux.collect(toImmutableSet())
+          .as(StepVerifier::create)
+          .assertNext(set -> assertThat(set).containsExactly(object));
+    }
+
+    @AfterTemplate
+    StepVerifier.Step<T> after(Flux<T> flux, T object) {
+      return flux.as(StepVerifier::create).expectNext(object);
+    }
+  }
+
   /** Prefer {@link StepVerifier.LastStep#verifyComplete()} over more verbose alternatives. */
   static final class StepVerifierLastStepVerifyComplete {
     @BeforeTemplate
@@ -1432,22 +1454,6 @@ final class ReactorRules {
     @AfterTemplate
     Duration after(StepVerifier.LastStep step, Duration duration) {
       return step.verifyTimeout(duration);
-    }
-  }
-
-  /** Avoid collecting when verifying only the given element is in the given {@link Flux}. */
-  static final class VerifyOnlyElementInFlux<T> {
-    @BeforeTemplate
-    Duration before(Flux<T> flux, T object) {
-      return flux.collect(toImmutableList())
-          .as(StepVerifier::create)
-          .assertNext(list -> assertThat(list).containsExactly(object))
-          .verifyComplete();
-    }
-
-    @AfterTemplate
-    Duration after(Flux<T> flux, T object) {
-      return flux.as(StepVerifier::create).expectNext(object).verifyComplete();
     }
   }
 }
--- a/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
+++ b/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
@@ -1,7 +1,6 @@
 package tech.picnic.errorprone.refasterrules;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
-import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.collect.MoreCollectors.toOptional;
 import static java.util.Comparator.reverseOrder;
 import static java.util.function.Function.identity;
@@ -423,18 +422,12 @@ final class ReactorRulesTest implements RefasterRuleCollectionTestCase {
         StepVerifier.create(Mono.just("baz")).expectNextMatches("qux"::equals));
   }
 
-  ImmutableSet<Duration> testFluxAsStepVerifierExpectNext() {
-    return ImmutableSet.of(
-        Flux.just(1)
-            .collect(toImmutableList())
-            .as(StepVerifier::create)
-            .assertNext(list -> assertThat(list).containsExactly(2))
-            .verifyComplete(),
-        Flux.just(3)
-            .collect(toImmutableSet())
-            .as(StepVerifier::create)
-            .assertNext(set -> assertThat(set).containsExactly(4))
-            .verifyComplete());
+  Duration testFluxAsStepVerifierExpectNext() {
+    return Flux.just(1)
+        .collect(toImmutableList())
+        .as(StepVerifier::create)
+        .assertNext(list -> assertThat(list).containsExactly(2))
+        .verifyComplete();
   }
 
   Duration testStepVerifierLastStepVerifyComplete() {
--- a/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
+++ b/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
@@ -1,6 +1,7 @@
 package tech.picnic.errorprone.refasterrules;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.collect.MoreCollectors.toOptional;
 import static java.util.Comparator.reverseOrder;
 import static java.util.function.Function.identity;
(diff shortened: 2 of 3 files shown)


Stephan202 (commented):
Nice draft, in the small hours of the day ;)

giovannizotta:
> Nice draft, in the small hours of the day ;)

The best ones, right? 😄

giovannizotta:
The first version of the `@BeforeTemplate` worked well:
```java
@BeforeTemplate
Duration before(Flux<T> flux, T object) {
  return flux.collect(toImmutableList())
      .as(StepVerifier::create)
      .assertNext(collection -> assertThat(collection).containsExactly(object))
      .verifyComplete();
}
```

Alongside `flux.collect(toImmutableList())`, I wanted to add support for `flux.collect(toImmutableSet())`, so I wrote the following:
```java
@BeforeTemplate
Duration before(Flux<T> flux, T object) {
  return Refaster.anyOf(flux.collect(toImmutableList()), flux.collect(toImmutableSet()))
      .as(StepVerifier::create)
      .assertNext(collection -> assertThat(collection).containsExactly(object))
      .verifyComplete();
}
```

I am not sure how `Refaster#anyOf` works, but the following is very weird. Let's say I have the following test:
```java
ImmutableSet<Duration> testVerifyOnlyElementInFlux() {
  return ImmutableSet.of(
      Flux.just(1)
          .collect(toImmutableList())
          .as(StepVerifier::create)
          .assertNext(list -> assertThat(list).containsExactly(1))
          .verifyComplete(),
      Flux.just(2)
          .collect(toImmutableSet())
          .as(StepVerifier::create)
          .assertNext(set -> assertThat(set).containsExactly(2))
          .verifyComplete());
}
```
I would expect the result to be:
```java
ImmutableSet<Duration> testVerifyOnlyElementInFlux() {
  return ImmutableSet.of(
      Flux.just(1).as(StepVerifier::create).expectNext(1).verifyComplete(),
      Flux.just(2).as(StepVerifier::create).expectNext(2).verifyComplete());
}
```
The fun part? Now the rule does not match `flux.collect(toImmutablelist())`, but instead it matches `flux.collect(toImmutableSet())`:
```diff
diff (-expected +actual):
    @@ -438,7 +438,11 @@
     
       ImmutableSet<Duration> testVerifyOnlyElementInFlux() {
         return ImmutableSet.of(
    -        Flux.just(1).as(StepVerifier::create).expectNext(1).verifyComplete(),
    +        Flux.just(1)
    +            .collect(toImmutableList())
    +            .as(StepVerifier::create)
    +            .assertNext(list -> assertThat(list).containsExactly(1))
    +            .verifyComplete(),
             Flux.just(2).as(StepVerifier::create).expectNext(2).verifyComplete());
       }
     }
```

Am I missing something? What kind of sorcery is this?

P.S. Now that I think of it, I can also extend it to varargs in the case of `toImmutableSet`

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestOutput.java
```diff
@@ -435,4 +435,8 @@ Duration testStepVerifierLastStepVerifyErrorMessage() {
   Duration testStepVerifierLastStepVerifyTimeout() {
     return StepVerifier.create(Mono.empty()).verifyTimeout(Duration.ZERO);
   }
+
+  Duration testVerifyOnlyElementInFlux() {
+    return Flux.just(1).as(StepVerifier::create).expectNext(1).verifyComplete();
```
rickie, line 440: Let's use distinct values in the tests :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1397,4 +1397,20 @@ Duration after(StepVerifier.LastStep step, Duration duration) {
...
+    Duration before(Flux<T> flux, T object) {
+      return flux.collect(toImmutableList())
+          .as(StepVerifier::create)
+          .assertNext(list -> assertThat(list).containsExactly(object))
+          .verifyComplete();
+    }
+
+    @AfterTemplate
+    Duration after(Flux<T> flux, T object) {
+      return flux.as(StepVerifier::create).expectNext(object).verifyComplete();
```
rickie, line 1413:
```suggestion
      return flux.as(StepVerifier::create).expectNext(object);
```
Besides the fact that the return types are now slightly different we can drop the `verifyComplete` part. 
Assuming the tests pass, we can do this rewrite as we test that the list literally contains one object. Strictly speaking this is not a behavior preserving change because of the updated return type. The `verifyComplete` should be able to handle it though.

As a result we should also move this up as it belongs with the other `StepVerifier.Step` return type rules.

rickie (commented):
> - The @BeforeTemplate is quite big, can it be a problem? I don't know if/how it can be reduced in size;

I proposed a change that makes it one line smaller. In general, this is not really a problem as we are matching a bigger statement in this case. 

> - I am pretty sure the name of the rule is not the best, please suggest one that follows existing standards if you have better ideas;

I applied our future naming algorithm which is based on the content of the after template.

> - It could be extended to toImmutableSet as well (see comment)
I used flux.collect(toImmutableList()) instead of flux.collectList() because of the rule FluxCollectToImmutableList;

Yes that is a good idea. I pushed some changes to make it work. I also tried it with `Refaster#anyOf` but it didn't really work that well 🤔. With the more specific before template it seems better to have two different before templates because of the `Immutable{List,Set}<T>` in the return types.

> How does it work if there is a match for this rule, but with flux.collectList()? Will error-prone perform "two passes" and first move flux.collectList() to flux.collect(toImmutableList()) and then apply my rule?

Yes Error Prone will fix it in two runs. That's why our internal `patch.sh` script runs again when changes are introduced by Error Prone.

Added a commit with some changes 😄, thanks for opening the PR 🚀 !

rickie:
> Am I missing something? Am I using Refaster#anyOf in a wrong way? What kind of sorcery is this?

Hmm, I tried looking into this and combining them but had the same issue.

As before template I had the following:
```
    @BeforeTemplate
    StepVerifier.Step<? extends ImmutableCollection<T>> before(Flux<T> flux, T object) {
      return Refaster.anyOf(flux.collect(toImmutableList()), flux.collect(toImmutableSet()))
          .as(StepVerifier::create)
          .assertNext(list -> assertThat(list).containsExactly(object));
    }
```
Had the same problem, requires some more brain cycles at a different point in time :). Now have to focus on something else.

giovannizotta:
Hey @rickie thanks for the review! I won't have much time in the next days, but I'll pick this up.

Just one thing about making it one line smaller: in my understanding, if we don't check for `verifyComplete()` we could have a situation where the input does not do `verifyComplete()`, and the output would give a warning `Unfinished step verifier`. 

I'll come back to this asap, probably in the weekend 😄 

Screenshot 2023-05-11 at 08 20 17

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
giovannizotta: Should've replied to this comment :) https://github.com/PicnicSupermarket/error-prone-support/pull/617#issuecomment-1543404265

Stephan202 (commented):
Has a quick peek; nice rule!

> Just one thing about making it one line smaller: in my understanding, if we don't check for `verifyComplete()` we could have a situation where the input does not do `verifyComplete()`, and the output would give a warning `Unfinished step verifier`. 

As-is the rule will match both with and without `verifyComplete()`, meaning it's more generic; a property we generally like. It's okay if a rule also matches code that is "otherwise wrong" :smile:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1295,6 +1295,28 @@ StepVerifier.Step<T> after(StepVerifier.Step<T> step, T object) {
...
+      return flux.collect(toImmutableList())
+          .as(StepVerifier::create)
+          .assertNext(list -> assertThat(list).containsExactly(object));
+    }
+
+    @BeforeTemplate
+    StepVerifier.Step<ImmutableSet<T>> before2(Flux<T> flux, T object) {
+      return flux.collect(toImmutableSet())
+          .as(StepVerifier::create)
+          .assertNext(set -> assertThat(set).containsExactly(object));
```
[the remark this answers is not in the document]
rickie: Hmm, that would indeed be the best option. Might find some time for that later this week, but not sure.

giovannizotta:
> As-is the rule will match both with and without `verifyComplete()`

Isn't this a problem? Let's say the original code is:
```java
Flux.just(1)
    .collect(toImmutableList())
    .as(StepVerifier::create)
    .assertNext(list -> assertThat(list).containsExactly(1));
```
This code is correct and does not give warnings. We are sure the flux is completed because we collected to list and then verified the only element in the list was the expected one.
In the new code:
```java
Flux.just(1).as(StepVerifier::create).expectNext(1);
```
We are not verifying if the flux is completed, which gives the warning `Unfinished StepVerifier`. So we are turning code that compiles into code that gives a warning, that does not seem right to me?

Stephan202:
> This code is correct and does not give warnings. We are sure the flux is completed because we collected to list and then verified the only element in the list was the expected one.

That code might not yield a warning, but that's just a limitation of IDEA; it is just as incorrect. _Neither_ variant verifies anything :grimacing:.

The following check should fail, but doesn't (because there's no verification):
```java
Flux.just(2)
    .collect(toImmutableList())
    .as(StepVerifier::create)
    .assertNext(list -> assertThat(list).containsExactly(1));
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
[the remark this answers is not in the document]
rickie: TBH, I think fixing this issue requires creating a Refaster rule in Error Prone and debugging why this specific `Refaster#anyOf` usage doesn't work. That sounds like a bit more work. I'd suggest to drop the `toImmutableSet()` and create an XXX and pick up that debugging later. WDYT? @Stephan202

giovannizotta:
I am sorry for not coming back at this for more than a month, it has been really busy lately 😄 I hope there's no rush.

To be honest, I still don't understand. Take the following snippet:
```java
@BeforeTemplate
StepVerifier.Step<ImmutableList<T>> before(Flux<T> flux, T object) {
  return flux.collect(toImmutableList())
      .as(StepVerifier::create)
      .assertNext(list -> assertThat(list).containsExactly(object));
}

@AfterTemplate
StepVerifier.Step<T> after(Flux<T> flux, T object) {
  return flux.as(StepVerifier::create).expectNext(object);
}
```
The code in the `@BeforeTemplate`  collects the flux in a `Mono<ImmutableList<T>>` and checks that the list contains _one and only one element_, and checks if it is specifically the one we're looking for. 
The code in the `@AfterTemplate` is checking if _the first element_ of the flux is the one we're looking for. However, it does not have any check on the length of the flux.
To me without the `verifyComplete()` on the `@AfterTemplate`, these 2 templates are testing different things and therefore not a great fit for error-prone, unless I am missing something (which is likely 😄).
Am I missing something? Please let me know. No rush though, and again, sorry for the delay 😅

mohamedsamehsalah:
> @giovannizotta Am I missing something? Please let me know. No rush though, and again, sorry for the delay 😅

Hello, just passing through 👀 

I think the confusion is that 
```
  return flux.collect(toImmutableList())
      .as(StepVerifier::create)
      .assertNext(list -> assertThat(list).containsExactly(object));
```
is _also_ checking for one and only one element (Since we only have one `assertNext` call)

As far as I can tell, the @AfterTemplate is the less verbose to check for this.

giovannizotta:
Hey @mohamedsamehsalah thanks for passing by!
The issue I see is the opposite: 

- the `@BeforeTemplate` is checking for _one and only one_ element
- the `@AfterTemplate` is checking for _at least one_ element

rickie:
Thanks for thoroughly thinking about this and explaining your thought process, I like it :smile:!

One important assumption we are making here, which should be true for 99% of the cases, is that someone is using something like `verifyComplete()` after the `assertNext`. As Stephan mentions [here](https://github.com/PicnicSupermarket/error-prone-support/pull/617#issuecomment-1555956360), while the code doesn't yield a warning, it doesn't verify anything.

Having the `@BeforeTemplate` without `.verifyComplete()` makes it more generic, which should also be fine in most cases. If it appears that this is not the case, we can always decide to make the rule more strict. For example, by creating a `Matcher` ([examples](https://github.com/PicnicSupermarket/error-prone-support/tree/master/refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers)) that verifies that the `assertNext` is followed by a `verifyComplete`. 
Another option what would maybe even be better is creating *another* check (probably a BugPattern would be better for this) that _enforces_ that an `assertNext` is followed by an `verifyComplete`. That would even be better and make this Refaster rule 100% valid :wink:. (We can probably make that rule more generic even).

> **these 2 templates are testing different things**

So together with the knowledge that an `assertNext` is not verifying anything, and therefore the assumption that we have an `verifyComplete()`, we _are_ actually doing a correct replacement. 

Let's consider the following failing test:
```
 Flux.just(1, 2)
         .collect(toImmutableList())
         .as(StepVerifier::create)
         .assertNext(list -> assertThat(list).containsExactly(1))
         .verifyComplete();
```

Would correctly be rewritten to: 
```
   Flux.just(1, 2)
        .as(StepVerifier::create)
        .expectNext(1)
        .verifyComplete();
```

Which would fail with: 
> java.lang.AssertionError: expectation "expectComplete" failed (expected: onComplete(); actual: onNext(2))

[Here](https://github.com/search?q=org%3APicnicSupermarket+%2F%5C.assertNext.*%5C%29%3B%2F&type=code) a quick (not fullproof) GitHub search on Picnic code that shows we are in all cases verifying.

If I'm missing anything in my explanation, please let me know 😄, happy to elaborate further!

giovannizotta:
@rickie thanks a lot for the amazing explanation! Also, for the offline discussion 😄 
Now it's very clear: I was confused on how the step verifier works, as I am really a reactive noobie. But the explanation makes a lot of sense 😄.

To finish up the PR, I decided to move on with the suggestion to remove the `toImmutableSet()` rule because of the different semantics. Also, at some point I would like to look into building the matcher you mentioned, but let's keep that for the future 🙂. 

Just a question about the test framework. I tried minimising the test by removing the `verifyComplete()`, but that leads to having a different return type for the function in the before and after template.
The test framework does not substitute the return type, so if we give the proper return type to the `After` template we end up with a mismatch:
```
-  StepVerifier.Step<Integer> testFluxAsStepVerifierExpectNext() {
+  StepVerifier.Step<ImmutableList<Integer>> testFluxAsStepVerifierExpectNext() {
```

And if we give `StepVerifier.Step<ImmutableList<Integer>>` as a return type to the `@After`:
```
incompatible types: reactor.test.StepVerifier.Step<java.lang.Integer> cannot be converted to reactor.test.StepVerifier.Step<com.google.common.collect.ImmutableList<java.lang.Integer>>
```

So I decided to just stick to also having a `verifyComplete()` in the test to have a `Duration` as a return type in both cases.

rickie (approved):
Changes LGTM. Let's indeed not consider the `toImmutableSet` for now.

rickie:
@giovannizotta Your reasoning about the tests is correct, we can use the `verifyComplete` there :).

Stephan202 (approved):
Nice discussions!

```
Introduce `FluxAsStepVerifierExpectNext` Refaster rule (#617)
```

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -422,6 +422,18 @@ ImmutableSet<StepVerifier.Step<String>> testStepVerifierStepExpectNext() {
...
+    return ImmutableSet.of(
+        Flux.just(1)
+            .collect(toImmutableList())
+            .as(StepVerifier::create)
+            .assertNext(list -> assertThat(list).containsExactly(2)),
+        Flux.just(3)
+            .collect(toCollection(ArrayList::new))
+            .as(StepVerifier::create)
+            .assertNext(list -> assertThat(list).containsExactly(4)));
+  }
```
Stephan202, line 435: After pushing this I realized that perhaps for rule one test case would suffice, or at least be more consistent with what we do elsewhere. :shrug:

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -422,6 +422,14 @@ ImmutableSet<StepVerifier.Step<String>> testStepVerifierStepExpectNext() {
         StepVerifier.create(Mono.just("baz")).expectNextMatches("qux"::equals));
   }
 
+  Duration testFluxAsStepVerifierExpectNext() {
+    return Flux.just(1)
+        .collect(toImmutableList())
+        .as(StepVerifier::create)
+        .assertNext(list -> assertThat(list).containsExactly(2))
+        .verifyComplete();
```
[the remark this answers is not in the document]
rickie: Ahh, we can do the `<?>` as return type indeed.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
rickie: I agree, I'll add a commit to drop the second test case.

rickie:
Will merge once 🍏.

PR #619, opened 2023-05-09
Introduce `StreamsConcat` Refaster rule

mlrprananta:
~:exclamation: This PR is on top of #623. :exclamation:~

Small one that came up recently 🙃 

Suggested commit message:
```
Introduce `StreamsConcat` Refaster rule (#619)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
@@ -623,4 +623,16 @@ final class StreamRules {
       return stream.flatMap(mapper).collect(collector);
     }
   }
+
+  static final class StreamConcat<T> {
+    @BeforeTemplate
+    Stream<T> before(Stream<T> s1, Stream<T> s2) {
+      return Stream.of(s1, s2).flatMap(Function.identity());
+    }
+
+    @AfterTemplate
+    Stream<T> after(Stream<T> s1, Stream<T> s2) {
+      return Stream.concat(s1, s2);
+    }
+  }
 }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
@@ -625,20 +625,9 @@ final class StreamRules {
     }
   }
 
-  static final class StreamConcat<T> {
-    @BeforeTemplate
-    Stream<T> before(Stream<T> s1, Stream<T> s2) {
-      return Stream.of(s1, s2).flatMap(Function.identity());
-    }
-
-    @AfterTemplate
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -624,4 +625,28 @@ R after(
...
+    @BeforeTemplate
+    Stream<T> before(Stream<T> s1, Stream<T> s2) {
+      return Stream.of(s1, s2).flatMap(Function.identity());
+    }
+
+    @AfterTemplate
+    Stream<T> after(Stream<T> s1, Stream<T> s2) {
+      return Stream.concat(s1, s2);
+    }
+  }
```
mlrprananta, line 639: This might be redundant if `StreamsConcat` turns `Stream.of(s1, s2).flatMap(Function.identity())` into `Streams.concat(s1, s2)` and `ConcatTwoStreams` turns it into `Stream.concat(s1, s2)`

mlrprananta: Dropped the rule 👍

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -626,20 +626,9 @@ R after(
...
-
-    @AfterTemplate
-    Stream<T> after(Stream<T> s1, Stream<T> s2) {
-      return Stream.concat(s1, s2);
-    }
-  }
-
   static final class StreamsConcat<T> {
     @BeforeTemplate
+    @SuppressWarnings("StreamOfArray")
```
Stephan202: Let's rebase on #623 :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -624,4 +625,17 @@ R after(
       return stream.flatMap(mapper).collect(collector);
     }
   }
+
+  static final class StreamsConcat<T> {
+    @BeforeTemplate
+    @SuppressWarnings("StreamOfArray")
+    Stream<T> before(@Repeated Stream<T> stream) {
+      return Stream.of(Refaster.asVarargs(stream)).flatMap(Function.identity());
```
Stephan202: It won't have functional impact, but we generally statically import `Function.identity()`.

Something that is functionally relevant, is that _for now_, we don't have a rule that favours `identity()` over `v -> v` or vice versa. (There are technical reasons for that.) So let's also capture that case.

Stephan202 (approved):
Rebased and added a commit. Nice!

rickie (approved):
Changes LGTM!

Stephan202:
Rebased; PR now targets the main branch. Will merge once built.

Issue #389, opened 2022-12-05
Prefer `BugCheckerRefactoringTestHelper#doTest(TestMode.TEXT_MATCH)` over alternatives


rickie:
### Problem

This is specific to the `BugCheckerRefactoringTestHelper` which is used in [replacement tests](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/EmptyMethodTest.java#L71-L92) for BugPatterns.
There are two `TestMode`s, namely `AST_MATCH` and `TEXT_MATCH`. During multiple code reviews we noticed that using `AST_MATCH` can hide bugs. Therefore we always prefer `TEXT_MATCH`. 

Besides that, we do not like the statically imported `TEXT_MATCH`, as it results in having a little less context. Therefore we want to rewrite all these occurrences as well, using Refaster.

### Description of the proposed new feature

- [x] Support a stylistic preference.
- [x] Avoid a common gotcha, or potential problem.

We would like to rewrite the following code using Refaster:
```java
refactoringTestHelper.addInputLines("A.java", "").addOutputLines("A.java", "").doTest();
refactoringTestHelper.addInputLines("B.java", "").addOutputLines("A.java", "").doTest(TestMode.AST_MATCH);
refactoringTestHelper.addInputLines("C.java", "").addOutputLines("A.java", "").doTest(TEXT_MATCH);
refactoringTestHelper.addInputLines("C.java", "").addOutputLines("A.java", "").doTest(BugCheckerRefactoringTestHelper.TestMode.TEXT_MATCH);
```

to:
```java
refactoringTestHelper.addInputLines("A.java", "").addOutputLines("A.java", "").doTest(TestMode.TEXT_MATCH);
refactoringTestHelper.addInputLines("B.java", "").addOutputLines("B.java", "").doTest(TestMode.TEXT_MATCH);
refactoringTestHelper.addInputLines("C.java", "").addOutputLines("C.java", "").doTest(TestMode.TEXT_MATCH);
refactoringTestHelper.addInputLines("C.java", "").addOutputLines("C.java", "").doTest(TestMode.TEXT_MATCH);
```
### Considerations

The rewrite from `TestMode.AST_MATCH` to `TestMode.TEXT_MATCH` is not a behavior preserving one. It's important to make a note about that by using `@Description` on the Refaster rule.

Stephan202:
[...]

PR #630, opened 2023-05-16
Align TestMode import across test classes

philleonard:
Small cleanup to align with other test classes in this package.

🧹 Whilst an ad-hoc cleanup when working on something else, I'm not sure it _can_ be generalised (and likely one you have already considered so I'm interested to know your opinions!).

What factors come into play when deciding between an import and a static import for accessing static members of a class? Ones I consider that cannot be generalised (trivially):
- Trade-off between the class qualification providing extra descriptive value, and adding to verbosity. Requires understanding the descriptive value of a the qualifying class name in the context it is (statically or not) imported.
- Frequency as a multiplier. Within one class I assume a check would be trivial, across many classes of the same structure not trivial with EP. In this case, it is common to all(?) `#replacement` tests and therefore you could argue we could drop the qualifier. This leads onto another question: should we not provide some form of test abstraction (harness) for `BugChecker` tests? Might also be nice to look at a cleaner test data setup (or perhaps there is a reason to providing varargs String's in `CompilationTestHelper#addSourceLines` over reading from a `.java` source file?

Suggested commit message (not much to say about this change so probably too verbose/redundant):

``` 
Qualify non-static `TestMode` imports across Bug Pattern test classes (#630)

Prefer non-static imports and qualification of `BugCheckerRefactoringTestHelper#TestMode` 
members across Bug Pattern test classes.

```


rickie:
Heyy @philleonard , thanks for opening the PR! :rocket: 

First some context related to this specific issue and the question about generalization:
- We have a ticket that is related to this specific issue: https://github.com/PicnicSupermarket/error-prone-support/issues/389. 
- @cernat-catalin opened[ a PR](https://github.com/PicnicSupermarket/error-prone-support/pull/450) for a related check that is the exact opposite of the `StaticImport` check.

Nice thoughts, and yes you are right, it is not easy to generalize this and auto-detect candidates for static importing. For that reason we have a [curated set of candidates](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java#L61) for statically importing things.  

We have the rule that we should not "lose" any context if we statically import something. This usually holds for imports that have 
words (or parts) that are (partially) duplicate with the method. Additionally it should not be ambiguous. For example, if there are multiple places from which you can import the constant `UTC`, it is not directly clear from where you imported something, so we don't want to do it.

I have to go now so will get back to your question about the test abstraction later 😄.

Stephan202 (approved):
Generally we try to combine PRs such as this one with a check that enforces the change, but as @rickie says we already have #450 for this. I'm fine with merging this PR as-is :)

I tweaked the suggested commit message :)

> This leads onto another question: should we not provide some form of test abstraction (harness) for `BugChecker` tests? Might also be nice to look at a cleaner test data setup, or perhaps there is a reason to providing varargs String's in `CompilationTestHelper#addSourceLines` over reading from a `.java` source file?

Error Prone does support moving the test code to separate resource files, but in practice this makes test maintenance harder, and as such we see that new upstream Error Prone tests also prefer the current approach.

That said... yes, we can do better :smile:. See #198, where we'll move to text blocks. That in turn unlocks IntelliJ IDEA language injection support :)

philleonard:
> where we'll move to text blocks. That in turn unlocks IntelliJ IDEA language injection support :)

Nice! Having played around with #198 this is already a much nicer experience for writing these test classes (which I have been doing in a scratch file for now). Copilot suggestions and IntelliJ autocomplete also "work" within language injections  so that is nice. After a quick scan, GitHub and the IntelliLang both have somewhat "incomplete" syntax highlighting, but already much better than what we currently have.

rickie (approved):

PR #636, opened 2023-05-18
Introduce `TypeMemberOrder` check

### LINKED ISSUE #595: Sort member variables, constructors, and methods
### Problem

We want to sort the order of the members in a class. 
Internally we are quite strict with our ordering for the sake of consistency.

The following order should be enforced in a class:
- static member variables
- non-static member variables
- constructor
- methods

- [ ] Improve performance.

Create an Error Prone check that changes the order of the members of a class without altering the empty lines between the members of the class.

```java
class A {
  char a = 'a';
  private static String FOO = "foo";
  static int ONE = 1;

  void m2() {}

  public A () {}

  private static String BAR = "bar";
  char b = 'b';

  void m1() {}
  static int TWO = 2;
}
```

to:
```java
class A {
  private final static String FOO = "foo";
  static int ONE = 1;
  private final static String BAR = "bar";

  static int TWO = 2;

  char a = 'a';

  char b = 'b';
  public A () {}

  void m1() {}
  void m2() {}
}
```

### Considerations

- For this check we don't want to sort based on the modifiers, this will be done later or in a separate Error Prone check.
- We don't want to change anything w.r.t. the empty lines, so we want to replace only on the lines where we already had a member variable defined. For that reason the example has a few variables below a method, usually we only have these declarations at the top of the class.

benhalasi:
#595

### Suggested commit message
```
Introduce `TypeMemberOrder` check (#636)
```


benhalasi:
Verifying that the whitespace is untouched may be a bit harder than I thought because [`TestMode.TEXT_MATCH` formats both input and output](https://github.com/google/error-prone/blob/31d420b282603d92ad105defed2f47c57ffd1691/test_helpers/src/main/java/com/google/errorprone/BugCheckerRefactoringTestHelper.java#L95)

rickie (commented):
There were some imports from `ASTHelpers` that we usually do not statically import :).

Added a commit! The setup looks nice, didn't get till the end, but already flushing my comments 😄.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
```diff
@@ -0,0 +1,217 @@
...
+
+  private static Stream<Tokens.Comment> getComments(
+      ClassTree classTree, Tree member, VisitorState state) {
+    return getTokensBeforeMember(classTree, member, state).findFirst().stream()
+        .map(ErrorProneToken::comments)
+        // xxx: Original impl sorts comments, but that seems unnecessary.
+        .flatMap(List::stream);
+  }
+
+  // XXX: From this point the code is just a fancy copy of `SuggestFixes.replaceIncludingComments` -
```
rickie, line 134: We can indeed drop this `replaceIncludingComments` method! Will push :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
```diff
@@ -0,0 +1,196 @@
...
+    for (int i = 0; i < tokens.size() - 1 /* EOF */; i++) {
+      if (tokens.get(i).kind() != memberTokens.get(i).kind()
+          || tokens.get(i).hasName() != memberTokens.get(i).hasName()
+          || (tokens.get(i).hasName()
+              && !tokens.get(i).name().equals(memberTokens.get(i).name()))) {
+        return false;
+      }
+    }
+    return true;
+  }
```
oxkitsune, line 177:
```suggestion
  /** Returns the class' members with their comments. */
  private static ImmutableList<MemberWithComments> getMembersWithComments(
      ClassTree classTree, VisitorState state) {
    List<ErrorProneToken> tokens =
        state.getOffsetTokens(
            ASTHelpers.getStartPosition(classTree), state.getEndPosition(classTree));

    ImmutableList.Builder<MemberWithComments> membersWithComments = ImmutableList.builder();
    for (Tree member : classTree.getMembers()) {
      int memberStartPos = ASTHelpers.getStartPosition(member);
      // we start and the end position of the previous token
      // this gives us a range that includes any comments
      int startPos =
          tokens.stream()
              .map(ErrorProneToken::endPos)
              .filter(i -> i < memberStartPos)
              .reduce((first, second) -> second)
              .orElse(memberStartPos);

      ImmutableList<ErrorProneToken> memberTokens =
          ImmutableList.copyOf(state.getOffsetTokens(startPos, state.getEndPosition(member)));
      if (memberTokens.isEmpty() || memberTokens.get(0).kind() == Tokens.TokenKind.EOF) {
        continue;
      }

      membersWithComments.add(
          new MemberWithComments(member, ImmutableList.copyOf(memberTokens.get(0).comments())));
    }
    return membersWithComments.build();
  }
```
`ErrorProneTokens#getTokens(String, Context)` doesn't include the comments because it uses the start position of the provided member, e.g. the position of `void`. And while comments are not tokens, they do offset the position of tokens.

Using `VisitorState#getOffsetTokens(int, int)` we're able to explicitly choose the start/end positions of the token range. If we start the range at the end position of the previous token and end it at the end of the current member, we get the same tokens, but since the range now includes the offset for any comments these will be included in the resulting `ErrorProneToken`.

oxkitsune:
`ErrorProneTokens#getTokens(String, Context)` doesn't include comments as it relies on the start position of the provided member (e.g., the position of `void`). Comments don't count as tokens but do affect token positions.

To include comments, we can use `VisitorState#getOffsetTokens(int, int)` to obtain tokens in a specific range. By starting the range at the end position of the previous token and ending it at the end of the current member, we'll obtain the same tokens, now including comments in the resulting `ErrorProneToken`.

benhalasi:
Thanks oxkitsune for the `ErrorProneTokens#getTokens(String, Context)`!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
```diff
@@ -0,0 +1,187 @@
...
+    if (member.getKind() == Tree.Kind.METHOD
+        && ASTHelpers.isGeneratedConstructor((MethodTree) member)) {
+      return ImmutableList.of();
+    }
+
+    checkState(
+        state.getEndPosition(member) != -1,
+        "Member's end position is not available (-1).\n - member=[%s]\n - source=[%s]",
+        member,
+        state.getSourceForNode(member));
```
benhalasi, line 148: 100% overkill, but it would be nice to not forget about this case or verify that it can never happen.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
```diff
@@ -0,0 +1,187 @@
...
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = STYLE)
+public final class MemberOrdering extends BugChecker implements BugChecker.ClassTreeMatcher {
+  private static final long serialVersionUID = 1L;
+
+  /** A comparator that sorts variable and method (incl. constructors) in a standard order. */
+  private static final Comparator<Tree> MEMBER_SORTING =
+      comparing(
+          (Tree tree) -> {
```
benhalasi, line 52: Saw this being renamed to `tree` from `memberTree`, but in other placed we renamed `memberTree` to `classMemberTree`.

Personally I don't mind either. I originally went with member tree because afaik, only classes have members in Java, specifying `class` would be redundant. One thing to note, if we go with `classMember*`, we should rename this BugChecker to `ClassMemberOrdering`

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
oxkitsune: I ran into this exact issue while testing, to my knowledge this case can only happen for generated members (e.g. generated constructors)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
oxkitsune: > Saw this being renamed to tree from memberTree while in other places we renamed memberTree to classMemberTree. Do you think we can use the same strategy everywhere?

I think this is purely based on the type we're referring to, e.g. `Tree` -> `tree`.
And since we renamed `MemberTreeWithComments` to `ClassMemberTreeWithComments` the variable names should reflect that.

> I originally went with member tree because afaik, only classes have members in Java, specifying class would be redundant.

`ClassMemberTreeWithComments` is a name @rickie preferred. I think it does add some clarity. That being said, [JEP 445](https://openjdk.org/jeps/445) introduces unnamed classes which could be a concern for the `MemberTreeWithComments` name :joy:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
benhalasi: I was thinking that Lombok and tools like it could also behave like this.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
```diff
@@ -0,0 +1,187 @@
...
+    ImmutableList<ErrorProneToken> tokens =
+        ImmutableList.copyOf(
+            state.getOffsetTokens(
+                ASTHelpers.getStartPosition(classTree), state.getEndPosition(classTree)));
+
+    int memberStartPos = ASTHelpers.getStartPosition(member);
+    Optional<Integer> previousMemberEndPos =
+        tokens.stream()
+            .map(ErrorProneToken::endPos)
+            .takeWhile(endPos -> endPos <= memberStartPos)
```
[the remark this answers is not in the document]
benhalasi: I did some investigation on the `<` or `<=` question. I don't think, it matters too much which way we go. It would matter if we could place a comment on a token that immediately follows another. This is impossible because if we comment a token, the token immediately follows its comment instead of its prior token.

```
// `int b` immidately follows the token `;`
int a;int b;
// `int b` doesn't follow immidately any error-prone token.
int a;/* comment */int b;
```
An interesting case would be a code like this:
```
...
/* comment */
;void method() { ... }
```
In this case, the comment belongs to the token `;` and not the method. Our current implementation will move the comment with the `void method() { ... }` instead of leaving it with the `;` token. This aligns with the implementation of `SuggestedFixes#replaceIncludingComments` and it provides a better UX than the alternative, but AFAIK it is technically incorrect.

I've added a test case for this, but again, this will be resolved by the simplest formatters.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
benhalasi: This makes two of us, I've applied changes assuming that `state.getEndPosition(member)` is only unavailable for members that are not in the source file.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
```diff
@@ -0,0 +1,187 @@
...
+                new ClassMemberWithComments(member, getMemberComments(state, classTree, member)))
+        .collect(toImmutableList());
+  }
+
+  private static ImmutableList<Tokens.Comment> getMemberComments(
+      VisitorState state, ClassTree classTree, Tree member) {
+    if (member.getKind() == Tree.Kind.METHOD
+        && ASTHelpers.isGeneratedConstructor((MethodTree) member)) {
+      return ImmutableList.of();
+    }
```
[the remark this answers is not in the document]
benhalasi: See thread above, I've replaced this condition with `state.getEndPosition(member) == -1` as the previous solution might have broke projects that use Lombok or tools like it.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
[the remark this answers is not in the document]
oxkitsune: Interesting case! 
I originally used `<=` over `<`, as I figured there was no reason to exclude the start position of the token, if for some reason a previous token has that as end position.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MemberOrdering.java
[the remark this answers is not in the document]
oxkitsune: Ah I forgot about Lombok :joy:. SGTM

oxkitsune:
Looks good!

However blocks (both static and instance) are also considered class members:
```java
class Foo {
	private static final int bar;
	private final int baz;


	// some static initializer block
	static {
		
		bar = 110 ^ 67;
	}

	// some instance initializer block
	{
		baz = (1395 ^ 1059) >> 3;
	}
}
```

Currently these are not sorted, I'm unsure what their order should be though. 
Something like this perhaps?
- static fields
- static *blocks*
- non-static fields
- non-static *blocks*

WDYT :smile:

benhalasi:
> However blocks (both static and instance) are also considered class members: ...

Good point, they're indeed non-handled. My reasoning to ignore this case was that using static blocks are very rare and specific. It would be hard to put these very specific edge cases into a rule, if even possible. Seems more practical to not sort them in this rule (or at all).

Afaik the general opinion on static blocks is that we should just not use them. Considering this, (imo) if we want to do anything with static blocks in error-prone, we should just flag them as errors.

Stephan202 (commented):
Rebased and added a commit. Didn't finish reviewing yet; timing of next round is TBD. Nice work! :rocket:

Stephan202 (commented):
Apparently back in October I didn't push my review comments :facepalm:. Doing so now; not sure what the state of this PR is; would need to get back to that some other time.

benhalasi:
Hehe, I'll check it out *soon* (this week 🤞) and will update it with the open questions!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ClassMemberOrdering.java
```diff
@@ -0,0 +1,181 @@
...
+    List<ErrorProneToken> classMemberTokens =
+        state.getOffsetTokens(previousClassTokenEndPos.orElse(memberStart), memberEnd);
+
+    return classMemberTokens.get(0).comments().stream()
+        .map(Tokens.Comment::getText)
+        .collect(toImmutableList());
+  }
+
+  @AutoValue
+  abstract static class ClassMemberWithComments {
```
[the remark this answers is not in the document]
benhalasi: Considering the `Class -> Type` rename we're doing, wdyt of `TypeMember` or just `Member`?

Asking because `TypeMember` sounded weird to me initially. Probably I'm just not used to it.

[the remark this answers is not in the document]
Stephan202: Yeah, `TypeMember` can work! (Doesn't sound so weird to me :smile:.)

benhalasi:
Seems PiTest in CI consistently fails in ~1m w/
```
Failed to execute goal org.pitest:pitest-maven:1.15.3:mutationCoverage (default-cli) on project error-prone-contrib: Execution default-cli of goal org.pitest:pitest-maven:1.15.3:mutationCoverage failed: Coverage generation minion exited abnormally! (UNKNOWN_ERROR)
```
Didn't deep-dive in its logs nor tried executing it locally yet, just flagging!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/TypeMemberOrdering.java
```diff
@@ -0,0 +1,236 @@
...
+    //  Loop once and identify for all members
+    // TODO: Check if this handles properly comments on the first member.
+    Optional<Integer> previousMemberEndPos =
+        state.getOffsetTokens(typeStart, typeEnd).stream()
+            .map(ErrorProneToken::endPos)
+            .takeWhile(endPos -> endPos < memberStart)
+            .reduce((earlierPos, laterPos) -> laterPos);
+
+    List<ErrorProneToken> typeMemberTokens =
+        state.getOffsetTokens(previousMemberEndPos.orElse(memberStart), memberEnd);
```
benhalasi, line 222: This TODO is for addressing the very not performant ways of this solution.

The challenge here to keep the comment and maybe other code too between members.

Options (I can think of atm):

---

- Somehow get all `ErrorProneTokens` (or source) of the class,
- divide them for each member, and
- reorder these token+member groups.

I believe this was described in this [comment](https://github.com/PicnicSupermarket/error-prone-support/pull/636#discussion_r1362763320).

With this I'm not sure how we can handle the whitespace properly, but I'm yet to fully discover this option.

---

- Go through the type's tokens once,
- gather all commented tokens, and put them in a `Map<Identifier, Comments>`
- use this map to get the comments of a member (`tree.getMembers()`)

I tried this quickly, using `ErrorProneToken#pos` and `ASTHelpers.getStartPosition(member)` as identifier.

This worked in general, but failed the `detached comment from method` test and the `replacementSuggestedFixDoesNotModifyWhitespace`.

---
---

Additional Note:

`VisitorState#getOffsetTokens` java doc says:
> This is moderately expensive (the source of the node has to be re-lexed), so it should only be used if a fix is already going to be emitted.

With this in mind, we should first determine if the members are ordered properly, and only then fetch the comments for them - if we opt to use this helper.

benhalasi (commented):
Open questions:
 - Specifically how to obtain the comments of a member / source between two members - here I have some ideas I'll try them in a timely manner :D
 - Whether regular identification and re-placement tests are sufficient to fully cover our logic - I believe this is highly dependent on how we resolve the comments.
- And some seemingly minor suggestions open to address.

Stephan202:
> Seems PiTest in CI consistently fails in ~1m w/
> 
> ```
> Failed to execute goal org.pitest:pitest-maven:1.15.3:mutationCoverage (default-cli) on project error-prone-contrib: Execution default-cli of goal org.pitest:pitest-maven:1.15.3:mutationCoverage failed: Coverage generation minion exited abnormally! (UNKNOWN_ERROR)
> ```
> 
> Didn't deep-dive in its logs nor tried executing it locally yet, just flagging!

If you rebase this should be fixed, thanks to #976 :)

W.r.t. other questions: I'll have to get back to you, hopefully later this week. Tnx for reviving this PR!

Stephan202:
Hey @benhalasi! I figured out why I hadn't pushed that October review until recently. I found a local branch with unpushed changes; apparently I just abandoned the review midway. I suspect that by now you've applied the relevant feedback, but just to be sure I pushed a commit with my local changes: 997099e957ecc2281b75b652d45fa8a3a0b9aaa1; perhaps there's something useful in it still (didn't check).

Will circle back to this PR later!

benhalasi:
Small progress update: I rewrote most of the checker's logic, (much inspired by https://github.com/PicnicSupermarket/error-prone-support/commit/997099e957ecc2281b75b652d45fa8a3a0b9aaa1).

The main changes:
- track unmovable members too so we can track source-code between members - including unmovable ones too.
- accommodate enums. I left `XXX: To accommodate enums` comments at their support's main pillars, and it's not even that horrifying!
- combine `BY_PREFERRED_TYPE_MEMBER_ORDER` Comparator with the logic that determines which members are movable.
- split the tests between classes, interfaces, and enums.

I'm yet to sleep one on it and have a final round of changes mark parts where I have minor questions, but ofc any input is appreciated!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/TypeMemberOrder.java
```diff
@@ -0,0 +1,265 @@
...
+       * that (part of) its code was generated. Even if the source code for a subset of its members
+       * is available, dealing with this edge case is not worth the trouble.
+       */
+      return Description.NO_MATCH;
+    }
+
+    return describeMatch(tree, sortTypeMembers(bodyStartPos, members, state));
+  }
+
+  private static boolean isGeneratedConstructor(Tree tree) {
```
rickie: We can maybe move this to `MoreAstHelpers` 🤔 , although it's used only once right now..

WDYT of this implementation:
```
  public static boolean isGeneratedConstructor(Tree tree) {
    return tree instanceof MethodTree methodTree && ASTHelpers.isGeneratedConstructor(methodTree);
  }
```

rickie: It's inversed, it returns true if `Tree` is  not of type `Kind.METHOD`, will push a commit to fix :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/TypeMemberOrder.java
```diff
@@ -0,0 +1,265 @@
...
+ *   <li>Constructors
+ *   <li>Methods
+ *   <li>Nested classes, interfaces and enums
+ * </ol>
+ *
+ * @see <a
+ *     href="https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/DeclarationOrderCheck.html">Checkstyle's
+ *     {@code DeclarationOrderCheck}</a>
+ */
+// XXX: Reference
```
rickie: I think we can drop this :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/TypeMemberOrder.java
```diff
@@ -0,0 +1,265 @@
...
+// https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/DeclarationOrderCheck.html
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Type members should be ordered in a standard way",
```
rickie:
```suggestion
    summary = "Type members should be defined in a canonical order",
```
What about this?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/TypeMemberOrder.java
```diff
@@ -0,0 +1,265 @@
...
+  public TypeMemberOrder() {}
+
+  @Override
+  public Description matchClass(ClassTree tree, VisitorState state) {
+    Kind kind = tree.getKind();
+    if (kind != Kind.CLASS && kind != Kind.INTERFACE && kind != Kind.ENUM) {
+      return Description.NO_MATCH;
+    }
+
+    // All members that can be moved or may lay between movable ones.
```
rickie: If we leave comments in the class, we usually have the `/* */` style.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/TypeMemberOrderInterfaceTest.java
```diff
@@ -0,0 +1,198 @@
...
+            "  @SuppressWarnings(\"TypeMemberOrder\")",
+            "  static final int baz = 1;",
+            "",
+            "  static enum InnerEnum {}",
+            "}")
+        .doTest(TestMode.TEXT_MATCH);
+  }
+
+  @Test
+  void replacementSuggestedFixConsidersDanglingComments() {
```
rickie: I think we can simplify the name of the tests :).

rickie (commented):
We discussed this earlier, but I think it would be nice to introduce this check in the `error-prone-experimental` module. From there we continue validation and move it to `error-prone-contrib` after that.

Man, this check is really cool but takes some time to review. Did another pass and had some style suggestions/tweaks. Please take a look and feel free to tweak :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/TypeMemberOrder.java
```diff
@@ -0,0 +1,260 @@
...
+    };
+  }
+
+  /**
+   * Returns the start position of the body of the given type, in the case of enums, it returns the
+   * position that follows the enumerated type's enumerations.
+   */
+  private static int getBodyStartPos(ClassTree tree, VisitorState state) {
+    CharSequence sourceCode = state.getSourceCode();
+    int typeStart = ASTHelpers.getStartPosition(tree);
```
rickie, line 154: Usually we call this `startPosition`, I think we can do the same here 🤔, not sure if we should make it specific to add "type"? Or am I missing a reason to do so?

rickie (commented):
Sorry I didn't put comments everywhere. The main idea behind the changes is that we now do only one "scan" and collect only the members that we can actually change. After that we check is the list sorted, and if not we do the same as before. But now, we know for sure that we only have items that we can and want to switch around :). Makes it a bit easier as we don't need to check for empty/present optionals and have all the data. 

So there is still a specific case not working and I didn't finish reviewing all the tests yet, but man this will be an epic check.

Oh and we should still move this to experimental :).

Credits to @Stephan202 as we paired on this.

rickie:
It's not working as this check is in the experimental module 😬. So we will need to tweak something, or to run it locally 😄 @benhalasi.

PR #641, opened 2023-05-21
Introduce `IllogicalReactorOperators` bug check

mohamedsamehsalah:
Draft PR to discuss an idea for a bug check observed in a Picnic PR.

I can provide the link to the original bug in private :wink:


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoFilterThen.java
```diff
@@ -0,0 +1,53 @@
...
+  /** Instantiates a new {@link MonoFilterThen} instance. */
+  public MonoFilterThen() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (MONO_THEN.matches(tree, state)
+        && ASTHelpers.getSymbol(ASTHelpers.getReceiver(tree))
+            .getSimpleName()
+            .toString()
+            .equals("filter")) {
```
mohamedsamehsalah, line 47: Ultimately, we check for all non-terminal operators here.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoFilterThen.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.errorprone.suppliers.Suppliers;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Type;
+
+/** TODO: Write Javadoc. */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "You probably meant .filter.when(), right ?",
```
Stephan202, line 26: The `when` method is static, so likely not ;)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoFilterThen.java
```diff
@@ -0,0 +1,53 @@
...
+    link = BUG_PATTERNS_BASE_URL + "MonoFilterThen",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = LIKELY_ERROR)
+public final class MonoFilterThen extends BugChecker implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Supplier<Type> MONO =
+      Suppliers.typeFromString("reactor.core.publisher.Mono");
```
Stephan202, line 34: Conceptually this check also applies to `Flux`es, right?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoFilterThen.java
```diff
@@ -0,0 +1,53 @@
...
+  private static final Matcher<ExpressionTree> MONO_THEN =
+      instanceMethod().onDescendantOf(MONO).named("then");
```
Stephan202, line 36: If we go ahead with this check then there are a few other "element dropping" operations we may want to flag.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoFilterThen.java
```diff
@@ -0,0 +1,53 @@
...
+  public MonoFilterThen() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (MONO_THEN.matches(tree, state)
+        && ASTHelpers.getSymbol(ASTHelpers.getReceiver(tree))
+            .getSimpleName()
+            .toString()
+            .equals("filter")) {
+      return describeMatch(tree);
```
Stephan202, line 48: My proposal would be to emit a suggested fix that drops the presumed no-op operator (`filter` in this case) :)

Stephan202 (commented):
Nice idea! :).

This relates to an idea that came up a few times now: each Reactor operator has some set of properties, and some of those properties are incompatible/non-sensical when combined, so ideally we flag such cases. If we'd have such a framework, then the check proposed here could be incorporated by flagging the composition of operators with the "should be pure" and "drops elements" properties, respectively.

... but that's a decidedly bigger task :sweat_smile: So fleshing out a check as proposed here SGTM!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoFilterThen.java
mohamedsamehsalah: I added something like
```
ImmutableSet.of("then", "thenEmpty", "thenMany")
```

Is there a reference/docs of other operators that can be good candidates for this ?

mohamedsamehsalah:
No plans to progress on this. Need better spec.

Closing for now.

PR #642, opened 2023-05-21
Introduce `NestedPublishers` bug check

mohamedsamehsalah:
Bug checker that gives a warning when nested publishers are used.

Suggested commit message:
```
Introduce `NestedPublishers` check (#642)
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -782,6 +782,7 @@ Flux<S> after(Flux<T> flux) {
   /** Prefer {@link Mono#flatMap(Function)} over more contrived alternatives. */
   static final class MonoFlatMap<S, T> {
     @BeforeTemplate
+    @SuppressWarnings("MonoOfPublishers")
```
mohamedsamehsalah, line 785: Does this mean we do not need these Refaster rules, since bug checker will warn this statement ?

rickie: Well yes there is overlap, but the `MonoOfPublishers` doesn't do autofixes. So if we suppress these rules, than some of the violations are auto-fixed by Refaster, which is nice 😄. Generally if there is overlap and the check comes with `SuggestedFix`es, we'd drop the Refaster rule./

mohamedsamehsalah: Makes sense 💯  😄

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoOfPublishers.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary =
+        "Avoid nesting `Publisher`s inside `Mono`s; the resultant code is hard to reason about",
+    link = BUG_PATTERNS_BASE_URL + "MonoOfPublishers",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = FRAGILE_CODE)
+public final class MonoOfPublishers extends BugChecker implements MethodInvocationTreeMatcher {
```
Venorcis, line 38: is it possible to make this something like `NestedPublishers` instead? do we care if it's a `Mono` specifically?

mohamedsamehsalah: At this point, as pointed out by @Stephan202 in the original comment (That I can't link here as it is a private repo), we do not want to flag Flux of publishers because this will warn about all `groupBy` operator usage.

Stephan202: Indeed, it would effectively disallow [groupBy](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#groupBy-java.util.function.Function-) usage. To be fair, that is a very sharp tool indeed, but we do use it in places. In fact, the `FluxFlatMapUsage` has [special logic](https://github.com/PicnicSupermarket/error-prone-support/blob/c53a3f64b6ecf40e8b70f56e6bf5d4cb6d8f5acc/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/FluxFlatMapUsage.java#L90-L99) for `Flux`es of nested publishes.

Venorcis: Ah yeah the `GroupedFlux` 🤪 Wouldn't it be possible to make an exception for it here as well though?

Venorcis: @mohamedsamehsalah it's the same in the example shared by @Stephan202 right? again I'm not 100% sure it's possible, but [there](https://github.com/PicnicSupermarket/error-prone-support/blob/c53a3f64b6ecf40e8b70f56e6bf5d4cb6d8f5acc/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/FluxFlatMapUsage.java#L90-L99) we first check if `FLUX_FLATMAP ` matches and then do a more specific check (which there goes to 2 different cases, but here 1 case could go to `Description.NO_MATCH` right?)

rickie: > Is this even feasible, since we are matching on the Class and generic level 

We can specify anything we want by using the methods from `MoreTypes`. There is a nice test that allows us to more clearly see how we can use these types to match specific types with generics or [unbounded](https://www.tutorialspoint.com/what-are-unbounded-wildcard-w-r-t-generics-method-in-java#:~:text=An%20unbounded%20wildcard%20is%20the,to%20declare%20an%20unbounded%20wildcard).  

[Here](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/MoreTypesTest.java#L122-L123) we have a test `BugChecker` that loops over the list of [test types](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/MoreTypesTest.java#L154) and goes over all the `MethodInvocation`s [here](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/MoreTypesTest.java#L44-L111) and reports method invocations that are a subtype of any type defined in the list of test types.

Okay this get's rather complicated and sounds more daunting than it is to check for `GroupedFlux` in an if statement, that should be doable. I think @Venorcis  is right and we can do something like that. That would make the check more generic as well as we would also consider `Flux`es. 

If you are like: hmm not sure if I fully understand & this gets a little more complicated than expected, I can just swing by and we can look at it together 😄. That is totally fine.

mohamedsamehsalah: > If you are like: hmm not sure if I fully understand & this gets a little more complicated than expected...

I do need time to digest this indeed 😅 ..

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java
```diff
@@ -0,0 +1,74 @@
...
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = FRAGILE_CODE)
+public final class NestedPublishers extends BugChecker implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Supplier<Type> PUBLISHER = type("org.reactivestreams.Publisher");
+  private static final Supplier<Type> PUBLISHER_OF_PUBLISHERS =
+      VisitorState.memoize(
+          generic(PUBLISHER, subOf(generic(type("org.reactivestreams.Publisher"), unbound()))));
+  private static final Supplier<Type> GROUPED_FLUX =
```
mohamedsamehsalah, line 44: Should we extend this to all classes that extend either a `Mono` or `Flux` ?

Reference for question: 
```
public abstract class GroupedFlux<K, V> extends Flux<V> {
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -435,6 +435,7 @@ Flux<T> after(Flux<T> flux) {
   /** Prefer {@link Flux#concatMap(Function)} over more contrived alternatives. */
   static final class FluxConcatMap<T, S> {
     @BeforeTemplate
+    @SuppressWarnings("NestedPublishers")
```
mohamedsamehsalah, line 438: Do we need this & the below 'before' checks here still ?

mohamedsamehsalah (commented):
Proposal from pair-programming session with @rickie

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoOfPublishers.java
mohamedsamehsalah: See [a27838d](https://github.com/PicnicSupermarket/error-prone-support/pull/642/commits/a27838d952e7fe86e64c864ae58545b9ad484b5e).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java
```diff
@@ -0,0 +1,74 @@
...
+    link = BUG_PATTERNS_BASE_URL + "NestedPublishers",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = FRAGILE_CODE)
+public final class NestedPublishers extends BugChecker implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Supplier<Type> PUBLISHER = type("org.reactivestreams.Publisher");
+  private static final Supplier<Type> PUBLISHER_OF_PUBLISHERS =
+      VisitorState.memoize(
+          generic(PUBLISHER, subOf(generic(type("org.reactivestreams.Publisher"), unbound()))));
```
Venorcis, line 43:
```suggestion
          generic(PUBLISHER, subOf(generic(PUBLISHER, unbound()))));
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.errorprone.suppliers.Supplier;
+import com.google.errorprone.suppliers.Suppliers;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Type;
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Mono;
+
+/**
+ * A {@link BugChecker} that flags nesting of {@link Publisher Publishers} inside {@link Mono
```
Venorcis, line 27:
```suggestion
 * A {@link BugChecker} that flags nesting of {@link Publisher Publishers}.
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java
```diff
@@ -0,0 +1,74 @@
...
+
+  /** Instantiates a new {@link NestedPublishers} instance. */
+  public NestedPublishers() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    Type publisherOfPublisherType = PUBLISHER_OF_PUBLISHERS.get(state);
+    Type groupedFluxType = GROUPED_FLUX.get(state);
+    Type treeType = ASTHelpers.getType(tree);
+    if ((publisherOfPublisherType == null
```
Venorcis, line 59: this if-statement is very hard to read for me; consider introducing some more (`private static`) functions for it (and avoid the nesting somehow)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java
Venorcis: not sure I understand the question; you're matching `GroupedFlux` with any `<K,V>` right (unbounded)?

mohamedsamehsalah: Sorry, I re-read my question and realised it is not clear 😓 .

I meant, we are using Type `GROUPED_FLUX` to exclude something like this `Flux<GroupedFlux<K,V>>` in the `matchMethodInvocation` method.

Since `GroupedFlux` extends `Flux`, I was asking if we can make `GROUPED_FLUX` type more generic and to match on something anything extending either a flux or a mono like `? extends Flux` ( See also `ParallelFlux`, `ConnectableFlux`). 

Do I make more sense now ?
It also could be there is something I am missing 😅

Venorcis (approved):
LGTM 👍🏻

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java
mohamedsamehsalah: Discussed offline: Let's exclude the only know use case for now. 👌

rickie (commented):
Rebased and added a commit. 

Nice changes @mohamedsamehsalah 😄, let's get your first BugPattern merged 💪🏻!

Thanks for the review @Venorcis 💪🏻 !

Suggested commit message:
```
Introduce `NestedPublishers` check (#642)
```

rickie (approved):
Rebased and added a commit. 

Nice changes @mohamedsamehsalah 😄, let's get your first BugPattern merged 💪🏻!

Thanks for the review @Venorcis 💪🏻 !

Suggested commit message:
```
Introduce `NestedPublishers` check (#642)
```

rickie (approved):
Rebased and added a commit. 

Nice changes @mohamedsamehsalah 😄, let's get your first BugPattern merged 💪🏻!

Thanks for the review @Venorcis 💪🏻 !

Suggested commit message:
```
Introduce `NestedPublishers` check (#642)
```

rickie:
(Okay, there was an issue with my internet, I reviewed + approved twice all at once ????, sorry for that).

mohamedsamehsalah:
Rebased ✅

rickie (changes requested):
Hey @mohamedsamehsalah , I think something went wrong with rebasing, I don't see my commit anymore 🤯.

```
commit a6546570a756965275a1b147782cef8ece7696b4
Author: Rick Ossendrijver <rick.ossendrijver@gmail.com>
Date:   Wed Jun 7 13:06:56 2023 +0200

    Suggestions

diff --git a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java
index e9566911..6f55bf8c 100644
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java
@@ -10,6 +10,7 @@ import static tech.picnic.errorprone.bugpatterns.util.MoreTypes.type;
 import static tech.picnic.errorprone.bugpatterns.util.MoreTypes.unbound;
 
 import com.google.auto.service.AutoService;
+import com.google.common.collect.Iterables;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.VisitorState;
 import com.google.errorprone.bugpatterns.BugChecker;
@@ -20,10 +21,8 @@ import com.google.errorprone.suppliers.Suppliers;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.MethodInvocationTree;
 import com.sun.tools.javac.code.Type;
-import java.util.function.Function;
 import org.jspecify.annotations.Nullable;
 import org.reactivestreams.Publisher;
-import reactor.core.publisher.Flux;
 
 /** A {@link BugChecker} that flags nesting of {@link Publisher Publishers}. */
 @AutoService(BugChecker.class)
@@ -51,31 +50,18 @@ public final class NestedPublishers extends BugChecker implements MethodInvocati
 
   @Override
-    Type publisherOfPublisherType = PUBLISHER_OF_PUBLISHERS.get(state);
-    Type groupedFluxType = GROUPED_FLUX.get(state);
-    Type treeType = ASTHelpers.getType(tree);
-    if (!isNestedPublisher(state, publisherOfPublisherType, treeType)
-        || isTypeArgumentGroupedFlux(state, groupedFluxType, treeType)) {
+    Type type = ASTHelpers.getType(tree);
+
+    if (!isSubType(type, PUBLISHER_OF_PUBLISHERS.get(state), state)
+        || isSubType(
+            Iterables.getOnlyElement(type.getTypeArguments()), GROUPED_FLUX.get(state), state)) {
       return Description.NO_MATCH;
     }
 
     return describeMatch(tree);
   }
 
-  private static boolean isNestedPublisher(
-      VisitorState state, @Nullable Type publisherOfPublisherType, Type treeType) {
-    return publisherOfPublisherType != null
-        && state.getTypes().isSubtype(treeType, publisherOfPublisherType);
-  }
-
-  /**
-   * Excluding the type when it matches {@code Flux<GroupedFlux<K, V>>} to not flag usages of {@link
-   * Flux#groupBy(Function)}.
-   */
-  private static boolean isTypeArgumentGroupedFlux(
-      VisitorState state, @Nullable Type groupedFluxType, Type treeType) {
-    return groupedFluxType != null
-        && treeType.getTypeArguments().stream()
-            .anyMatch(typez -> ASTHelpers.isSameType(typez, groupedFluxType, state));
+  private static boolean isSubType(Type subType, @Nullable Type type, VisitorState state) {
+    return type != null && state.getTypes().isSubtype(subType, type);
   }
 }
diff --git a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/NestedPublishersTest.java b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/NestedPublishersTest.java
index c4dc6a3e..f1ac337d 100644
--- a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/NestedPublishersTest.java
+++ b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/NestedPublishersTest.java
@@ -14,6 +14,8 @@ final class NestedPublishersTest {
             "",
             "class A {",
             "  void m() {",
+            "    Mono.empty();",
+            "    Mono.just(1);",
             "    Mono.just(Mono.empty());",
@@ -27,6 +29,8 @@ final class NestedPublishersTest {
             "    Mono.just(1).map(Flux::just);",
             "",
+            "    Flux.empty();",
+            "    Flux.just(1);",
             "    Flux.just(Flux.empty());",
```

Would be nice to incorporate them again 😉.

mohamedsamehsalah:
> @rickie  @mohamedsamehsalah , I think something went wrong with rebasing, I don't see my commit anymore 🤯.

Thanks for pointing that out. 🙏 

That was a mistake on my side forgetting to pull origin branch before rebasing 🙈 and now it seems I've lost the commit.

Can you check again ? 👀

mohamedsamehsalah:
```
Error:  Failed to execute goal on project documentation-support: Could not resolve dependencies for project tech.picnic.error-prone-support:documentation-support:jar:0.12.1-SNAPSHOT: Could not find artifact com.github.PicnicSupermarket.error-prone:error_prone_annotation:jar:v2.20.0-picnic-1 in jitpack.io (https://jitpack.io/)
```

Also getting this error locally 😬

rickie (approved):
Sorry for the late re-approve but it looks goood! ✅ 

Thanks @mohamedsamehsalah 🚀 !

Stephan202 (approved):
Added a commit with quite some changes; consider re-reviewing :)

Suggested commit message:
```
Introduce `NestedPublishers` check (#642)

And introduce a few utility methods that simplify `Tree`- and `Type`-based
matching, allowing the existing (and similar) `NestedOptionals` check to be
simplified as well.
```

Venorcis (approved):
Still LGTM 👍🏻

rickie (approved):
Added a tiny Javadoc tweak. Really nice changes, we'll be able to use this in many more future cases 🚀 !

rickie:
Rebased, will merge once 🟢.

PR #644, opened 2023-05-22
Introduce `HasTypeArguments` matcher

oxkitsune:
Introduces a  `Matcher` that matches `{MethodInvocation,NewClass}Tree` instances declared with type arguments.

Should unblock #46


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreMatchers.java
```diff
@@ -32,6 +37,35 @@ public static <T extends AnnotationTree> Matcher<T> hasMetaAnnotation(String ann
...
+  /**
+   * Returns a {@link Matcher} that determines whether a given {@link ExpressionTree} has type
+   * arguments.
+   *
+   * @param <T> The type of tree to match against.
+   * @return A {@link Matcher} that matches trees with type annotations.
+   */
+  public static <T extends ExpressionTree> Matcher<T> hasTypeArguments() {
+    return (tree, state) -> {
+      switch (tree.getKind()) {
```
oxkitsune, line 49: Not sure if there's any other cases I might've missed?

refaster-support/src/test/java/tech/picnic/errorprone/refaster/matchers/HasTypeArgumentsTest.java
```diff
@@ -0,0 +1,83 @@
...
+            "  }",
+            "",
+            "  <E> ImmutableSet<E> positive2() {",
+            "    // BUG: Diagnostic contains:",
+            "    return new ImmutableSet.Builder<E>().build();",
+            "  }",
+            "",
+            "  Foo positive3() {",
+            "    // BUG: Diagnostic contains:",
+            "    return new <List<Object>>Foo(List.of());",
```
oxkitsune, line 49: There's is probably a better example of this

refaster-support/src/test/java/tech/picnic/errorprone/refaster/matchers/HasTypeArgumentsTest.java
```diff
@@ -0,0 +1,83 @@
...
+            "    return new ArrayList<>();",
+            "  }",
+            "",
+            "  Foo negative4() {",
+            "    return new Foo(\"foo\");",
+            "  }",
+            "",
+            "  <E> ImmutableSet<E> positive1() {",
+            "    // BUG: Diagnostic contains:",
+            "    return ImmutableSet.<E>builder().build();",
```
rickie, line 39: We can probably simplify by:

```
"  List<Integer> positive1() {",
"    return new ArrayList<Integer>();",
"  }",
"",
```

(Will apply this tomorrow when I get back to this.

rickie (commented):
Added a small commit. Not done reviewing yet but wanted to flush what I have 😄.

Stephan202 (commented):
I like the idea behind this new `Matcher`, but I wonder how it's meant to be used in #46, because the Refaster template methods modified in that PR don't accept any parameters: the `new X()` / `new X<...>()` constructor invocations against which we want to match are explicitly defined in the method bodies, so there's no place to put the `@Matches`/`@NotMatches` annotations? :thinking:

oxkitsune:
That is something I ran into as well, but haven't had a super good look into it.

PR #649, opened 2023-05-26
Introduce `MongoFullTextSearchQueryUsage` check 

philleonard:
## What?

A simple check which discourages the usage of `Filters#text` API in Mongo queries.

## Why?

We have observed at Picnic that using MongoDB to serve full text search queries can lead to major cluster outages (Mong server OOMs) when query strings are tokenised into even just a handful of tokens, invoking an index scan (`IXSCAN`) for each individual token in a query string. As Mongo does not gracefully handle such queries, and it is not possible to ban the usage of this filter on the (Mongo) server-side, our best effort is to introduce a Bug Check. MongoDB advises against implementing FTS queries on Mongo Atlas, and to use alternative solutions like Atlas Search, and other search engine databases like Elastic Search, or by simply implementing bespoke queries. 

## Implementation notes
- Description "Usage of the Mongo full-text search ($text) queries is discouraged" doesn't actually hold true, perhaps there are other reasons why you would need to use the Filters API to obtain the `Bson` query and not use it in the Java driver - this Bug Check would still flag this. I can add the driver dependencies and match on the execution of a query using the drivers, or we can settle on this being good enough.
- ~Named enum value `MONGO`, like others in `ThirdPartyLibrary` unsure why we don't refer to the specific name of the FQC but likely we can revisit later when we need to rename in order to encompass more than just the Filters API in this case.~ 
- Excluding the `org.mongodb:bson-record-codec` artifact , not required for testing, but contains classes violating compatibility with byte code enforcement limited by the max JDK version (11). The other option is to (see commit) ignore the class via Maven byteCodeEnforcement ( `org.bson.codecs.record.*` ) but I think this would be more disruptive. 
- Perhaps a more elaborate summary is desired to help others better understand the underlying problem, but typically we keep these pretty succinct?

## Suggested commit message

```
Introduce `MongoFullTextSearchQueryUsage` check (#649) 

Discourage the inefficient usage of `mongodb-driver-core`'s `Filters#text` 
for constructing full-text search queries. 
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MongoFullTextSearchQueryUsage.java
```diff
@@ -0,0 +1,50 @@
...
+  /** Instantiates a new {@link MongoFullTextSearchQueryUsage} instance. */
+  public MongoFullTextSearchQueryUsage() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!ThirdPartyLibrary.MONGO.isIntroductionAllowed(state)
+        || !TEXT_FILTER_INVOCATION.matches(tree, state)) {
+      return Description.NO_MATCH;
+    }
+    return describeMatch(tree);
```
rickie, line 48: As we do not have a `SuggestedFix` we don't need to have this `ThirdPartyLibrary.MONGO` part. That is to prevent us from suggesting a Mongo fix while someone might not want / need this. Here we are only flagging something, so we can omit that part :).

rickie (commented):
Nice work @philleonard! Just a quick comment :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MongoFullTextSearchQueryUsage.java
philleonard: Ah, I misread. I thought the intention of `ThirdPartyLibrary` was to list all possibly not present dependencies for bug checks even without a suggested fix. Good catch, will update.

philleonard:
If I understood @Stephan202 correctly from an offline discussion we can downgrade this to a `SUGGESTION` and for our fork of `error-prone` utilise this upcoming feature to https://github.com/PicnicSupermarket/error-prone/pull/33. However this would apply to 25 other `SUGGESTION`s if we were to utilise this flag internally.

Question, would there be an option to add an additional tag from an enum to indicate the Picnic escalation on a per check basis using our fork's processing of the annotation yet still ensuring compatibility with `error-prone`? Something like:

```java
enum PicnicSeverityEscalation {
    WARNING_TO_ERROR,
    SUGGESTION _TO_ERROR,
    SUGGESTION_TO_WARNING
}
```

or simply just

```java
enum PicnicSeverity {
    ERROR,
    WARNING,
    SUGGESTION
}
```

such that for usage in `@BugPattern`s:

```java
@BugPattern(
    summary =
        "Usage of Mongo for full-text search queries via the `$text` operator is discouraged.",
    link = BUG_PATTERNS_BASE_URL + "MongoFullTextSearchQueryUsage",
    linkType = CUSTOM,
    severity = WARNING,
    tags = {PERFORMANCE, PicnicSeverityEscalation.SUGGESTION_TO_WARNING})
...
```

then in our fork modify `BugCheckerInfo` to iterate over the tags and override the `BugCheckerInfo#defaultSeverity` where necessary?

Interested to hear your thoughts!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java
```diff
@@ -32,6 +32,7 @@ public enum ThirdPartyLibrary {
    * @see <a href="https://github.com/google/guava">Guava on GitHub</a>
    */
   GUAVA("com.google.common.collect.ImmutableList"),
+
```
rickie: Let's make sure to not sneak in any unrelated changes 😉.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MongoFullTextSearchQueryUsage.java
```diff
@@ -0,0 +1,48 @@
...
+    summary =
+        "Usage of Mongo for full-text search queries via the `$text` operator is discouraged.",
+    link = BUG_PATTERNS_BASE_URL + "MongoFullTextSearchQueryUsage",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = PERFORMANCE)
+public final class MongoFullTextSearchQueryUsage extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<ExpressionTree> TEXT_FILTER_INVOCATION =
```
rickie: We usually use `_METHOD` instead of `INVOCATION`. In this case I made a suggestion to make it a bit clearer what the matches wants to match. Maybe we can drop the `MONGO_` prefix, but now its inline with the class and order of matching.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MongoFullTextSearchQueryUsage.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @see <a href="https://www.mongodb.com/docs/manual/text-search/">Mongo Text Search</a>
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary =
+        "Usage of Mongo for full-text search queries via the `$text` operator is discouraged.",
+    link = BUG_PATTERNS_BASE_URL + "MongoFullTextSearchQueryUsage",
+    linkType = CUSTOM,
+    severity = WARNING,
```
rickie: Let's change to a `SUGGESTION` as discussed in the other comments.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MongoFullTextSearchQueryUsage.java
```diff
@@ -0,0 +1,48 @@
...
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<ExpressionTree> TEXT_FILTER_INVOCATION =
+      staticMethod().onClass("com.mongodb.client.model.Filters").named("text");
+
+  /** Instantiates a new {@link MongoFullTextSearchQueryUsage} instance. */
+  public MongoFullTextSearchQueryUsage() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!TEXT_FILTER_INVOCATION.matches(tree, state)) {
```
rickie: We can do a `?:` here, that'll be readable still. We have one occurrence of this in the `MoreMatchersTest` :).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MongoFullTextSearchQueryUsageTest.java
```diff
@@ -0,0 +1,28 @@
...
+        .addSourceLines(
+            "A.java",
+            "import com.mongodb.client.model.Filters;",
+            "import com.mongodb.client.model.TextSearchOptions;",
+            "import org.bson.conversions.Bson;",
+            "",
+            "class A {",
+            "",
+            "  void m() {",
+            "    Bson allowed = Filters.eq(\"a\", \"b\");",
```
rickie: We don't need the `Bson allowed =` and other variables :).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MongoFullTextSearchQueryUsageTest.java
```diff
@@ -0,0 +1,28 @@
...
+            "    // BUG: Diagnostic contains:",
+            "    Bson textSearch = Filters.text(\"Some text\");",
+            "    // BUG: Diagnostic contains:",
+            "    Bson textSearch2 = Filters.text(\"Some text\", new TextSearchOptions().caseSensitive(true));",
```
rickie: Added one extra case, just for completeness but might not be that necessary...

rickie (approved):
Rebased and added a commit. Looks nice! 
Only doubt is about the name if it can be shorter, but didn't find a good alternative.

Suggested commit message:
```
Introduce `MongoFullTextSearchQueryUsage` check (#649)
```

rickie:
An extra thought: we could've also used a Checkstyle module here to flag this method. However, I imagine this check being generalized when we find another method we want to flag. 

We can have a check that flags multiple specific methods and have a custom message for every thing that we want to flag. Additionally, we should see if we can provide an optional fix for the methods that do have a simple replacement. So the check will contain methods without possible replacement, or simpler replacements then the ones from the `FluxFlatMapUsage` check.

rickie:
Rebased.

pom.xml
```diff
@@ -443,6 +443,21 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>org.mongodb</groupId>
+                <artifactId>mongodb-driver-core</artifactId>
+                <version>4.9.1</version>
```
Stephan202: Will upgrade.

Stephan202 (approved):
Rebased and added a commit. Indeed we might want to eventually generalize this to a check which flags methods with a custom message.

Tweaked suggested commit message, given the proposed rename:
```
Introduce `MongoDBTextFilterUsage` check (#649)
```

rickie:
Rebased.

rickie:
Rebased, will merge!

PR #691, opened 2023-06-20
Optimized reductant boxing a value

nishantsikarwar:
The lastIndexOf method in Java is typically used to find the last occurrence of a character or substring within a string. 

Issue here : 
https://github.com/PicnicSupermarket/error-prone-support/blob/b81ec973a1eaabfcd2cc75735fc809ea9dcf8ff0/refaster-compiler/src/main/java/tech/picnic/errorprone/refaster/plugin/RefasterRuleCompilerTaskListener.java#L142

Attempting to find the last occurrence of the byte value representing a dot. It may cause unexpected behaviour.

### DIFF
--- a/refaster-compiler/src/main/java/tech/picnic/errorprone/refaster/plugin/RefasterRuleCompilerTaskListener.java
+++ b/refaster-compiler/src/main/java/tech/picnic/errorprone/refaster/plugin/RefasterRuleCompilerTaskListener.java
@@ -139,7 +139,7 @@ final class RefasterRuleCompilerTaskListener implements TaskListener {
 
   private static CharSequence toSimpleFlatName(ClassSymbol symbol) {
     Name flatName = symbol.flatName();
-    int lastDot = flatName.lastIndexOf((byte) '.');
+    int lastDot = flatName.lastIndexOf('.');
     return lastDot < 0 ? flatName : flatName.subSequence(lastDot + 1, flatName.length());
   }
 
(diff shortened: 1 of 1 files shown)


Stephan202:
Hey @nishantsikarwar! This particular code invokes [`com.sun.tools.javac.util.Name#lastIndexOf(byte)`](https://github.com/openjdk/jdk/blob/master/src/jdk.compiler/share/classes/com/sun/tools/javac/util/Name.java#L158) rather than [`java.lang.String#lastIndexOf(int)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#lastIndexOf(int)). The `Name` class does not have a `char`-accepting overload, so the cast is necessary. We could of course first convert the `Name` into a `String`, but there does not seem to be any benefit to that here: I am not aware of any unexpected behavior in this case, because the result of `(byte) '.'` is well-defined.

nishantsikarwar:
Hello @Stephan202 
I Apologise for the confusion,  I meant that the (byte) '.' casting is unnecessary, and using the byte value directly, in this case, is safe.

> Hey @nishantsikarwar! This particular code invokes [`com.sun.tools.javac.util.Name#lastIndexOf(byte)`](https://github.com/openjdk/jdk/blob/master/src/jdk.compiler/share/classes/com/sun/tools/javac/util/Name.java#L158) rather than [`java.lang.String#lastIndexOf(int)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#lastIndexOf(int)). The `Name` class does not have a `char`-accepting overload, so the cast is necessary. We could of course first convert the `Name` into a `String`, but there does not seem to be any benefit to that here: I am not aware of any unexpected behavior in this case, because the result of `(byte) '.'` is well-defined.

Stephan202:
>  I meant that the (byte) '.' casting is unnecessary, and using the byte value directly, in this case, is safe.

@nishantsikarwar the cast is _not_ unnecessary, as the compilation error shows.

nishantsikarwar:
Hi @Stephan202, thanks for the explanation; I got confused while reading the code. This version is correct and works.

Best regards,
Nishant

rickie:
No problem, these things can happen. 😄

PR #700, opened 2023-06-26
Have Renovate split minor and patch upgrades for Maven API

rickie:
Based on @Stephan202's comment https://github.com/PicnicSupermarket/error-prone-support/pull/434#issuecomment-1596054545. 

Tested this locally with the exact same property: 
image

I'm not sure whether we should add entries for the "human readable name" variant of these packages or leave it as is. @Badbond

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -6,6 +6,7 @@
   "packageRules": [
     {
       "matchPackagePatterns": [
+        "Maven API",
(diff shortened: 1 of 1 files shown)


Badbond:
Would it be beneficial to do this globally instead such that we also benefit from it internally? I see we are missing out in said patch release too.

rickie:
Filed an internal PR for updating our global Renovate config.

Issue #716, opened 2023-07-07
`DirectReturn` BugPattern crashes with Lombok's `@Data` annotation


pacbeckh:
### Describe the bug

- [X] I have verified that the issue is reproducible against the latest version
      of the project.
- [X] I have searched through existing issues to verify that this issue is not
      already known.

### Minimal Reproducible Example

```java
package org.example;

import lombok.Data;

@Data
public class AnyClass {

  private String anything;

}
```

A full example is available at https://github.com/pacbeckh/error-prone-lombok-issue,  run `mvn clean install` 

<details>
<summary>Logs</summary>

```sh
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project errorproneissue: Compilation failure
[ERROR] /<redacted>/error-prone-lombok-issue/src/main/java/org/example/AnyClass.java:[5,1] An unhandled exception was thrown by the Error Prone static analysis plugin.
[ERROR]      Please report this at https://github.com/google/error-prone/issues/new and include the following:
[ERROR]   
[ERROR]      error-prone version: 2.20.0
[ERROR]      BugPattern: DirectReturn
[ERROR]      Stack Trace:
[ERROR]      java.lang.IllegalArgumentException: Replacement{range=[43..48), replaceWith=} conflicts with existing replacement Replacement{range=[43..48), replaceWith=return @Data;}
[ERROR]         at com.google.errorprone.fixes.Replacements.add(Replacements.java:166)
[ERROR]         at com.google.errorprone.fixes.SuggestedFix.getReplacements(SuggestedFix.java:93)
[ERROR]         at com.google.errorprone.fixes.AppliedFix$Applier.apply(AppliedFix.java:69)
[ERROR]         at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1134)
[ERROR]         at com.google.errorprone.JavacErrorDescriptionListener.lambda$new$1(JavacErrorDescriptionListener.java:94)
[ERROR]         at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
[ERROR]         at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177)
[ERROR]         at java.base/java.util.Collections$2.tryAdvance(Collections.java:4747)
[ERROR]         at java.base/java.util.Collections$2.forEachRemaining(Collections.java:4755)
[...]

Issue #557, opened 2023-03-31
False positives for Lombok's @Data annotation


ksiczek:
### Describe the bug

- [x] I have verified that the issue is reproducible against the latest version
      of the project.
- [x] I have searched through existing issues to verify that this issue is not
      already known.

### Minimal Reproducible Example

Given below class 

```java
package com.loomis.pay.contract.api.model;

import com.fasterxml.jackson.annotation.JsonProperty;
import lombok.Data;

@Data
public class ErrorProneTestObject {

    @JsonProperty("event")
    private String event;

}

```

we get unclear recommendations saying that we might replace `@Data` with `@Data` (see Logs section). It seems it happens only for Lombok `@Data`, or `@Setter`:
``` java
@Setter
public class ErrorProneTestObject {

    @JsonProperty("event")
    private String event;

    public String getEvent() {
        return event;
    }

} 
```

<details>
<summary>Logs</summary>

```sh
.../ErrorProneTestObject.java:6: Note: [CanonicalAnnotationSyntax] Omit redundant syntax from annotation declarations
@Data
^
    (see https://error-prone.picnic.tech/bugpatterns/CanonicalAnnotationSyntax)
  Did you mean '@Data'?
```
</details>

I tried several different combinations of the annotations and configurations and it seems that:
- it always happens if you use the ones I specified above
- it stops happening if I remove one of them
- it also stops happening if I remove `JsonProperty`'s parameter

### Expected behavior
I'm pretty sure that the error message does not help much so it is either a bug in the analysis or in the message. Does it sound familiar to you guys? 

### Setup

- MacOS Ventura
- openjdk version "19.0.2" 2023-01-17
- Error Prone version `2.18.0`
- Error Prone Support version `0.8.0`

### Additional context
I just started using `error-prone` and it might be a good opportunity for me to see how custom rules work. If you consider reported behavior a bug I think I could dive deeper into it ☺️

rickie:
Hi @ksiczek, thanks for taking the time to report an issue!

This sure looks like a bug and should be fixed. The thing is that Lombok probably generates extra parentheses "under the hood" that are flagged by this check. 

> it also stops happening if I remove JsonProperty's parameter

This is interesting and didn't expect that though 🤔.

[...]

PR #728, opened 2023-07-23
`DirectReturn` bugpattern clash with lombok's @Data

mohamedsamehsalah:
Suggested commit message:
```
Have `DirectReturn` ignore classes annotated with Lombok's `@Data` (#728)

While there, introduce the `MoreMatchers#HAS_LOMBOK_DATA` matcher.
```

**TL;DR `DirectReturn` bug pattern (If not all bug patterns that use replacements) detects bugs for generated resources, but replaces the suggested fixes in the original source code.**

Notes 📝 
This is the generated class from the project associated with this ticket. (#716 )
```
public class AnyClass {
    private String anything;

    @java.lang.SuppressWarnings(value = "all")
    public AnyClass() {
        super();
    }

    @java.lang.SuppressWarnings(value = "all")
    public String getAnything() {
        return this.anything;
    }

    @java.lang.SuppressWarnings(value = "all")
    public void setAnything(final String anything) {
        this.anything = anything;
    }

    @java.lang.Override
    @java.lang.SuppressWarnings(value = "all")
    public boolean equals(final java.lang.Object o) {
        if (o == this) return true;
        if (!(o instanceof AnyClass)) return false;
        final AnyClass other = (AnyClass)o;
        if (!other.canEqual((java.lang.Object)this)) return false;
        final java.lang.Object this$anything = this.getAnything();
        final java.lang.Object other$anything = other.getAnything();
        if (this$anything == null ? other$anything != null : !this$anything.equals(other$anything)) return false;
        return true;
    }

    @java.lang.SuppressWarnings(value = "all")
    protected boolean canEqual(final java.lang.Object other) {
        return other instanceof AnyClass;
    }

    @java.lang.Override
    @java.lang.SuppressWarnings(value = "all")
    public int hashCode() {
        final int PRIME = 59;
        int result = 1;
        final java.lang.Object $anything = this.getAnything();
        result = result * PRIME + ($anything == null ? 43 : $anything.hashCode());
        return result;
    }

    @java.lang.Override
    @java.lang.SuppressWarnings(value = "all")
    public java.lang.String toString() {
        return "AnyClass(anything=" + this.getAnything() + ")";
    }
}
```

The issue seems to be when `DirectReturn` tries to optimize line 
```
        result = result * PRIME + ($anything == null ? 43 : $anything.hashCode());
```
in the `hashCode` method.

Due to the (helper?) methods used inside `DirectReturn` error prone parses and tries to replace the "source code" (The original annotated class) and not the generated resource.

**Fix:**

1. Ignore files that are using `@Data` annotation. (See [this](https://github.com/PicnicSupermarket/error-prone-support/pull/728#pullrequestreview-1544783907))
~~2. Force `DirectReturn` to replace the generated sources instead of source code, but:
  a. Should Error prone change generated files or files of other libraries ?
  b. I dont know how to do that (Any references would be helpful 😄 )~~

### DIFF
--- a/error-prone-contrib/pom.xml
+++ b/error-prone-contrib/pom.xml
@@ -182,6 +182,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.reactivestreams</groupId>
             <artifactId>reactive-streams</artifactId>
             <scope>provided</scope>
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
@@ -12,7 +12,7 @@ import static com.google.errorprone.matchers.Matchers.returnStatement;
 import static com.google.errorprone.matchers.Matchers.staticMethod;
 import static com.google.errorprone.matchers.Matchers.toType;
 import static tech.picnic.errorprone.bugpatterns.util.Documentation.BUG_PATTERNS_BASE_URL;
-import static tech.picnic.errorprone.bugpatterns.util.MoreMatchers.isClassAnnotatedWithLombokData;
+import static tech.picnic.errorprone.bugpatterns.util.MoreMatchers.HAS_LOMBOK_DATA;
 
 import com.google.auto.service.AutoService;
 import com.google.common.collect.Streams;
@@ -62,8 +62,6 @@ public final class DirectReturn extends BugChecker implements BlockTreeMatcher {
       allOf(
           not(toType(MethodInvocationTree.class, argument(0, isSameType(Class.class.getName())))),
           staticMethod().onClass("org.mockito.Mockito").namedAnyOf("mock", "spy"));
-  private static final Matcher<ClassTree> IS_CLASS_ANNOTATED_WITH_LOMBOK_DATA =
-      isClassAnnotatedWithLombokData();
 
   /** Instantiates a new {@link DirectReturn} instance. */
   public DirectReturn() {}
@@ -72,8 +70,7 @@ public final class DirectReturn extends BugChecker implements BlockTreeMatcher {
   public Description matchBlock(BlockTree tree, VisitorState state) {
     List<? extends StatementTree> statements = tree.getStatements();
     ClassTree enclosingNode = ASTHelpers.findEnclosingNode(state.getPath(), ClassTree.class);
-    if (statements.size() < 2
-        || IS_CLASS_ANNOTATED_WITH_LOMBOK_DATA.matches(enclosingNode, state)) {
+    if (statements.size() < 2 || HAS_LOMBOK_DATA.matches(enclosingNode, state)) {
       return Description.NO_MATCH;
     }
 
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
@@ -12,6 +12,7 @@ import static com.google.errorprone.matchers.Matchers.returnStatement;
 import static com.google.errorprone.matchers.Matchers.staticMethod;
 import static com.google.errorprone.matchers.Matchers.toType;
 import static tech.picnic.errorprone.bugpatterns.util.Documentation.BUG_PATTERNS_BASE_URL;
+import static tech.picnic.errorprone.bugpatterns.util.MoreMatchers.isClassAnnotatedWithLombokData;
 
 import com.google.auto.service.AutoService;
 import com.google.common.collect.Streams;
@@ -25,6 +26,7 @@ import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.AssignmentTree;
 import com.sun.source.tree.BlockTree;
+import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.ExpressionStatementTree;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.IdentifierTree;
@@ -60,6 +62,8 @@ public final class DirectReturn extends BugChecker implements BlockTreeMatcher {
       allOf(
           not(toType(MethodInvocationTree.class, argument(0, isSameType(Class.class.getName())))),
           staticMethod().onClass("org.mockito.Mockito").namedAnyOf("mock", "spy"));
+  private static final Matcher<ClassTree> IS_CLASS_ANNOTATED_WITH_LOMBOK_DATA =
+      isClassAnnotatedWithLombokData();
 
   /** Instantiates a new {@link DirectReturn} instance. */
   public DirectReturn() {}
@@ -67,7 +71,9 @@ public final class DirectReturn extends BugChecker implements BlockTreeMatcher {
   @Override
   public Description matchBlock(BlockTree tree, VisitorState state) {
     List<? extends StatementTree> statements = tree.getStatements();
-    if (statements.size() < 2) {
+    ClassTree enclosingNode = ASTHelpers.findEnclosingNode(state.getPath(), ClassTree.class);
+    if (statements.size() < 2
+        || IS_CLASS_ANNOTATED_WITH_LOMBOK_DATA.matches(enclosingNode, state)) {
       return Description.NO_MATCH;
     }
 
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
@@ -69,8 +69,8 @@ public final class DirectReturn extends BugChecker implements BlockTreeMatcher {
   @Override
   public Description matchBlock(BlockTree tree, VisitorState state) {
     List<? extends StatementTree> statements = tree.getStatements();
-    ClassTree enclosingNode = ASTHelpers.findEnclosingNode(state.getPath(), ClassTree.class);
-    if (statements.size() < 2 || HAS_LOMBOK_DATA.matches(enclosingNode, state)) {
+    ClassTree enclosingClass = ASTHelpers.findEnclosingNode(state.getPath(), ClassTree.class);
+    if (statements.size() < 2 || HAS_LOMBOK_DATA.matches(enclosingClass, state)) {
       return Description.NO_MATCH;
     }
 
--- a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
+++ b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
@@ -225,14 +225,16 @@ final class DirectReturnTest {
   }
 
   @Test
-  void excludeClassesAnnotatedWithLombokData() {
+  void ignoreClassesAnnotatedWithLombokData() {
     BugCheckerRefactoringTestHelper.newInstance(DirectReturn.class, getClass())
+        .addInputLines("Data.java", "package lombok;", "public @interface Data {}")
+        .expectUnchanged()
         .addInputLines(
             "A.java",
             "import lombok.Data;",
             "",
             "@Data",
-            "public class A {",
+            "class A {",
             "  private String field;",
             "}")
         .addOutputLines(
@@ -240,7 +242,7 @@ final class DirectReturnTest {
             "import lombok.Data;",
             "",
             "@Data",
-            "public class A {",
+            "class A {",
             "  private String field;",
             "}")
         .doTest(TestMode.TEXT_MATCH);
--- a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
+++ b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
@@ -226,10 +226,8 @@ final class DirectReturnTest {
 
   @Test
   void ignoreClassesAnnotatedWithLombokData() {
-    BugCheckerRefactoringTestHelper.newInstance(DirectReturn.class, getClass())
-        .addInputLines("Data.java", "package lombok;", "", "public @interface Data {}")
-        .expectUnchanged()
-        .addInputLines(
+    CompilationTestHelper.newInstance(DirectReturn.class, getClass())
+        .addSourceLines(
             "A.java",
             "import lombok.Data;",
             "",
@@ -237,14 +235,7 @@ final class DirectReturnTest {
             "class A {",
             "  private String field;",
             "}")
-        .addOutputLines(
-            "A.java",
-            "import lombok.Data;",
-            "",
-            "@Data",
-            "class A {",
-            "  private String field;",
-            "}")
-        .doTest(TestMode.TEXT_MATCH);
+        .expectNoDiagnostics()
+        .doTest();
   }
 }
--- a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
+++ b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
@@ -223,4 +223,26 @@ final class DirectReturnTest {
             "}")
         .doTest(TestMode.TEXT_MATCH);
   }
+
+  @Test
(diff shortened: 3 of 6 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
```diff
@@ -91,12 +97,26 @@ public Description matchBlock(BlockTree tree, VisitorState state) {
                     SuggestedFix.builder()
                         .replace(
                             precedingStatement,
-                            String.format("return %s;", SourceCode.treeToString(resultExpr, state)))
-                        .delete(finalStatement)
+//                            String.format("return %s;", SourceCode.treeToString(resultExpr, state)))
+                            String.format("return %s;", SourceCode.treeToString(resultExpr)))
+                            //Same case for trying to delete wrongly assumed "source code".
+                            //Note 2: Seems we can only use ONE fix because of the way the errorprone replaces "source code"
```
mohamedsamehsalah, line 103:
```suggestion
                            //Note 2: Seems we can only use ONE fix because of the way the errorprone replaces "source code" where we have two replacements in the same start/end position causing the "conflicting replacement" exception.
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
```diff
@@ -223,4 +225,26 @@ void replacement() {
...
+                      "private String anything;",
+                    "}")
+            .addOutputLines(
+                    "AnyClass.java",
+                    "import lombok.Data;",
+                    "",
+                    "@Data",
+                    "public class AnyClass {",
+                      "private String anything;",
+                    "}")
```
mohamedsamehsalah, line 247: For reference, here is the actual output 😢 
```suggestion
    import lombok.Data;
    
    return result * PRIME + ($anything == null ? 43 : $anything.hashCode());
    public class AnyClass {
    private String anything;
    }
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/SourceCode.java
```diff
@@ -38,10 +38,42 @@ private SourceCode() {}
...
+      return source.subSequence(start, end).toString();
+     */
     String src = state.getSourceForNode(tree);
     return src != null ? src : tree.toString();
   }
 
+
+
+  /**
+   * TODO: Call this conditionally from DirectReturn ??
```
mohamedsamehsalah, line 70: This will not help since the correctly replaced line is going to be replaced in the wrong block (read source code instead of generated resurces)

error-prone-contrib/pom.xml
```diff
@@ -216,6 +216,12 @@
             <artifactId>testng</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>${version.lombok}</version>
```
rickie, line 222: Small tip; if we only have one version declaration as we have here, we don't need to extract it to a `<version.lombok>`.

rickie (commented):
Thanks again Mo for picking this up and the research 🌟 💯 🥇! 

I think the best way (and easiest way) for now would be to exclude classes from analysis where Lombok is involved indeed.
You can check [this example](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ScheduledTransactionTrace.java#L45-L46) where we don't want to match specific `Tree`s that are not annotated with `Scheduled`. So we need to have this for all the Lombok annotatations that can have an impact on this check. 
Please note that we need to use the fully qualified class name (FQCN) to match against. 

The reason I'm all in for excluding Lombok classes from analysis is that it is hard to make a correct suggestion and we should not want to go down that road.

---

While we are at it, maybe it's nice to introduce a new `Matcher` for this to `MoreMatchers.java` that we can reuse in the `CanonicalAnnotationSyntax` check. That way we can fix both tickets at once, as we could introduce a usage for this ticket as well https://github.com/PicnicSupermarket/error-prone-support/issues/557. 

Ideally we do not have to add the Lombok annotation to EPS (in the end it is probably inevatable though). We can use the FCQN and that should be enough.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
```diff
@@ -223,4 +223,26 @@ void replacement() {
             "}")
         .doTest(TestMode.TEXT_MATCH);
   }
+
+  @Test
+  void excludeClassesAnnotatedWithLombokData() {
```
rickie:
```suggestion
  void ignoreClassesAnnotatedWithLombokData() {
```

error-prone-contrib/pom.xml
```diff
@@ -176,6 +176,12 @@
             <artifactId>mockito-core</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
rickie: I think we can work around having to add this dependency.

rickie (commented):
Made some suggestions 😄, please take a look :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
```diff
@@ -67,7 +69,8 @@ public DirectReturn() {}
   @Override
   public Description matchBlock(BlockTree tree, VisitorState state) {
     List<? extends StatementTree> statements = tree.getStatements();
-    if (statements.size() < 2) {
+    ClassTree enclosingNode = ASTHelpers.findEnclosingNode(state.getPath(), ClassTree.class);
```
rickie: One thing I considered is implementing `matchClass` and do this check there, but I think this makes more sense. CC @Stephan202.

rickie (approved):
Thanks for picking this up @mohamedsamehsalah !

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
```diff
@@ -67,7 +69,8 @@ public DirectReturn() {}
   @Override
   public Description matchBlock(BlockTree tree, VisitorState state) {
     List<? extends StatementTree> statements = tree.getStatements();
-    if (statements.size() < 2) {
+    ClassTree enclosingClass = ASTHelpers.findEnclosingNode(state.getPath(), ClassTree.class);
+    if (statements.size() < 2 || HAS_LOMBOK_DATA.matches(enclosingClass, state)) {
```
Stephan202, line 73: Didn't check the commit history, but I see a comment about having Lombok as a dependency. Personally I think that we _should_ have a proper test that triggers the regression before the fix. Did we check whether this is possible with (perhaps with `setArgs`?) without enabling Lombok for the main build?

(Another (less preferred) option is to introduce a separate Maven module with a number of fixtures like [Immutables does](https://github.com/immutables/immutables/tree/master/value-fixture/src/org/immutables/fixture). That would be a rather indirect check, though, because it'd rely on the self-check build.)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/MoreMatchersTest.java
```diff
@@ -47,9 +50,33 @@ void hasMetaAnnotation() {
...
+            "}")
+        .addSourceLines("D.java", "public class D {}")
+        .doTest();
+  }
+
+  /** A {@link BugChecker} that delegates to {@link MoreMatchers#hasMetaAnnotation(String)}. */
   @BugPattern(summary = "Interacts with `MoreMatchers` for testing purposes", severity = ERROR)
-  public static final class TestMatcher extends BugChecker implements AnnotationTreeMatcher {
+  public static final class TestTemplateMatcher extends BugChecker
+      implements AnnotationTreeMatcher {
```
Stephan202, line 79: In #642 I renamed this type to `HasMetaAnnotationTestChecker`, in line with approaches taken elsewhere.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/MoreMatchersTest.java
```diff
@@ -47,9 +50,33 @@ void hasMetaAnnotation() {
         .doTest();
   }
 
-  /** A {@link BugChecker} that delegates to {@link MoreMatchers#hasMetaAnnotation(String)} . */
+  @Test
+  void hasLombokDataAnnotation() {
```
Stephan202, line 54: Given the the constant field is listed first, I'd move this method up (and likewise for the test checker definition below).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/MoreMatchersTest.java
```diff
@@ -59,4 +86,16 @@ public Description matchAnnotation(AnnotationTree tree, VisitorState state) {
       return DELEGATE.matches(tree, state) ? describeMatch(tree) : Description.NO_MATCH;
     }
   }
+
+  /** A {@link BugChecker} that delegates to {@link MoreMatchers#HAS_LOMBOK_DATA}. */
+  @BugPattern(summary = "Interacts with `MoreMatchers` for testing purposes", severity = ERROR)
+  public static final class LombokDataAnnotationMatcher extends BugChecker
```
Stephan202, line 92: This would then become `HasLombokDataTestChecker`.

Stephan202 (commented):
I'll need to find time later (when TBD) for a closer look at this PR.

Checking @pacbeckh's reproduction case with a debugger, it seems that all involved `Tree` instances have `@Data` as their source representation, with the full compilation unit source matching the original source code. So I'd like to better understand whether other checks may also be impacted, and thus whether generalizations of this approach apply. (For example, perhaps we should more generally identify and suppress overlapping suggestions. Not sure.)

rickie:
Added a commit and applied the suggestions from @Stephan202.

> Didn't check the commit history, but I see a comment about having Lombok as a dependency. Personally I think that we should have a proper test that triggers the regression before the fix. Did we check whether this is possible with (perhaps with setArgs?) without enabling Lombok for the main build?

Yes this makes sense. For now I was inclined to go with what we have right now. A more generic solution for Lombok is of course nice, but having fixes for the checks that are impacted right now is something I valued more for now. 

> Checking @pacbeckh's reproduction case with a debugger, it seems that all involved Tree instances have @Data as their source representation, with the full compilation unit source matching the original source code. So I'd like to better understand whether other checks may also be impacted, and thus whether generalizations of this approach apply. (For example, perhaps we should more generally identify and suppress overlapping suggestions. Not sure.)

I would like to propose to defer looking into such a bigger solution for now. Don't get me wrong, it would be good to spent time on this, but while having limited time right now there are some other EPS things I like to focus on more :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/DirectReturn.java
rickie: Sorry I don't fully get what you mean with the first part.

I created a failing test with this diff:
```
diff --git a/error-prone-contrib/pom.xml b/error-prone-contrib/pom.xml
index cd4016df..d04e1b5e 100644
--- a/error-prone-contrib/pom.xml
+++ b/error-prone-contrib/pom.xml
@@ -181,6 +181,11 @@
             <artifactId>mongodb-driver-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <scope>test</scope>
+        </dependency>
         <dependency>
             <groupId>org.reactivestreams</groupId>
             <artifactId>reactive-streams</artifactId>
diff --git a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
index 42e18894..d5ef8895 100644
--- a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
+++ b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
@@ -223,4 +223,26 @@ final class DirectReturnTest {
             "}")
         .doTest(TestMode.TEXT_MATCH);
   }
+
+  @Test
+  void exclude() {
+    BugCheckerRefactoringTestHelper.newInstance(DirectReturn.class, getClass())
+        .addInputLines(
+            "A.java",
+            "import lombok.Data;",
+            "",
+            "@Data",
+            "public class A {",
+            "  private String field;",
+            "}")
+        .addOutputLines(
+            "A.java",
+            "import lombok.Data;",
+            "",
+            "@Data",
+            "public class A {",
+            "  private String field;",
+            "}")
+  }
 }
diff --git a/pom.xml b/pom.xml
index a474e356..d08c940a 100644
--- a/pom.xml
+++ b/pom.xml
@@ -456,6 +456,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <version>1.18.28</version>
+            </dependency>
```

Do you mean having a failing test like this?

Stephan202: If that already triggers the issue (I'm surprised that Lombok is auto-enabled; it might not be when the tests are executed with JDK 11 :eyes:): yes, something like that!

rickie: Yes that triggers the issue. It's funny because this is basically what @mohamedsamehsalah proposed and I changed it. So I'll add a commit to restore that state 😄.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
```diff
@@ -223,4 +223,19 @@ void replacement() {
...
+    CompilationTestHelper.newInstance(DirectReturn.class, getClass())
+        .addSourceLines(
+            "A.java",
+            "import lombok.Data;",
+            "",
+            "@Data",
+            "class A {",
+            "  private String field;",
+            "}")
+        .expectNoDiagnostics()
```
rickie, line 238: We can drop this, but I think it makes it quite clear though 🤔. Will drop for now.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/DirectReturnTest.java
```diff
@@ -223,4 +223,18 @@ void replacement() {
...
+        .addSourceLines(
+            "A.java",
+            "import lombok.Data;",
+            "",
+            "@Data",
+            "class A {",
+            "  private String field;",
+            "}")
+        .doTest();
+  }
```
Stephan202, line 239: If I revert the fix then this test doesn't fail. Contrary to what's claimed further up, as best as I can tell Lombok is _not_ executed here. Seems we need to explicitly enable the annotation processor:
```
.setArgs("-processor", "lombok.launch.AnnotationProcessorHider$AnnotationProcessor")
```

Stephan202 (commented):
Rebased and added a commit. Looking at the generated AST, I notice that the methods generated by Lombok are annotated with `@java.lang.SuppressWarnings(value = "all")`. Suppressing based on _that_ seems more robust (and also covers e.g. Immutables-generated code).

Stephan202:
> Looking at the generated AST, I notice that the methods generated by Lombok are annotated with `@java.lang.SuppressWarnings(value = "all")`. Suppressing based on _that_ seems more robust (and also covers e.g. Immutables-generated code).

We might be able to handle this upstream; see https://github.com/google/error-prone/issues/4065#issuecomment-1710540845 and subsequent discussion. I'll have a look at that in the coming days.

Stephan202:
Opened a PR upstream: google/error-prone#4076.

mohamedsamehsalah:
> Opened a PR upstream: https://github.com/google/error-prone/pull/4076.

Thanks a lot for taking care @Stephan202 🙏 

Shall we close this ticket or ? 🤔

Stephan202:
> Shall we close this PR / ticket or ? 🤔

I verified that the unit test added to `DirectReturnTest` fails when running without the upstream PR, and passes with it. So indeed we can likely close this ticket. But since the upstream PR hasn't been merged/released yet, perhaps it makes sense to keep this open as a fallback. (Alternatively we can also reopen later.)

Stephan202:
NB: Either way: _thanks_ for the effort on this one!

Stephan202:
The upstream PR has been merged; will close this PR. But again: thanks for the effort :muscle:

PR #733, opened 2023-07-28
Introduce `EmptyMonoZip` check

DimaLegeza:
The rule that verifies the codebase on the occurrences of the `a.zip(b)` (where `a` or `b` are `Mono<Void>`) as well as for the following construct `Mono.zip(a,b,c)` (where any of params is `Mono<Void>`). 
Such constructs are allowed by Reactor but don't behave as expected as the Reactor chain stops upon the first empty publisher encountered.

Explicit usage of empty publishers (`Mono<Void>` and `Mono.empty`) within the Reactor zipping logic will be reported as a bug. No replacement will be provided as the original implementation should be checked and fixed by the developer 

⚠️N.B. >> 
Usage of `Mono.empty().zipWith` won't be reported. But I think it's logical enough not to create such a construct


DimaLegeza:
Right now we have observed a strange behaviour of the matcher as it also matches valid Reactor usages.
For some reason `Mono<Integer>` (for example) is treated as a subtype of `Mono<Void>`

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,130 @@
...
+
+    // Method references are expected to occur very infrequently; generating both variants of
+    // suggested fixes is not worth the trouble.
+    return describeMatch(tree);
+  }
+
+  private static Matcher<MethodInvocationTree> hasArgumentOfType(Supplier<Type> type) {
+    return (tree, state) ->
+        tree.getArguments().stream()
+            .anyMatch(arg -> ASTHelpers.isSubtype(ASTHelpers.getType(arg), type.get(state), state));
```
oxkitsune, line 128: This will match both `Mono<Void>` and `Mono<Integer>`, because `ASTHelpers#isSubType` uses the **type erased** type, which would be `Mono` and `Mono` in this case. 

Comparing the full types can be done using:
```suggestion
            .anyMatch(arg -> state.getTypes().isSameType(ASTHelpers.getType(arg), type.get(state)));
```

DimaLegeza: @rickie promised to take a look 🙂

DimaLegeza: But let me try the suggested approach.

oxkitsune:
```java
Mono.zip(Mono.empty(), Mono.empty())
```
does not get matched, as the type of `Mono#empty()` is `Mono<java.lang.Object>`. 
I discussed it offline with @rickie and we think we can safely not flag this case. Rick said he'll take another look :pray:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,130 @@
...
+ * <p>{@link Mono#zip(Mono, Mono)} and {@link Mono#zipWith(Mono)} perform incorrectly upon retrieval
+ * of the empty publisher and prematurely terminates the reactive chain from the execution. In most
+ * cases this is not the desired behaviour and {@link Mono#concatWith(Publisher)} or {@link
+ * Mono#then(Mono)} should be preferred, as it produces consistent results and performs in the
+ * predictable manner.
+ *
+ * <p>NB: Mono&lt;?>#zipWith(Mono&lt;Void>) is allowed be the Reactor API, but it is an incorrect
+ * usage of the API. It will be replaced with Mono&lt;?>#concatWith(Mono&lt;Void>) but it will lead
+ * to the compilation errors, indicating that the logic under the hood should be revisited and
+ * fixed.
```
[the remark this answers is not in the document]
rickie: Discussed with @DimaLegeza offline. Flagging cases of which we can be quite sure that it indicates a problem sound like a nice plan.  

We were wondering if you have a solution / idea about this problem https://github.com/PicnicSupermarket/error-prone-support/pull/733#discussion_r1277458027 @Stephan202. 

@DimaLegeza will play around with trying to get the logic for flagging the likely error cases right. After that he'll push the changes for a proposal for the implementation of this check.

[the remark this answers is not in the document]
Stephan202: Nice to see this PR resurrected!

> We were wondering if you have a solution / idea about this problem [#733 (comment)](https://github.com/PicnicSupermarket/error-prone-support/pull/733#discussion_r1277458027) @Stephan202.

@oxkitsune's initial suggestion there is correct. The fact that it's [not enough](https://github.com/PicnicSupermarket/error-prone-support/pull/733#discussion_r1277498380), shows that type matching won't suffice: one would need to look at the "shape" of the expressions, e.g. by explicitly looking for `Mono.empty()`.

But... since this PR was opened we introduced the [`IsEmpty`](https://github.com/PicnicSupermarket/error-prone-support/blob/1d25b78c4c6bbc426b32766aaa373c1361a45a6e/refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsEmpty.java) matcher, which already [recognizes](https://github.com/PicnicSupermarket/error-prone-support/blob/1d25b78c4c6bbc426b32766aaa373c1361a45a6e/refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsEmpty.java#L98-L104) `Mono.isEmpty()`, and it has a [TODO](https://github.com/PicnicSupermarket/error-prone-support/blob/1d25b78c4c6bbc426b32766aaa373c1361a45a6e/refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers/IsEmpty.java#L52-L53) for supporting e.g. `Mono<Void>`. So I see two ways forward:
1. Just write a few Refaster rules for the most common cases (extending `IsEmpty` if necessary).
2. Use the `IsEmpty` logic as a foundation for a `BugChecker`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+          toType(MethodInvocationTree.class, hasGenericArgumentOfExactType(MONO_VOID_TYPE)));
+
+  /** Instantiates a new {@link MonoZipOfMonoVoidUsage} instance. */
+  public MonoZipOfMonoVoidUsage() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    boolean dynamicMono = MONO_ZIP_AND_WITH.matches(tree, state);
+    boolean staticMono = STATIC_MONO_ZIP.matches(tree, state);
+    if (!dynamicMono && !staticMono) {
```
mohamedsamehsalah, line 86: Can be inlined.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+      return Description.NO_MATCH;
+    }
+
+    return buildDescription(tree)
+        .setMessage(
```
mohamedsamehsalah, line 91: IIUC, we use this method to set the message when we don't supply a message inside the `@BugPattern` annotation.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+            "`Mono#zip` and `Mono#zipWith` should not be executed against empty publisher; "
+                + "remove it or suppress this warning and add a comment explaining its purpose")
+        .addFix(SuggestedFixes.addSuppressWarnings(state, canonicalName()))
+        .build();
```
mohamedsamehsalah, line 95: I think the convention is to use a `describeMatch` method with a "built" `SuggestedFixBuilder`.

```suggestion
    return describeMatch(tree, fix);
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+
+  // On instance mono.zipWith, at least one element should match empty in order to proceed.
+  private static final Matcher<ExpressionTree> MONO_ZIP_AND_WITH =
+      anyOf(
+          allOf(
+              instanceMethod().onDescendantOf(MONO).named("zipWith"),
+              toType(MethodInvocationTree.class, hasGenericArgumentOfExactType(MONO_VOID_TYPE))),
+          allOf(
+              instanceMethod().onDescendantOf(MONO).named("zipWith"),
+              toType(MethodInvocationTree.class, staticMethod().onClass(MONO).named("empty"))),
```
mohamedsamehsalah, line 70: The tests still pass if I remove this matcher 🤔 
```suggestion

```

mohamedsamehsalah (commented):
1st pass review 💪 

Good stuff, left some comments that I also don't know the answer to. (Maybe @rickie can help here 👀 )

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
rickie: We can probably even combine the `MONO_ZIP_AND_WITH` and `STATIC_MONO_ZIP` into one Matcher! I see they are both used only in this place.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
rickie: We don't need to use `setMessage` indeed, but @mohamedsamehsalah note the `SuggestedFix` provides a `@SuppressWarnings` :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+   * <p>- or we have a method invocation, like {@code Mono.just(Object)} or {@code Mono.empty()},
+   * for which we can also infer type.
+   *
+   * <p>Similarly, we can infer the matching type
+   */
+  private static boolean hasSameGenericType(
+      Type genericArgumentType, Type genericDesiredType, VisitorState state) {
+    Type argumentType = Iterables.getFirst(genericArgumentType.allparams(), Type.noType);
+    Type requiredType = Iterables.getOnlyElement(genericDesiredType.allparams());
+    return isSameType(argumentType, requiredType, state);
```
rickie, line 141: Let's not statically import this one :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /**
+   * We need to extract real types from the generics because {@link ASTHelpers} cannot distinguish
+   * {@code Mono<Integer>} and {@code Mono<Void>} and reports those being the same.
+   *
+   * <p>In case of {@code Mono}, we can infer the real type out of the parameters of the invocation
+   * ({@link MethodInvocationTree#getArguments()}):
+   *
+   * <p>- either we have explicit variable declared and the provided type which will be inferred,
```
rickie, line 130: Let's create a summation like this if necessary:
```
 * <ul>
 *   <li>item
 *   <li>item
 * </ul>
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+            "`Mono#zip` and `Mono#zipWith` should not be executed against empty publisher; "
+                + "remove it or suppress this warning and add a comment explaining its purpose")
+        .addFix(SuggestedFixes.addSuppressWarnings(state, canonicalName()))
+        .build();
+  }
+
+  private static Matcher<ExpressionTree> onClassWithMethodName(
+      Supplier<Type> genericDesiredType, String methodName) {
+    return (tree, state) -> {
+      JCTree.JCExpression methodSelect = ((JCTree.JCMethodInvocation) tree).getMethodSelect();
```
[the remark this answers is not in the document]
rickie:
```suggestion
 MemberSelectTree ms = (MemberSelectTree) tree.getMethodSelect();
```
Something like that? Didn't check out so I'm not sure.

If we use that, the rest of the method probably looks a bit nicer.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+            + "please revisit the parameters used and make sure to supply correct publishers instead",
+    link = BUG_PATTERNS_BASE_URL + "MonoZipOfMonoVoidUsage",
+    linkType = CUSTOM,
+    severity = ERROR,
+    tags = LIKELY_ERROR)
+public final class MonoZipOfMonoVoidUsage extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Supplier<Type> MONO = type("reactor.core.publisher.Mono");
+  // In fact, we use `Mono<Void>` everywhere in codebases instead of `Mono<Object>` to represent
```
rickie, line 57: We usually do not use these comments. If we do, we should use `/* */` instead of `//`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * A {@link BugChecker} that flags usages of {@code Mono.zip(Mono, Mono)} and {@code
+ * Mono.zipWith(Mono)} with {@code Mono.empty()} parameters.
+ *
+ * <p>{@code Mono.zip(Mono, Mono)} and {@code Mono.zipWith(Mono)} perform incorrectly upon retrieval
+ * of the empty publisher and prematurely terminates the reactive chain from the execution. In most
+ * cases this is not the desired behaviour.
+ *
+ * <p>NB: {@code Mono<?>.zipWith(Mono<Void>)} is allowed be the Reactor API, but it is an incorrect
```
rickie, line 40:
```suggestion
 * <p>NB: {@code Mono<?>.zipWith(Mono<Void>)} is allowed by the Reactor API, but it is an incorrect
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+ * usage of the API. It will be flagged by ErrorProne but the fix won't be supplied. The problem
+ * with the original code should be revisited and fixed in a structural manner by the developer.
```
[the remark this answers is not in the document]
rickie: Hmm good one, will push a proposal to slightly rephrase this when I review a bit deeper.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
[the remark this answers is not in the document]
DimaLegeza: I've double checked. We need the real implementation class, not the abstract interface, otherwise I don't have access to the needed fields under the hood

DimaLegeza:
I fixed a lot of occurrences for `Mono.empty` usage, which turned out to be quite easy to accomplish.
I did not include logic for the following construct: `Mono.empty().zipWith(`, which feels quite logical - should not be used at all and will be 100% spotted in the client codebase within the PR review.

I have reflected this in the PR description as well

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,143 @@
...
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary =
+        "`Mono#zip` and `Mono#zipWith` should not be executed against `Mono#empty` or `Mono<Void>` parameter; "
+            + "please revisit the parameters used and make sure to supply correct publishers instead",
+    link = BUG_PATTERNS_BASE_URL + "MonoZipOfMonoVoidUsage",
+    linkType = CUSTOM,
+    severity = ERROR,
```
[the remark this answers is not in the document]
DimaLegeza: Personally I think it should be ERROR

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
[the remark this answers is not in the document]
DimaLegeza: 👌🏼

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,161 @@
...
+public final class MonoZipOfMonoVoidUsage extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Supplier<Type> MONO = type("reactor.core.publisher.Mono");
+
+  /**
+   * In fact, we use {@code Mono<Void>} everywhere in codebases instead of {@code Mono<Object>}
+   * (actual return type of {@code Mono.empty()}) to represent empty publisher.
+   */
+  private static final Supplier<Type> MONO_VOID_TYPE =
```
mohamedsamehsalah, line 69: WDYT about ?
```suggestion
  private static final Supplier<Type> MONO_VOID_TYPE_SUPPLIER =
```

mohamedsamehsalah (commented):
Sorry for late review, more stylistic suggestions from my side.

Will look into the logic deeper later if @rickie does not do so beforehand 👍

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
DimaLegeza: Also fine by me. Will change other one accordingly

Stephan202: @mohamedsamehsalah I don't think that suggestion is consistent with the rest of the codebase:
```
$ git grep 'private static final Supplier'
error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AssociativeMethodInvocation.java:  private static final Supplier<Type> ITERABLE = Suppliers.typeFromClass(Iterable.class);
error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/FluxFlatMapUsage.java:  private static final Supplier<Type> FLUX =
error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/FluxFlatMapUsage.java:  private static final Supplier<Type> FLUX_OF_PUBLISHERS =
error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/FluxImplicitBlock.java:  private static final Supplier<Type> STREAM =
error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedOptionals.java:  private static final Supplier<Type> OPTIONAL = Suppliers.typeFromClass(Optional.class);
error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NestedPublishers.java:  private static final Supplier<Type> PUBLISHER = type("org.reactivestreams.Publisher");
error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StringJoin.java:  private static final Supplier<Type> CHAR_SEQUENCE_TYPE =
error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StringJoin.java:  private static final Supplier<Type> FORMATTABLE_TYPE = Suppliers.typeFromClass(Formattable.class);
refaster-runner/src/main/java/tech/picnic/errorprone/refaster/runner/CodeTransformers.java:  private static final Supplier<ImmutableListMultimap<String, CodeTransformer>>
```

(In this context the `Supplier` wrapper is mostly a performance optimization; adding it to the name doesn't really increase readability.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,163 @@
...
+    linkType = CUSTOM,
+    severity = ERROR,
+    tags = LIKELY_ERROR)
+public final class MonoZipOfMonoVoidUsage extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Supplier<Type> MONO_SUPPLIER = type("reactor.core.publisher.Mono");
+
+  /**
+   * In fact, we use {@code Mono<Void>} everywhere in codebases instead of {@code Mono<Object>}
```
mohamedsamehsalah, line 64: (Asking because I don't know)

Do we need to keep this in the docs ?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,163 @@
...
+        "`Mono#zip` and `Mono#zipWith` should not be executed against `Mono#empty` or `Mono<Void>` parameter; "
+            + "please revisit the parameters used and make sure to supply correct publishers instead",
+    link = BUG_PATTERNS_BASE_URL + "MonoZipOfMonoVoidUsage",
+    linkType = CUSTOM,
+    severity = ERROR,
+    tags = LIKELY_ERROR)
+public final class MonoZipOfMonoVoidUsage extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Supplier<Type> MONO_SUPPLIER = type("reactor.core.publisher.Mono");
```
mohamedsamehsalah, line 61: Tests still pass if I use `MONO_VOID_TYPE_SUPPLIER` in the code instead of this.

mohamedsamehsalah: And I have no clue why 😢

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,163 @@
...
+              toType(
+                  MethodInvocationTree.class,
+                  hasGenericArgumentOfExactType(MONO_VOID_TYPE_SUPPLIER))),
+          allOf(
+              staticMethod().onClass(MONO_SUPPLIER).named(MONO_ZIP_METHOD),
+              toType(
+                  MethodInvocationTree.class,
+                  hasArguments(
+                      AT_LEAST_ONE,
+                      staticMethod().onClass(MONO_SUPPLIER).named(MONO_EMPTY_METHOD)))));
```
mohamedsamehsalah, line 97: Honestly, did not look deeper if we need all of these matchers, but they definetly _can_ be simplified. I was able to re-write into something like this, WDYT ?
```suggestion
  private static final Supplier<Type> MONO_VOID_TYPE_SUPPLIER =
      VisitorState.memoize(
          generic(type(Mono.class.getCanonicalName()), type(Void.class.getCanonicalName())));

  private static final Matcher<ExpressionTree> MONO_ZIP_AND_MONO_ZIP_WITH =
      anyOf(
          instanceMethod().onDescendantOf(MONO_VOID_TYPE_SUPPLIER).named("zipWith"),
          staticMethod().onClass(MONO_VOID_TYPE_SUPPLIER).named("zip"));

  private static final Matcher<ExpressionTree> GENERIC_ARGUMENT_DERIVED_FROM_MONO_TYPE =
      toType(MethodInvocationTree.class, hasGenericArgumentOfExactType(MONO_VOID_TYPE_SUPPLIER));
  private static final Matcher<ExpressionTree> HAS_MONO_EMPTY_AS_ARGUMENT =
      toType(
          MethodInvocationTree.class,
          hasArguments(
              AT_LEAST_ONE, staticMethod().onClass(MONO_VOID_TYPE_SUPPLIER).named("empty")));
  private static final Matcher<ExpressionTree> OPERATORS_WITH_MONO_VOID_GENERIC_ARGUMENT =
      allOf(
          MONO_ZIP_AND_MONO_ZIP_WITH,
          anyOf(GENERIC_ARGUMENT_DERIVED_FROM_MONO_TYPE, HAS_MONO_EMPTY_AS_ARGUMENT));

  private static final Matcher<ExpressionTree> ANY_MONO_VOID_IN_PUBLISHERS =
      anyOf(
          OPERATORS_WITH_MONO_VOID_GENERIC_ARGUMENT,
          onClassWithMethodName(MONO_VOID_TYPE_SUPPLIER, "zipWith"));
```

mohamedsamehsalah: Feel free to propose better constant names if you agree with this 👍

mohamedsamehsalah (commented):
2.5 pass 🙌  (Sorry for slow progress 🐢)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsageTest.java
```diff
@@ -0,0 +1,118 @@
...
+   * processed by the rule but will be scanned anyway.
+   */
+  @Test
+  void identification() {
+    CompilationTestHelper.newInstance(MonoZipOfMonoVoidUsage.class, getClass())
+        .addSourceLines(
+            "A.java",
+            "import reactor.core.publisher.Mono;",
+            "",
+            "public final class A<T> {",
```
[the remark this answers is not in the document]
DimaLegeza: We use generics because the implementation used to complain about some valid use cases, e.g. `ReactorRules`, line 191:
```
@BeforeTemplate
    Mono<Tuple2<T, S>> before(Mono<T> mono, Mono<S> other) {
      return mono.zipWith(other);
    }
```

That's why I intentionally introduced generic test

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsageTest.java
```diff
@@ -0,0 +1,118 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import com.google.errorprone.BugCheckerRefactoringTestHelper;
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+final class MonoZipOfMonoVoidUsageTest {
+  /**
+   * Line 21 is needed to simulate the unwanted intrinsic operations, which are not intended to be
```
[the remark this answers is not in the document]
DimaLegeza: Here I mean any java operations that require java AST tree lookup (e.g. `getClass()`). Such operations will enter AST-tree specific implementation of the rule, so it was important to verify such lines would be ignored.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
DimaLegeza: Don't have a strong preference

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
DimaLegeza: Yes, makes sense to me.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,157 @@
...
+public final class MonoZipOfMonoVoidUsage extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+
+  /**
+   * In fact, we use {@code Mono<Void>} everywhere in codebases instead of {@code Mono<Object>}
+   * (actual return type of {@link Mono#empty()}) to represent empty publisher.
+   */
+  private static final String MONO_ZIP_WITH_METHOD = "zipWith";
+
```
mohamedsamehsalah, line 66:
```suggestion
  private static final String MONO_ZIP_WITH_METHOD = "zipWith";
```

mohamedsamehsalah (approved):
Nothing else stands out to me 👍 

Thanks @DimaLegeza

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
DimaLegeza: This line is returned back by the formatter :)

DimaLegeza: I'll restructure it a bit

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsageTest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+final class MonoZipOfMonoVoidUsageTest {
+  @Test
+  void identification() {
+    CompilationTestHelper.newInstance(MonoZipOfMonoVoidUsage.class, getClass())
+        .expectErrorMessage(
+            "X",
+            m ->
```
rickie: We don't need this setup :), using 

```
```
Instead of the `matches: X` means exactly the same.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsageTest.java
```diff
@@ -0,0 +1,110 @@
...
+            "",
+            "    Mono.zip(a, a);",
+            "    a.zipWith(a);",
+            "  }",
+            "",
+            "  public void m2() {",
+            "    Class clazz = getClass();",
+            "  }",
+            "}")
+        .doTest(BugCheckerRefactoringTestHelper.TestMode.TEXT_MATCH);
```
rickie:
```suggestion
        .doTest(TestMode.TEXT_MATCH);
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,151 @@
...
+            + "please revisit the parameters used and make sure to supply correct publishers instead",
+    link = BUG_PATTERNS_BASE_URL + "MonoZipOfMonoVoidUsage",
+    linkType = CUSTOM,
+    severity = ERROR,
+    tags = LIKELY_ERROR)
+public final class MonoZipOfMonoVoidUsage extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final String MONO_ZIP_WITH_METHOD = "zipWith";
+  private static final String MONO_ZIP_METHOD = "zip";
```
rickie: If a field is used only once, we can simply inline it.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * A {@link BugChecker} that flags usages of {@link Mono#zip(Mono, Mono)} and {@link
+ * Mono#zipWith(Mono)} with {@link Mono#empty()} parameters.
+ *
+ * <p>{@link Mono#zip(Mono, Mono)} and {@link Mono#zipWith(Mono)} perform incorrectly upon retrieval
+ * of the empty publisher and prematurely terminates the reactive chain from the execution. In most
+ * cases this is not the desired behaviour.
+ *
+ * @apiNote {@code Mono<?>.zipWith(Mono<Void>)} is allowed by the Reactor API, but it is an
+ *     incorrect usage of the API. It will be flagged by ErrorProne but the fix won't be supplied.
```
rickie:
```suggestion
 *     incorrect usage of the API. 
```
We can omit the rest :).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsageTest.java
```diff
@@ -0,0 +1,110 @@
...
+            "public final class A {",
+            "  @SuppressWarnings(\"MonoZipOfMonoVoidUsage\")",
+            "  public void m() {",
+            "    Mono<Void> a = Mono.empty();",
+            "",
+            "    Mono.zip(a, a);",
+            "    a.zipWith(a);",
+            "  }",
+            "",
+            "  public void m2() {",
```
rickie: We can drop this method, in the `identificaiton` method we already showed that the flagging works correctly.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsageTest.java
```diff
@@ -0,0 +1,110 @@
...
+            "",
+            "  private Mono<Integer> zipWith(Mono<Void> param) {",
+            "    return Mono.empty();",
+            "  }",
+            "}")
+        .doTest();
+  }
+
+  @Test
+  void replacementSuggestedFix() {
```
rickie:
```suggestion
  void replacement() {
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsageTest.java
```diff
@@ -0,0 +1,110 @@
...
+                    "`Mono#zip` and `Mono#zipWith` should not be executed against "
+                        + "`Mono#empty` or `Mono<Void>` parameter; please revisit the parameters used and make sure to "
+                        + "supply correct publishers instead"))
+        .addSourceLines(
+            "A.java",
+            "import reactor.core.publisher.Mono;",
+            "",
+            "public final class A {",
+            "  <T> void m(T t) {",
+            "    // This line is needed to simulate the unwanted intrinsic operations, which are not intended to",
```
rickie: We can drop this, it's fine that we need write some code to use in our tests.

rickie (commented):
Added a commit.

First round of mostly style changes :). The matcher logic needs a longer sit to focus.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsageTest.java
DimaLegeza: I wanted to explicitly check on the error message though @rickie

rickie:
Hi, I did a first pass over the code and I have quite some changes locally. It's not in a shape where I can push it, but I'm getting close to a full round of the review. Next week or tomorrow I hope to finish it 😄.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipOfMonoVoidUsage.java
```diff
@@ -0,0 +1,144 @@
...
+      anyOf(
+          instanceMethod().onDescendantOf(MONO_VOID_TYPE).named("zipWith"),
+          staticMethod().onClass(MONO_VOID_TYPE).named("zip"));
+  private static final Matcher<ExpressionTree> GENERIC_ARGUMENT_DERIVED_FROM_MONO_TYPE =
+      toType(MethodInvocationTree.class, hasGenericArgumentOfExactType(MONO_VOID_TYPE));
+  private static final Matcher<ExpressionTree> HAS_MONO_EMPTY_AS_ARGUMENT =
+      toType(
+          MethodInvocationTree.class,
+          hasArguments(AT_LEAST_ONE, staticMethod().onClass(MONO_VOID_TYPE).named("empty")));
+  private static final Matcher<ExpressionTree> OPERATORS_WITH_MONO_VOID_GENERIC_ARGUMENT =
```
rickie, line 71: Not this matcher specific, but I found we can use something from the `MoreTypePredicates#isSubType`.

rickie (commented):
I rewrote some of the logic to make the `Matcher` logic a bit different and slightly less complex. Let me know what you think of the improvements. 

One thing I tried to do with the changes is have a bit more logic in the `matchMethodInvocation` and a bit less in the `Matcher`s 😄.

As I'm not that knowledgeable of Reactor it was really helpful to have some explanation throughout the PR, so thanks a lot for that! I reworded some of the things to try and make it even more clear :).

Thanks for the setup, the tests, and for pushing this @DimaLegeza 🚀 !

rickie (approved):
Made some last tweaks.

Two notes;
- The last mutant requires a look.
- I think a `Mono#zipWith` on a `Mono` that is of type `Mono<Void>` doesn't make sense either, so I added that case as well.

From my side nothing else right now!

@Stephan202 It's ready for your eyes 👀.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipUsage.java
```diff
@@ -0,0 +1,90 @@
...
+    summary = "Don't pass a `Mono<Void>` argument or `Mono.empty()` to Mono#{zip,With}`",
+    link = BUG_PATTERNS_BASE_URL + "MonoZipUsage",
+    linkType = CUSTOM,
+    severity = ERROR,
+    tags = LIKELY_ERROR)
+public final class MonoZipUsage extends BugChecker implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Supplier<Type> MONO =
+      Suppliers.typeFromString("reactor.core.publisher.Mono");
+  private static final Matcher<ExpressionTree> MONO_VOID_OR_MONO_EMPTY =
```
Stephan202: Or just:
```suggestion
  private static final Matcher<ExpressionTree> EMPTY_MONO =
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MonoZipUsage.java
```diff
@@ -0,0 +1,90 @@
...
+      anyOf(
+          staticMethod().onDescendantOf(MONO).named("empty"),
+          typePredicateMatcher(isSubTypeOf(generic(MONO, type(Void.class.getCanonicalName())))));
+  private static final Matcher<ExpressionTree> MONO_ZIP_OR_ZIP_WITH =
+      anyOf(
+          instanceMethod().onDescendantOf(MONO).named("zipWith"),
+          staticMethod().onClass(MONO).named("zip"));
```
Stephan202: Given the structure of the code, I'd swap these two fields.

Stephan202 (approved):
```
Introduce `EmptyMonoZip` check (#733)
```

I like how this check "dried up"!

Stephan202:
Ah, there was still that one mutant to handle; pushed one more commit.

rickie:
Changes LGTM! Rebased it, let's merge this one 🟢 !

PR #745, opened 2023-08-07
Have Renovate file GitHub CodeQL upgrade PRs once every four weeks

rickie:
As discussed [here](https://github.com/PicnicSupermarket/error-prone-support/pull/722#issuecomment-1649187860).

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -12,9 +12,9 @@
       "separateMinorPatch": true
     },
     {
-      "matchDepNames": ["github/codeql-action"],
-      "matchPackagePatterns": [
-        "^ruby\\/setup-ruby$"
+      "matchDepNames": [
+        "github/codeql-action",
+        "ruby/setup-ruby"
       ],
       "schedule": "every 4 weeks on Monday"
     }
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -12,10 +12,11 @@
       "separateMinorPatch": true
     },
     {
+      "matchDepNames": ["github/codeql-action"],
       "matchPackagePatterns": [
         "^ruby\\/setup-ruby$"
       ],
-      "schedule": "* * 1 * *"
+      "schedule": "every 4 weeks on Monday"
     }
   ],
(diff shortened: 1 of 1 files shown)


mohamedsamehsalah (approved):
⏰

.renovaterc.json
```diff
@@ -12,10 +12,11 @@
       "separateMinorPatch": true
     },
     {
+      "matchDepNames": ["github/codeql-action"],
       "matchPackagePatterns": [
         "^ruby\\/setup-ruby$"
       ],
```
Stephan202, line 18: Two questions:
1. These rules are disjunctive (either, but not both need to match)?
2. Since `ruby/setup-ruby` is also a GitHub action, can it also be specified using `matchDepNames` (or, if we want to consistently keep using `matchPackagePatterns`, can `github/codeql-action` be also expressed as such)?

Stephan202 (commented):
The suggested commit message should also mention `setup-ruby`? :thinking:

LGTM otherwise!

.renovaterc.json
rickie: 1. Yes, for example; it opened the CodeQL PR while the PSM is already up-to-date (in my local testing setup).
2. I see there is no entry for Ruby yet in our config, so I can file a PR there to add it.

(I was wondering why we didn't get any updates, but we are still on the latest: https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby).

rickie:
Suggested commit message:
```
Have Renovate file GitHub CodeQL upgrade PRs once every four weeks (#745)

While there, match on the `depName` of `ruby/setup-ruby`. 
```

.renovaterc.json
rickie: Verified with local testing that the `depName` is `ruby/setup-ruby` 😄.

Added a commit to use the deps in a similar way.

Stephan202 (approved):
Nice!

rickie:
Rebased, will merge once 🍏!

PR #757, opened 2023-08-19
Introduce mutation test script for changed code

Stephan202:
Suggested commit message:
```
Introduce mutation test script for changed code (#757)

While there, optimize the existing `run-mutation-tests.sh` script.
```

### DIFF
--- a/run-branch-mutation-tests.sh
+++ b/run-branch-mutation-tests.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+# Executes Pitest to determine the mutation test coverage of changed code. By
+# default code that is changed relative to the default branch is exercised,
+# though another change set may be specified; see the Arcmutate Git integration
+# documentation for details [1]. The results are found in each Maven module's
+# `target/pit-reports` directory.
+#
+# [1] https://docs.arcmutate.com/docs/git-integration.html
+
+set -e -u -o pipefail
+
+if [ "${#}" -gt 1 ]; then
+  echo "Usage: ${0} [DiffSpec]"
+  exit 1
+fi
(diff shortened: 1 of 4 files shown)


nov1n (approved):

run-branch-mutation-tests.sh
```diff
@@ -0,0 +1,23 @@
...
+# though another change set may be specified; see the Arcmutate Git integration
+# documentation for details [1]. The results are found in each Maven module's
+# `target/pit-reports` directory.
+#
+# [1] https://docs.arcmutate.com/docs/git-integration.html
+
+set -e -u -o pipefail
+
+if [ "${#}" -gt 1 ]; then
+  echo "Usage: ${0} [DiffSpec]"
```
rickie, line 14: I think we want to have something like `opt:` in here right? In case of optional arguments we internally usually do:
`<opt: diff-spec>`.

rickie: I see we don't do it for the other, but now we worked a lot more with bash so we made some improvements here and there.

rickie (commented):
One thing to consider.

run-branch-mutation-tests.sh
Stephan202: Hmm, [square brackets represent optionality](https://en.wikipedia.org/wiki/Usage_message#Pattern); this alternative is less idiomatic IMHO.

rickie (approved):

Stephan202:
Wait, we should also update the main README; not sure how I missed that. Will add a commit.

rickie (approved):
Proposal LGTM!

PR #764, opened 2023-08-25
Introduce assorted `Optional` Refaster rules

Stephan202:
Suggested commit message:
```
Introduce assorted `Optional` Refaster rules (#764)

While there, move the `OptionalOrElseGet` Refaster rule test code to the
correct location.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
@@ -101,7 +101,7 @@ final class OptionalRules {
   }
 
   /** Prefer {@link Optional#equals(Object)} over more contrived alternatives. */
-  static final class OptionalHasValue<T, S> {
+  static final class OptionalEqualsOptional<T, S> {
     @BeforeTemplate
     boolean before(Optional<T> optional, S value) {
       return Refaster.anyOf(
@@ -454,5 +454,5 @@ final class OptionalRules {
 
   // XXX: Add a rule for:
   // `optional.flatMap(x -> pred(x) ? Optional.empty() : Optional.of(x))` and variants.
-  // (Maybe canonicalize the inner expression. Maybe we rewrite already.)
+  // (Maybe canonicalize the inner expression. Maybe we rewrite it already.)
 }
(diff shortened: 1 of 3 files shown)


mohamedsamehsalah (approved):

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -100,6 +100,20 @@ Function<Optional<T>, T> after() {
     }
   }
 
+  /** Prefer {@link Optional#equals(Object)} over more contrived alternatives. */
+  static final class OptionalHasValue<T, S> {
```
rickie, line 104: Hmm, I'm not sure if I agree with the name based on our naming algorithm 🤔.

`OptionalEquals` or `OptionalEqualsOptional` seems more accurate, WDYT?

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -422,12 +437,22 @@ Optional<? extends T> after(Optional<S> optional, Function<? super S, ? extends
...
+
+    @AfterTemplate
+    Stream<T> after(Optional<T> optional) {
+      return optional.stream();
+    }
+  }
+
   // XXX: Add a rule for:
   // `optional.flatMap(x -> pred(x) ? Optional.empty() : Optional.of(x))` and variants.
   // (Maybe canonicalize the inner expression. Maybe we rewrite already.)
```
rickie, line 457:
```suggestion
  // (Maybe canonicalize the inner expression. Maybe we rewrite it already.)
```
A tiny improvement IMO.

rickie (approved):
Added a commit with small suggestions. Otherwise LGTM so already approving.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
Stephan202: I guess even `OptionalEqualsOptionalOf`; but this work for me for now :+1:

PR #770, opened 2023-08-29
Introduce `ExhaustiveRefasterTypeMigration` check

Stephan202:
Suggested commit message:
```
Introduce `ExhaustiveRefasterTypeMigration` check (#770)

The new `@TypeMigration` annotation can be placed on Refaster rule
collections to indicate that they migrate most or all public methods of
an indicated type. The new check validates the claim made by the
annotation.
```

This will help flag new to-be-migrated methods any time the JUnit or TestNG dependency is upgraded. (This PR also flags that `TestNGToAssertJRules` is no longer nearly as exhaustive as it once was. Resolving that will be deferred to a separate PR. I started a [TestNG](https://github.com/PicnicSupermarket/error-prone-support/compare/sschroevers/more-testng-to-assertj-rules) and [JUnit](https://github.com/PicnicSupermarket/error-prone-support/compare/sschroevers/more-junit-to-assertj-rules) branch. That said, given the repetition involved, perhaps introducing a bug checker or two is in order.)


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationListing.java
```diff
@@ -36,6 +36,8 @@
  * <p>The idea behind this checker is that maintaining a sorted sequence simplifies conflict
  * resolution, and can even avoid it if two branches add the same annotation.
  */
+// XXX: Duplicate entries are often a mistake. Consider introducing a similar `BugChecker` that
```
mohamedsamehsalah, line 39: Interested 👀

refaster-support/src/main/java/tech/picnic/errorprone/refaster/annotation/TypeMigration.java
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * The signatures of public methods and constructors that are not (yet) migrated by the annotated
+   * Refaster rule(s).
+   *
+   * @return A possibly empty enumeration of method and constructor signatures, formatted according
+   *     to {@link
+   *     com.google.errorprone.util.Signatures#prettyMethodSignature(com.sun.tools.javac.code.Symbol.ClassSymbol,
+   *     com.sun.tools.javac.code.Symbol.MethodSymbol)}.
+   */
+  String[] unmigratedMethods() default {};
```
mohamedsamehsalah, line 36: Intellij flags this as a spelling mistake 🤔 
Maybe:
```suggestion
  String[] nonMigratedMethods() default {};
```
or
```suggestion
  String[] notMigratedMethods() default {};
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ExhaustiveRefasterTypeMigrationTest.java
```diff
@@ -0,0 +1,252 @@
...
+            "        \"publicStringMethodWithArg(int)\",",
+            "        \"publicStaticVoidMethod()\"",
+            "      })",
+            "  class AnnotatedEmptyClass {}",
+            "",
+            "  @TypeMigration(",
+            "      of = Util.class,",
+            "      unmigratedMethods = {",
+            "        \"publicStaticVoidMethod()\",",
+            "        \"publicStringMethodWithArg(int)\",",
```
mohamedsamehsalah, line 59: Do we need to test overloaded methods here ?
```suggestion
            "        \"publicStringMethodWithArg(int)\",",
            "        \"publicStringMethodWithArg(String)\",",
```

mohamedsamehsalah: Which brings me to my question on why do we need to flag all the variants of the overloaded method ? 🤔 

Is it wrong to assume that if a rule does not apply to one type set of arguments, will not be applied to other ?

mohamedsamehsalah (approved):
Nothing stands out to me 🚀

Stephan202 (commented):
Rebased and added two small commits.

Stephan202:
I'll rebase this branch and move the check to the new `guidelines` module.

rickie (approved):
Amazing check 🚀 !

refaster-support/src/main/java/tech/picnic/errorprone/refaster/annotation/TypeMigration.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Indicates that a Refaster rule or group of Refaster rules is intended to migrate away from the
+ * indicated type.
+ */
+// XXX: Add support for `#unmigratedFields()`.
+// XXX: Consider making this annotation `@Repeatable`, for cases where a single Refaster rule
+// collection migrates away from multiple types.
+@Retention(RetentionPolicy.SOURCE)
+@Target(ElementType.TYPE)
+public @interface TypeMigration {
```
Stephan202, line 17: On the topic of annotation sorting: makes sense. We should extend `LexicographicalAnnotationListing`; will add a comment.

refaster-support/src/main/java/tech/picnic/errorprone/refaster/annotation/TypeMigration.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Indicates that a Refaster rule or group of Refaster rules is intended to migrate away from the
+ * indicated type.
+ */
+// XXX: Add support for `#unmigratedFields()`.
+// XXX: Consider making this annotation `@Repeatable`, for cases where a single Refaster rules
+// collection migrates away from multiple types.
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.SOURCE)
```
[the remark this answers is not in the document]
Stephan202: I agree that we don't lose much, but the current variant does seem a bit clearer to me. Since this annotation will only be present on a separate line in (almost always) small files, I don't think there's much point in forcing a static import here. Internally only 2 out of 67 usages statically import it.

error-prone-guidelines/src/test/java/tech/picnic/errorprone/guidelines/bugpatterns/ExhaustiveRefasterTypeMigrationTest.java
```diff
@@ -0,0 +1,263 @@
...
+            "  @TypeMigration(of = Util.class, unmigratedMethods = \"publicStringMethodWithArg(int)\")",
+            "  class AnnotatedPartiallyMigrated {",
+            "    @BeforeTemplate",
+            "    void before() {",
+            "      Util.publicStaticVoidMethod();",
+            "      Util.publicStaticIntMethod2();",
+            "    }",
+            "  }",
+            "",
+            "  // BUG: Diagnostic contains: The set of unmigrated methods listed by the `@TypeMigration`",
```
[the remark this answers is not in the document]
Stephan202: This check emits two kinds of messages, so for each violation I'm asserting that the correct message is emitted.

Stephan202 (commented):
Rebased and added a small commit. Will merge once built.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationListing.java
mohamedsamehsalah: #1070

PR #771, opened 2023-08-30
Introduce `PatternRules` Refaster rule collection

Stephan202:
Suggested commit message:
```
Introduce `PatternRules` Refaster rule collection (#771)

While there, configure `StaticImport` to not require static importing of
`com.google.common.base.Predicates.contains`.

The new rules triggered cleanup of some `CompilationTestHelper` usages.

See google/guava#6483.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
@@ -155,6 +155,7 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
   @VisibleForTesting
   static final ImmutableSetMultimap<String, String> STATIC_IMPORT_EXEMPTED_MEMBERS =
       ImmutableSetMultimap.<String, String>builder()
+          .put("com.google.common.base.Predicates", "contains")
           .put("com.mongodb.client.model.Filters", "empty")
           .putAll(
               "java.util.Collections",
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
@@ -1,0 +1,46 @@
+package tech.picnic.errorprone.refasterrules;
+
+import static com.google.common.base.Predicates.containsPattern;
+
+import com.google.common.base.Predicates;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
+
+/** Refaster rules related to code dealing with regular expressions. */
+@OnlineDocumentation
+final class PatternRules {
+  private PatternRules() {}
+
+  /** Prefer {@link Pattern#asPredicate()} over non-JDK alternatives. */
+  // XXX: This rule could also replace `s -> pattern.matcher(s).find()`, though the lambda
+  // expression may match functional interfaces other than `Predicate`. If we do add such a rule, we
+  // should also add a rule that replaces `s -> pattern.matcher(s).matches()` with
+  // `pattern.asMatchPredicate()`.
+  static final class PatternAsPredicate {
+    @BeforeTemplate
+    Predicate<CharSequence> before(Pattern pattern) {
+      return Predicates.contains(pattern);
+    }
+
+    @AfterTemplate
+    Predicate<String> after(Pattern pattern) {
+      return pattern.asPredicate();
+    }
+  }
+
+  /** Prefer {@link Pattern#asPredicate()} over non-JDK alternatives. */
+  static final class PatternCompileAsPredicate {
+    @BeforeTemplate
+    Predicate<CharSequence> before(String pattern) {
+      return containsPattern(pattern);
+    }
+
+    @AfterTemplate
+    Predicate<String> after(String pattern) {
+      return Pattern.compile(pattern).asPredicate();
+    }
+  }
+}
--- a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/FluxImplicitBlockTest.java
+++ b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/FluxImplicitBlockTest.java
@@ -1,14 +1,12 @@
 package tech.picnic.errorprone.bugpatterns;
 
-import static com.google.common.base.Predicates.and;
-import static com.google.common.base.Predicates.containsPattern;
-import static com.google.common.base.Predicates.not;
 import static com.google.errorprone.BugCheckerRefactoringTestHelper.FixChoosers.SECOND;
 import static com.google.errorprone.BugCheckerRefactoringTestHelper.FixChoosers.THIRD;
 
 import com.google.errorprone.BugCheckerRefactoringTestHelper;
 import com.google.errorprone.BugCheckerRefactoringTestHelper.TestMode;
 import com.google.errorprone.CompilationTestHelper;
+import java.util.stream.Stream;
 import org.junit.jupiter.api.Test;
 import org.reactivestreams.Publisher;
 import reactor.core.CorePublisher;
@@ -20,10 +18,7 @@ final class FluxImplicitBlockTest {
     CompilationTestHelper.newInstance(FluxImplicitBlock.class, getClass())
         .expectErrorMessage(
             "X",
-            and(
-                containsPattern("SuppressWarnings"),
-                containsPattern("toImmutableList"),
-                containsPattern("toList")))
+            m -> Stream.of("SuppressWarnings", "toImmutableList", "toList").allMatch(m::contains))
         .addSourceLines(
             "A.java",
             "import com.google.common.collect.ImmutableList;",
@@ -63,7 +58,7 @@ final class FluxImplicitBlockTest {
   void identificationWithoutGuavaOnClasspath() {
     CompilationTestHelper.newInstance(FluxImplicitBlock.class, getClass())
         .withClasspath(CorePublisher.class, Flux.class, Publisher.class)
-        .expectErrorMessage("X", not(containsPattern("toImmutableList")))
+        .expectErrorMessage("X", m -> !m.contains("toImmutableList"))
         .addSourceLines(
             "A.java",
             "import reactor.core.publisher.Flux;",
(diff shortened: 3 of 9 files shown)


refaster-runner/src/test/java/tech/picnic/errorprone/refaster/runner/RefasterTest.java
```diff
@@ -167,13 +164,18 @@ void severityAssignment(
...
-        .isInstanceOf(AssertionError.class)
-        .message()
-        .satisfies(
-            message ->
-                assertThat(extractRefasterSeverities("A.java", message))
-                    .containsExactlyElementsOf(expectedSeverities));
+
+    if (expectedSeverities.isEmpty()) {
+      compilationTestHelper.doTest();
+    } else {
```
Stephan202, line 170: Previously this wasn't necessary, because the `expectErrorMessage` declarations would trigger a "I didn't see anything" `AssertionError`, which would then cause the `satisfies` clause to succeed. But that's obscure.

rickie (approved):
Only one suggestion. Will approve nonetheless. Feel free when the second approval is in 😉.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
```diff
@@ -155,6 +155,7 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
   @VisibleForTesting
   static final ImmutableSetMultimap<String, String> STATIC_IMPORT_EXEMPTED_MEMBERS =
       ImmutableSetMultimap.<String, String>builder()
+          .put("com.google.common.base.Predicates", "contains")
```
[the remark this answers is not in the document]
Stephan202: Interesting suggestion. Perhaps yes, though (hypothetical?) static `contains`/`empty` methods that deal with "regular" collection membership would not necessarily be problematic when imported statically. (The currently exempted methods instead return predicates with custom semantics.)

So... let's revisit if/when we add more such exclusions?

Stephan202 (commented):
Tnx for the review!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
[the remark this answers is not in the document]
rickie: Deal, let's do that :).

rickie:
Rebased.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
```diff
@@ -0,0 +1,46 @@
...
+  /** Prefer {@link Pattern#asPredicate()} over non-JDK alternatives. */
+  static final class PatternCompileAsPredicate {
+    @BeforeTemplate
+    Predicate<CharSequence> before(String pattern) {
+      return containsPattern(pattern);
+    }
+
+    @AfterTemplate
+    Predicate<String> after(String pattern) {
+      return Pattern.compile(pattern).asPredicate();
```
Venorcis, line 43: IMO the Guava static builder is a nice short alternative; not sure I see the advantage in standardizing here (they also do not recommend themselves to not use it like with some other utilities we don't rewrite like `Lists.transform`)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
```diff
@@ -0,0 +1,46 @@
...
+    Predicate<String> after(Pattern pattern) {
+      return pattern.asPredicate();
+    }
+  }
+
+  /** Prefer {@link Pattern#asPredicate()} over non-JDK alternatives. */
+  static final class PatternCompileAsPredicate {
+    @BeforeTemplate
+    Predicate<CharSequence> before(String pattern) {
+      return containsPattern(pattern);
```
Venorcis, line 38: why the static import here and not above?

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
```diff
@@ -0,0 +1,46 @@
...
+  // `pattern.asMatchPredicate()`.
+  static final class PatternAsPredicate {
+    @BeforeTemplate
+    Predicate<CharSequence> before(Pattern pattern) {
+      return Predicates.contains(pattern);
+    }
+
+    @AfterTemplate
+    Predicate<String> after(Pattern pattern) {
+      return pattern.asPredicate();
```
Venorcis, line 30: also here I'd personally argue that the Guava variant is more descriptive (nor do they recommend against using it)

I'm also not sure it's 100% equivalent as JDK uses `String` vs. Guava's `CharSequence`

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
Stephan202: That relates to this change:
> While there, configure `StaticImport` to not require static importing of
`com.google.common.base.Predicates.contains`.

(I.e. the change to `StaticImport`.) We discourage static imports of methods with a rather generic name, which is the case for `contains`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
Stephan202: I can see how this suggestion is slightly controversial, but my thinking is that, besides being a JDK alternative:
1. This code makes explicit that the operation involves pattern compilation; something one will generally want to move to a static field whenever possible.
    * Or, as in the cleaned-up test code, avoid altogether by simply calling `String#contains` instead. 
3. This code better conveys (through the explicit `Pattern` reference) that the predicate performs regex matching, rather than any other kind of pattern matching.

W.r.t. `List#transform`: indeed we don't rewrite that variant, but that's primarily because realistically it can only be rewritten when considering the surrounding context. That method's [documentation](https://guava.dev/releases/32.1.1-jre/api/docs/com/google/common/collect/Lists.html#transform(java.util.List,com.google.common.base.Function)) does suggest that Java 8+ users should prefer `Stream#map`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
Stephan202: > also here I'd personally argue that the Guava variant is more descriptive (nor do they recommend against using it)

On descriptiveness I guess we'll just have to disagree :sweat_smile:. W.r.t. recommendations: while `contains` itself doesn't call it out, the returned [`com.google.common.base.Predicate`](https://guava.dev/releases/32.1.1-jre/api/docs/com/google/common/base/Predicate.html) is described as a legacy type that should be avoided whenever possible.


> I'm also not sure it's 100% equivalent as JDK uses `String` vs. Guava's `CharSequence`

Indeed it isn't; this is a Refaster limitation that unfortunately more of our rules suffer from. (Longer-term I plan to automatically flag and annotate such rules such that they can e.g. be skipped.)

Stephan202 (commented):
Partially the discussion may be moot; a `git grep 'Predicates.contains'` against the Picnic code base yields only the Error Prone Support code modified in this PR :smile:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
Venorcis: > the returned [com.google.common.base.Predicate](https://guava.dev/releases/32.1.1-jre/api/docs/com/google/common/base/Predicate.html) is described as a legacy type that should be avoided whenever possible.

good point, should that be 'attacked' even more generically then? essentially the entire `Predicates` class shouldn't be used then 😅

benhalasi (commented):
Cool stuff!

Left just one question about the testing.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.common.base.Predicates;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
+
+/** Refaster rules related to code dealing with regular expressions. */
+@OnlineDocumentation
+final class PatternRules {
```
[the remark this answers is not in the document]
rickie: That means that this one is not actually being tested, so we definitely should add it.
Now that I think of it, we forgot that part for the #811 PR 🤯, will open a PR right now.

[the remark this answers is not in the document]
Stephan202: Nice catch; will fix! (And indeed, cue the "we should automate this" message :sweat_smile:)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PatternRules.java
Stephan202: Ideally yes! `git grep com.google.common.base.Predicate -- '*.java'` shows that internally not too many usages remain, so it's doable. But in some cases the result may be passed to a `com.google.common.base.Predicate`-expecting method, in which case a Refaster approach isn't great. That's the same issue as we have here, but I expect that the impact will be larger when replacing `or` and `not`. So propose to defer that to a separate PR.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/FluxImplicitBlockTest.java
```diff
@@ -20,10 +18,7 @@ void identification() {
     CompilationTestHelper.newInstance(FluxImplicitBlock.class, getClass())
         .expectErrorMessage(
             "X",
-            and(
-                containsPattern("SuppressWarnings"),
-                containsPattern("toImmutableList"),
-                containsPattern("toList")))
+            m -> Stream.of("SuppressWarnings", "toImmutableList", "toList").allMatch(m::contains))
```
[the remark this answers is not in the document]
Stephan202: Hmm, I would say that such a utility method has an insufficient [utility times ubiquity (or power-to-weight) ratio](https://github.com/google/guava/wiki/PhilosophyExplained#when-in-doubt). Separately from that, when I see `str -> str.contains("foo")` I unambiguously know what that code will do, while with `contains("foo")` I'm like "that's funny, a helper method for such a simple operation; let me check the implementation to see whether it does anything special". So all-in-all I favour the current approach.

Stephan202 (commented):
Tnx for all the careful reviews! :heart:

Venorcis (approved):

benhalasi (approved):

rickie:
Rebased, will merge once 📗 , nice discussions 🚀 !

Issue #632, opened 2023-05-17
Canonical SLF4J Logger usage


rickie:
### Problem

Internally we heavily use SLF4J for logging in our classes.
We would like to have a check to canonicalize the usage of such loggers. 
For that reason we would like a check that looks at the following things:
- Is the variable declared with the modifiers `private static final`?
- Is the variable name of the logger `LOG`?
- Is the correct class name passed to `LoggerFactory#getLogger`?

### Description of the proposed new feature

- [x] Support a stylistic preference.
- [x] Avoid a common gotcha, or potential problem.
- [ ] Improve performance.

I would like to rewrite the following code:
```java
final class SimpleExample {
  public Logger LOGGER = LoggerFactory.getLogger(WrongClass.class);
  ...
}
```

to:
```java
final class SimpleExample {
  private static final Logger LOG = LoggerFactory.getLogger(SimpleExample.class);
  ...
}
```

### Considerations

For interfaces we have to make an exception when it comes to the `private static final`, as in interfaces this is not an option.
As raised by @mlrprananta himself, we could parameterize the variable name of the logger. Perhaps we can add an option that people can override the allowed variable name of the logger via the `ErrorProneFlags`.

For this check we don't need to look at the line on which the logger is defined, we will do that in a different check. 

### Additional context

This project has some related checks [KengoTODA/errorprone-slf4j](https://github.com/KengoTODA/errorprone-slf4j). 
However, we want a check that is specific to our internal standards. Additionlly, in EPS we have our own best practices that we want to use for the code of this check.

Later, we might want to add more features to this check. This would be a good first version 😄.

Stephan202:
I wonder whether we should default to a more permissible name pattern (e.g. `LOG(GER)?`), and then internally fix that to just `LOG`. This may ease external adoption. (Perhaps using CodeQL we can find which variants are most common in open source software :thinking:.)

benhalasi:
In the issue description:
> - Is the check declared with the modifiers private static final?

Do you mean *"Is the variable declared"*?

rickie:
@mlrprananta Are you still planning on working on this issue? Otherwise it's no problem but I'll unassign you. I'm just curious 😄.

PR #783, opened 2023-09-08
Introduce `Slf4jLoggerDeclaration` check

mohamedsamehsalah:
#632 

Suggested commit message:
```
Introduce `Slf4jLoggerDeclaration` check (#783)
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,129 @@
...
+    summary = "Make sure SLF4J log declarations follow the SLF4J conventions inside classes.",
+    link = BUG_PATTERNS_BASE_URL + "Slf4jLogDeclaration",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = LIKELY_ERROR)
+public final class Slf4jLogDeclaration extends BugChecker
+    implements ClassTreeMatcher, MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<ClassTree> TEST_CLASS_WITH_LOGGER =
+      allOf(hasIdentifier(isSubtypeOf("org.slf4j.Logger")));
```
mohamedsamehsalah, line 53: (It's late at night so I am sure I am missing something big about Java 😅 )
I am curious how `isSubtypeOf` is implictly casted here 😬 
Obviously the compiler complain when I do this
```suggestion
private static final Matcher<Tree> LOGGER = isSubtypeOf("org.slf4j.Logger");
      allOf(hasIdentifier((Matcher<IdentifierTree>)LOGGER));
```
but not when the method is passed as an argument to the `hasIdentifier` method ?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
+    for (Tree member : tree.getMembers()) {
+      if (LOGGER.matches(member, state)) {
+        VariableTree variable = (VariableTree) member;
+        SuggestedFixes.addModifiers(
+                member, state, Modifier.PRIVATE, Modifier.STATIC, Modifier.FINAL)
+            .ifPresent(fixBuilder::merge);
+        if (!variable.getName().toString().startsWith(canonicalizedLoggerName)) {
```
Stephan202, line 91: Hmm, why `startsWith`? :thinking:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+      }
+    }
+
+    return describeMatch(tree, fixBuilder.build());
+  }
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!GET_LOGGER_METHOD.matches(tree, state)) {
```
Stephan202, line 102: We're now creating separate fixes for the field declaration (modifiers + name) and for the referenced class. I wonder whether this means that we should have two _separate_ checks. I can imagine that some users will appreciate the second, while finding the first too conservative.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclarationTest.java
```diff
@@ -0,0 +1,93 @@
...
+            "",
+            "interface A {",
+            "  Logger LOG = LoggerFactory.getLogger(A.class);",
+            "",
+            "  Logger LOG1 = LoggerFactory.getLogger(A.class);",
+            "",
+            "  class B {}",
+            "}")
+        .doTest(TestMode.TEXT_MATCH);
+  }
```
Stephan202, line 92: For this we can instead have a regular `identification` test with a nested interface, and then verify that it isn't flagged.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
Stephan202: This happens because `hasIdentifier` requires a `Matcher<IdentifierTree>`, while `LOGGER` is a `Matcher<Tree>`; that's just how Java's type system works. (But in this case the point is moot: as mentioned below: we don't need this `TEST_CLASS_WITH_LOGGER` check.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,129 @@
...
+    if (tree.getKind() == Kind.INTERFACE || !TEST_CLASS_WITH_LOGGER.matches(tree, state)) {
+      return Description.NO_MATCH;
+    }
+
+    SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
+    for (Tree member : tree.getMembers()) {
+      if (LOGGER.matches(member, state)) {
+        VariableTree variable = (VariableTree) member;
+        SuggestedFixes.addModifiers(
+                member, state, Modifier.PRIVATE, Modifier.STATIC, Modifier.FINAL)
```
Stephan202, line 89: Making the field `final` or `private` _could_ make the code fail to compile. Likely this is still the right trade-off, but perhaps something worth calling out.

Stephan202 (commented):
Quick skim; hope this helps :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
mohamedsamehsalah: Does it make sense to make the suggested modifiers EP flags instead ?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
mohamedsamehsalah: Pushed a change using TreeScanner (Conceptually still two checks using`visitClass` & `visitMethodInvocationTree`).

What do you think about making the separation into an EP flag ?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
mohamedsamehsalah: The idea was that we can have a `LOG1` & a `LOG2` within the same class, so we can canonicalize the begining of the variable name.

(This means we assume each class will have one `Logger`; Which is valid assumption I know, but an assumption nevertheless)

Updated ✅

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,129 @@
...
+   * @param flags Any provided command line flags.
+   */
+  @Inject
+  Slf4jLogDeclaration(ErrorProneFlags flags) {
+    canonicalizedLoggerName = getCanonicalizedLoggerName(flags);
+  }
+
+  @Override
+  public Description matchClass(ClassTree tree, VisitorState state) {
+    if (tree.getKind() == Kind.INTERFACE || !TEST_CLASS_WITH_LOGGER.matches(tree, state)) {
```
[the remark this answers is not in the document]
mohamedsamehsalah: > I don't think we need to skip interfaces

The intention was to not apply modifiers to inteface variables but still suggest/fix field name.

> : for those we can still suggest (but not auto-fix) the canonical field name.

Why only suggest and not auto-fix ?

> (Separately from that we should drop redundant modifiers, but that should be a separate check, modeled after [this Checkstyle rule](https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/modifier/RedundantModifierCheck.html).)

Should this check be part of this PR ?

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclarationTest.java
```diff
@@ -0,0 +1,98 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import com.google.errorprone.BugCheckerRefactoringTestHelper;
+import com.google.errorprone.BugCheckerRefactoringTestHelper.TestMode;
+import org.junit.jupiter.api.Test;
+
+final class Slf4jLogDeclarationTest {
+  @Test
+  void replacement() {
+    BugCheckerRefactoringTestHelper.newInstance(Slf4jLogDeclaration.class, getClass())
```
rickie, line 10: Usually we have the `identification` test which is the most extensive one where we test all the edge cases. The replacement test only shows that the replacement logic is correct. Apart from that we don't want to have all the different cases in the `replacement` test.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  /**
+   * Instantiates a customized {@link Slf4jLogDeclaration}.
+   *
+   * @param flags Any provided command line flags.
+   */
+  @Inject
+  Slf4jLogDeclaration(ErrorProneFlags flags) {
+    canonicalizedLoggerName = getCanonicalizedLoggerName(flags);
```
rickie, line 70: We should add 1 test case where we pass a different logger name to this BugPattern and validate that it is correct. 

See this code from `LexicographicalAnnotationAttributeListingTest`:
```
 CompilationTestHelper.newInstance(LexicographicalAnnotationAttributeListing.class, getClass())
        .setArgs(
            ImmutableList.of(
                "-XepOpt:LexicographicalAnnotationAttributeListing:Includes=pkg.A.Foo,pkg.A.Bar",
                "-XepOpt:LexicographicalAnnotationAttributeListing:Excludes=pkg.A.Bar#value"))
```
We can do the same but for our flags here :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,136 @@
...
+  private static void fixLoggerVariableModifiers(
+      Tree member, VisitorState state, SuggestedFix.Builder fixBuilder) {
+    SuggestedFixes.addModifiers(member, state, Modifier.PRIVATE, Modifier.STATIC, Modifier.FINAL)
+        .ifPresent(fixBuilder::merge);
+  }
+
+  private void canonicalizeLoggerVariable(
+      Tree member, VisitorState state, SuggestedFix.Builder fixBuilder) {
+    VariableTree variable = (VariableTree) member;
+    if (!variable.getName().toString().equals(canonicalizedLoggerName)) {
```
rickie, line 99:
```suggestion
    if (!variable.getName().contentEquals(canonicalizedLoggerName)) {
```
I think this would work too.

rickie (commented):
Some comments we discussed offline :).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclarationTest.java
mohamedsamehsalah: Thanks @rickie .

Updating the test actually made me spot a fault in the implementation logic. I updated the code in the checker and the tests. PTAL 🙏

mohamedsamehsalah (commented):
Sorry for the slow progress 🐢

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,132 @@
...
+@BugPattern(
+    summary = "Make sure SLF4J log declarations follow the SLF4J conventions inside classes.",
+    link = BUG_PATTERNS_BASE_URL + "Slf4jLogDeclaration",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = LIKELY_ERROR)
+public final class Slf4jLogDeclaration extends BugChecker implements ClassTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<Tree> LOGGER = isSubtypeOf("org.slf4j.Logger");
+  private static final String CANONICAL_LOGGER_NAME_FLAG =
```
rickie, line 50: We can rename from `canonicalized` to `canonical` in many cases, I think it reads nicer :).

rickie (commented):
Sorry I forgot to add comments on the PR. I noticed some improvements that we could do on the PR. 

Here we have the same that the bug is reported on the class instead of on the actual variable itself... If we fix that, this PR is close to completion I think.

rickie (commented):
Three things:
- The following test fails the code:
```
            "  static final class NegativeOne {",
            "    private static final Logger LOG = LoggerFactory.getLogger(\"NegativeOne\");",
            "  }",
```
There is an overload we need to take into account :).

- There are some mutants here that we need to kill, probably by adding more test cases. 
For example we are missing the normal test case:
```
            "  private static final Logger LOG = LoggerFactory.getLogger(A.class);",
```
And some others have overlap, I think it would be nice to test one thing per test and have a clear description of the class as well. 
I think `static class F` and `static class G` are testing the same thing.

- Do we want to consider _dropping_ modifiers in some cases? We could also make an XXX about that and defer that. That could be a nice idea.

rickie:
I added another commit with a proposal for the test cases, LMKWYT and we can discuss it tomorrow of course.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,154 @@
...
+      public @Nullable Void visitMethodInvocation(
+          MethodInvocationTree tree, Name enclosingElementName) {
+        if (GET_LOGGER_METHOD.matches(tree, state)) {
+          ExpressionTree arg = Iterables.getOnlyElement(tree.getArguments());
+          String argumentName = SourceCode.treeToString(arg, state);
+
+          String argumentClassName;
+          if (arg.getKind() == Kind.STRING_LITERAL) {
+            java.util.regex.Matcher matcher = STRING_LITERAL_ARGUMENT.matcher(argumentName);
+            checkArgument(matcher.matches(), "Invalid argument name.");
```
mohamedsamehsalah, line 126: Here and below, this check is used to invoke the matching.

mohamedsamehsalah (commented):
Thanks @rickie , I resolved the comments from the pair programming sessions and tried to clean up some of the mutants 🐞. PTAL 🙏 .

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,154 @@
...
+      VisitorState state) {
+    new TreeScanner<@Nullable Void, Name>() {
+      @Override
+      public @Nullable Void visitMethodInvocation(
+          MethodInvocationTree tree, Name enclosingElementName) {
+        if (GET_LOGGER_METHOD.matches(tree, state)) {
+          ExpressionTree arg = Iterables.getOnlyElement(tree.getArguments());
+          String argumentName = SourceCode.treeToString(arg, state);
+
+          String argumentClassName;
```
rickie, line 123: We can simplify the code  a little: 

```
       java.util.regex.Matcher matcher;
          if (arg.getKind() == Kind.STRING_LITERAL) {
            matcher = STRING_LITERAL_ARGUMENT.matcher(argumentName);
          } else {
            matcher = CLASS_ARGUMENT.matcher(argumentName);
          }

          checkArgument(matcher.matches(), "Invalid argument name.");
          String argumentClassName = matcher.group(1);
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  private void canonicalizeLoggerVariable(
+      VariableTree variableTree, SuggestedFix.Builder fixBuilder, VisitorState state) {
+    if (!variableTree.getName().contentEquals(canonicalLoggerName)) {
+      fixBuilder.merge(SuggestedFixes.renameVariable(variableTree, canonicalLoggerName, state));
+    }
+  }
+
+  private static void updateGetLoggerArgument(
```
rickie, line 110: I'm not seeing why we need a `TreeScanner` to walk over a `VariableTree`, I think we can improve this :).

rickie (commented):
Did another pass, some small improvements here and there :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
mohamedsamehsalah: We do this because we need to match on the method invocation of the variable. In case the variable is in a nested class, without this, the match will be on the outer class level (Which was flagged [here](https://github.com/PicnicSupermarket/error-prone-support/blob/14f46b59a8dccbec453b836cbca08a62963fc451/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java#L48)).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclarationTest.java
```diff
@@ -0,0 +1,178 @@
...
+            "    }",
+            "  }",
+            "",
+            "  static class WrongArgumentGetLogger {",
+            "    private static final Logger LOG = LoggerFactory.getLogger(String.class);",
+            "  }",
+            "",
+            "  static class WrongArgumentOverloadedGetLogger {",
+            "    // BUG: Diagnostic contains:",
+            "    private static final Logger LOG = LoggerFactory.getLogger(\"OverloadedGetLogger\");",
```
rickie, line 109: On a second look, I'm not really sure how we should handle these string overloads... There are quite some different ways of using them, if we look outside of [Picnic to usages](https://github.com/search?q=LoggerFactory.getLogger%28%22&type=code). Within Picnic we have a few like `Something.NameOfclass`, so not sure what we want to do there... 

CC: @Stephan202 What do you think?

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclarationTest.java
```diff
@@ -0,0 +1,178 @@
...
+            "      LOG.trace(\"foo\");",
+            "    }",
+            "  }",
+            "",
+            "  static class WrongArgumentGetLogger {",
+            "    private static final Logger LOG = LoggerFactory.getLogger(WrongArgumentGetLogger.class);",
+            "  }",
+            "",
+            "  static class WrongArgumentOverloadedGetLogger {",
+            "    // BUG: Diagnostic contains:",
```
rickie: We can drop this one in a replacement test 😉.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclarationTest.java
```diff
@@ -0,0 +1,178 @@
...
+    CompilationTestHelper.newInstance(Slf4jLogDeclaration.class, getClass())
+        .addSourceLines(
+            "A.java",
+            "import org.slf4j.Logger;",
+            "import org.slf4j.LoggerFactory;",
+            "",
+            "class A {",
+            "  private static final long serialVersionUID = 1L;",
+            "",
+            "  private static final Logger LOG = LoggerFactory.getLogger(A.class);",
```
rickie, line 21: I put the "positive" examples at the top of the file :).

rickie (commented):
I think there is only one open question for me that we should answer before approving. Apart from that, I have almost nothing to add here!

Really cool stuff @mohamedsamehsalah 🚀 !

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclarationTest.java
mohamedsamehsalah: > ...if we look outside of [Picnic to usages](https://github.com/search?q=LoggerFactory.getLogger%28%22&type=code)....

You're missing the filter on [Java code](https://github.com/search?q=LoggerFactory.getLogger%28%22+language%3AJava&type=code&l=Java) for this

Stephan202: Good question @rickie. It does seem safer to ignore these cases, or or put this logic under a flag. But I'm not sure there's much use in the latter, because users should favour the `Class` overload anyway when possible. So any string variant should likely be considered "custom". TL;DR: seems better to cover only `Class`-accepting `getLogger` calls. WDYT?

mohamedsamehsalah: > But I'm not sure there's much use in the latter, because users should favour the Class overload anyway when possible.

Can / should we flag this as part of the check?

rickie (approved):
With the final discussion resolved, it's actually time for the approval. Nice work again @mohamedsamehsalah ! 🚀 Can't wait to roll this out.

Badbond:
While trying to introduce an ArchUnit rule (for the interim period), we spotted another interesting use-case that you might want to consider for this bug check. Essentially there may be a reason to not have a `static `logger and use `this.getClass()` for the logger class:

```java
public abstract class AbstractHandler {
  /** A logger for the concrete class. */
  private final Logger LOG = LoggerFactory.getLogger(this.getClass());
  
  public void someMethod() {
    LOG.info("Doing something") // Logs using the concrete class, MyHandler in this example.
  }
}

final class MyHandler extends AbstractHandler {}
```

We have several such cases in the internal code base and I think this is quite an elegant solution to make sure the right class is mentioned in the log lines.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
Stephan202: Indeed; as I noticed this also [here](https://github.com/PicnicSupermarket/error-prone-support/pull/794/files#r1818105009), I proposed a Refaster rule for this: #1379.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLogDeclaration.java
```diff
@@ -0,0 +1,134 @@
...
+ *     arguments}.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Make sure SLF4J log declarations follow the SLF4J conventions inside classes.",
+    link = BUG_PATTERNS_BASE_URL + "Slf4jLogDeclaration",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = STYLE)
+public final class Slf4jLogDeclaration extends BugChecker implements VariableTreeMatcher {
```
Stephan202: While indeed we prefer calling the constant `LOG`, the type is a `Logger`. As such `Slf4jLoggerDeclaration` seems like a better name.

Stephan202 (commented):
I rebased and added a commit. Quick some things changed (including support for the pattern flagged by @Badbond), so PTAL.

Suggested commit message:
```
Introduce `Slf4jLoggerDeclaration` check (#783)
```

Stephan202 (approved):

Stephan202:
After some sleep I added another commit with some comments and tweaks.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/Slf4JLoggerDeclarationTest.java
```diff
@@ -0,0 +1,219 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import com.google.common.collect.ImmutableList;
+import com.google.errorprone.BugCheckerRefactoringTestHelper;
+import com.google.errorprone.BugCheckerRefactoringTestHelper.TestMode;
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+final class Slf4JLoggerDeclarationTest {
```
mohamedsamehsalah:
```suggestion
final class Slf4jLoggerDeclarationTest {
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLoggerDeclaration.java
```diff
@@ -0,0 +1,185 @@
...
+import com.sun.source.tree.ModifiersTree;
+import com.sun.source.tree.Tree.Kind;
+import com.sun.source.tree.VariableTree;
+import com.sun.tools.javac.code.Symbol;
+import java.util.EnumSet;
+import javax.inject.Inject;
+import javax.lang.model.element.Modifier;
+import tech.picnic.errorprone.utils.MoreASTHelpers;
+
+/** A {@link BugChecker} that flags non-canonical SLF4J logger declarations. */
```
mohamedsamehsalah, line 42: Do we usually mention in the docs that this checker accepts custom names?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLoggerDeclaration.java
```diff
@@ -0,0 +1,185 @@
...
+    Symbol symbol = ASTHelpers.getSymbol(receiver);
+    return symbol != null
+        && symbol.asType().tsym.equals(ASTHelpers.getSymbol(getEnclosingClass(state)));
+  }
+
+  private static ClassTree getEnclosingClass(VisitorState state) {
+    ClassTree clazz = state.findEnclosing(ClassTree.class);
+    // XXX: Review whether we should relax this constraint in the face of so-called anonymous
+    // classes. See
+    // https://docs.oracle.com/en/java/javase/23/language/implicitly-declared-classes-and-instance-main-methods.html
```
mohamedsamehsalah, line 181: TIL 💡

mohamedsamehsalah (commented):
Pushed the test rename. THanks @Stephan202 ❤️ 

Clean 🧹

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/Slf4jLoggerDeclaration.java
Stephan202: Based on `git grep -l 'this(ErrorProneFlags.empty())' | xargs grep -P -A7 '^/\*\*'`: generally not. (Maybe we should, but then I propose that we introduce a machine-readable mechanism for that, such that we can also document this on the website.)

rickie (approved):
Nice improvements!!

PR #788, opened 2023-09-11
Drop `ScheduledTransactionTrace` check

Stephan202:
Suggested commit message:
```
Drop `ScheduledTransactionTrace` check

As well as any other references to New Relic.
```

:exclamation: Before merging this, we should validate that Datadog does indeed collect traces for `@Scheduled` transactions out of the box.


rickie (approved):

mohamedsamehsalah (approved):
LGTM apart from the _validating that Datadog does indeed collect traces for @Scheduled transactions out of the box_ comment.

Stephan202:
> Before merging this, we should validate that Datadog does indeed collect traces for `@Scheduled` transactions out of the box.

While I didn't test this in production, the [agent source code](https://github.com/DataDog/dd-trace-java/blob/6e2751fb61b2f14cf553dff63df6bde725a35f14/dd-java-agent/instrumentation/spring-scheduling-3.1/src/test/groovy/SpringSchedulingTest.groovy) does have a test case for this. So I expect this'll work fine. If not, we'll review how to deal with that.

PR #794, opened 2023-09-17
Introduce `ConstantNaming` check

mohamedsamehsalah:
Suggested commit message:
```
Introduce `ConstantNaming` check (#794)

This check flags static constants that do not follow the upper snake
case naming convention.

While there, introduce the `Flags#getSet` utility method, and use it to
replace `Flags#getList` in relevant places.
```


pom.xml
```diff
@@ -1732,6 +1732,7 @@
                                     readable than the alternative. -->
                                     -Xep:YodaCondition:OFF
                                     -XepOpt:CheckReturnValue:CheckAllConstructors=true
+                                    -XepOpt:CanonicalConstantFieldName:ExcludedConstantFliedNames=serialVersionUID
```
mohamedsamehsalah, line 1735: Alternative is to follow the convention for this variable obviously 😅

mohamedsamehsalah (commented):
In addition to class variables within the same compilation unit, should we also match on constants imported from other classes ?
Ideally, the checker will suggest the fix in that class 🤔 

Which raises the other question, do we really need to suggest a fix for the other classes referencing the constants in `this` compilation unit ?

@Stephan202 suggested the use of `TreeScanner` as in [this](https://github.com/PicnicSupermarket/error-prone-support/pull/775/files#diff-d243187cd87cb51bc8283dc88945daff74442b8fd683a7f8b6b377b34ff7869dR113-R121) example but I could not seem to make this work 😓 

Help appreciated 🙏

pom.xml
Stephan202: I think this should be a default exclusion; otherwise _all_ users will have to make this change (`serialVersionUID` cannot be renamed: Java requires it to be spelled exactly like this).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,114 @@
...
+    summary =
+        "Make sure that all constant variables follow the `UPPER_SNAKE_CASE` naming convention.",
+    link = BUG_PATTERNS_BASE_URL + "CanonicalConstantFieldName",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = LIKELY_ERROR)
+public final class CanonicalConstantFieldName extends BugChecker implements VariableTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<Tree> IS_CONSTANT =
+      allOf(hasModifier(Modifier.STATIC), hasModifier(Modifier.FINAL));
```
Stephan202, line 61: We should (at least by default) restrict this check to private variables; otherwise the rename is not safe.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,114 @@
...
+    if (!IS_CONSTANT.matches(tree, state)) {
+      return Description.NO_MATCH;
+    }
+    SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
+
+    VarSymbol variableSymbol = ASTHelpers.getSymbol(tree);
+    String variableName = variableSymbol.getSimpleName().toString();
+
+    if (!isVariableUpperSnakeCase(variableName) && !isVariableNameExcluded(variableName)) {
+      fixBuilder.merge(SuggestedFixes.renameVariable(tree, toUpperSnakeCase(variableName), state));
```
Stephan202, line 93: IIUC this change will suggest many distinct fixes. It might work in practice, but it'll be nicer if there's a single report against the field, which, when applied, renames all references to that `Symbol`. (That's why offline I suggested using a `TreeScanner` :)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldNameTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.jupiter.api.Test;
+
+final class CanonicalConstantFieldNameTest {
+  @Test
+  void replacement() {
+    BugCheckerRefactoringTestHelper.newInstance(CanonicalConstantFieldName.class, getClass())
+        .addInputLines(
+            "A.java",
+            "class A {",
+            "  private static final int number = 1;",
```
Stephan202, line 14: What if we have a field named `number` and another field already named `NUMBER`? :upside_down_face:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,114 @@
...
+  private static boolean isVariableUpperSnakeCase(String variableName) {
+    return variableName.equals(toUpperSnakeCase(variableName));
+  }
+
+  private boolean isVariableNameExcluded(String variableName) {
+    return excludedConstantFliedNames.contains(variableName);
+  }
+
+  private static String toUpperSnakeCase(String variableName) {
+    return variableName.replaceAll("([a-z])([A-Z])", "$1_$2").toUpperCase(Locale.ROOT);
```
Stephan202, line 108: It's generally preferred to compile patterns only once. So let's introduce a `Pattern` constant for this.

Stephan202 (commented):
Tnx for working on this @mohamedsamehsalah; it'll allow for some nice cleanup I expect. Did leave some comments based on a quick skim :)

Stephan202:
(I didn't check the comments before reviewing; sorry.)

> In addition to class variables within the same compilation unit, should we also match on constants imported from other classes ?
> Ideally, the checker will suggest the fix in that class 🤔

Error Prone doesn't support this; for cases like that OpenRewrite would be better suited. (But for now, let's just focus on the private fields.)

> @Stephan202 suggested the use of TreeScanner as in [this](https://github.com/PicnicSupermarket/error-prone-support/pull/775/files#diff-d243187cd87cb51bc8283dc88945daff74442b8fd683a7f8b6b377b34ff7869dR113-R121) example but I could not seem to make this work 😓

Anywhere we can see the code that doesn't work; perhaps I can take a look.

mohamedsamehsalah:
@Stephan202 I covered your first pass comments, however, I still could not make use of the `TreeScanner`. Or in other words, I couldn't find what advantage does the `TreeScanner` give that the `TreeMatcher` does not.

AFAIU, the suggested fix builder used in this PR will execute what you are suggesting in this comment:
> IIUC this change will suggest many distinct fixes. It might work in practice, but it'll be nicer if there's a single report against the field, which, when applied, renames all references to that Symbol. (That's why offline I suggested using a TreeScanner :)

Happy to hear your thoughts (again? 😅) 🙏

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,180 @@
...
+        VarSymbol variableSymbol = ASTHelpers.getSymbol(variableTree);
+        if (variableSymbol == null) {
+          return Description.NO_MATCH;
+        }
+
+        String variableName = variableSymbol.getSimpleName().toString();
+        if (!isVariableUpperSnakeCase(variableName) && !isVariableNameExcluded(variableName)) {
+          String replacement = toUpperSnakeCase(variableName);
+
+          if (!classVariables.contains(replacement)) {
```
Stephan202, line 133: IIUC, here we avoid renaming a variable if the target name is already declared in the exact same scope. Due to the global search and replace that we do, I suspect that any usage of the target name in this `CompilationUnit` could cause ambiguity, shadowing or confusion. So for this case, too, a `TreeScanner` may be employed.

Stephan202 (commented):
> @Stephan202 I covered your first pass comments, however, I still could not make use of the `TreeScanner`. Or in other words, I couldn't find what advantage does the `TreeScanner` give that the `TreeMatcher` does not.

Actually, you _are_ using a `TreeScanner` now, just through a method I didn't know existed: have a close look at [`SuggestedFixes.renameVariable`](https://github.com/google/error-prone/blob/11471840328880711e58d009b0d4a29517620d16/check_api/src/main/java/com/google/errorprone/fixes/SuggestedFixes.java#L702-L744) :eyes: :smile: This should work!

Before I dive into a more thorough review: would you like to have a crack at resolving the surviving mutants?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
mohamedsamehsalah: > would you like to have a crack at resolving the surviving mutants?

Yes, sir 🖖 

> IIUC, here we avoid renaming a variable if the target name is already declared in the exact same scope. Due to the global search and replace that we do, I suspect that any usage of the target name in this CompilationUnit could cause ambiguity, shadowing or confusion. So for this case, too, a TreeScanner may be employed.

Will look into this 👀 

Thanks.

mohamedsamehsalah (commented):
Sorry for the slow progress on this @Stephan202 , PTAL 🙏

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,184 @@
...
+                    "Suggested fix for constant name conflicts with an already defined variable `%s`.",
+                    replacement))
+            .build());
+  }
+
+  private boolean isFieldAccessModifierApplicable(VariableTree tree, VisitorState state) {
+    return includePublicConstantFieldNames || IS_PRIVATE.matches(tree, state);
+  }
+
+  private static boolean isVariableUpperSnakeCase(String variableName) {
```
rickie: The method name does not have something to do with `variable`, so we can omit that part :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * A {@link BugChecker} that warns and suggests the fix when a constant variable does not follow the
+ * upper snake case naming convention.
+ *
+ * <p>Example:
+ *
+ * <p>This checker will re-write the following variables with all its references:
+ *
+ * <ul>
```
rickie: Let's use ` <pre>{@code` and `simpleNumber` to show that it is rewritten upper snake case style :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,184 @@
...
+import com.sun.tools.javac.code.Symbol.VarSymbol;
+import java.util.Locale;
+import java.util.regex.Pattern;
+import javax.inject.Inject;
+import javax.lang.model.element.Modifier;
+import org.jspecify.annotations.Nullable;
+import tech.picnic.errorprone.bugpatterns.util.Flags;
+
+/**
+ * A {@link BugChecker} that warns and suggests the fix when a constant variable does not follow the
```
rickie: Pushing a small suggestion.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * <p>To the following:
+ *
+ * <ul>
+ *   <li>private static final int NUMBER = 1;
+ * </ul>
+ *
+ * @apiNote This {@link BugChecker checker} has two optional flags:
+ *     <ul>
+ *       <li>ExcludedConstantFliedNames: A list of field names to exclude from this check.
```
rickie:
```suggestion
 *       <li>`ExcludedConstantFieldNames`: A list of field names to exclude from this check.
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,184 @@
...
+ *       <li>IncludePublicConstantFields: Whether to include public constants when running this
+ *           check.
+ *     </ul>
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary =
+        "Make sure that all constant variables follow the `UPPER_SNAKE_CASE` naming convention.",
```
rickie:
```suggestion
        "Make sure that all constant variables follow the `UPPER_SNAKE_CASE` naming convention",
```
Pushing a small suggestion :)

rickie (approved):
I had 15 spare minutes 😄 so did some small tweaks.

This is going in the right direction! Very nice man. More in depth review soon to follow.

mohamedsamehsalah:
> I had 15 spare minutes 😄 so did some small tweaks.
> 
> This is going in the right direction! Very nice man. More in depth review soon to follow.

@rickie tweaks LGTM, Thanks ❤️

rickie:
I see I approved, that was not my idea yet haha sorry for that.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * A {@link BugChecker} that flags constant variables that do not follow the upper snake case naming
+ * convention.
+ *
+ * <p>This check will rewrite the following variables with all its references:
+ *
+ * <pre>{@code
+ * private static final int simpleNumber = 1;
+ *
```
rickie:
```suggestion
```
This is my own fault IIRC haha.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldNameTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @Test
+  void replacement() {
+    BugCheckerRefactoringTestHelper.newInstance(CanonicalConstantFieldName.class, getClass())
+        .addInputLines(
+            "A.java",
+            "class A {",
+            "  private static final int number = 1;",
+            "",
+            "  int nonConstantNumber = 2;",
```
rickie, line 43: We can move these up to the `identification` test as well :).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldNameTest.java
```diff
@@ -0,0 +1,156 @@
...
+            "",
+            "  private static final int foo = 3;",
+            "}")
+        .doTest(TestMode.TEXT_MATCH);
+  }
+
+  @Test
+  void includePublicConstants() {
+    BugCheckerRefactoringTestHelper.newInstance(CanonicalConstantFieldName.class, getClass())
+        .setArgs("-XepOpt:CanonicalConstantFieldName:IncludePublicConstantFields=true")
```
rickie, line 139: Hmm, interesting. This can lead to quite some unexpected situations right 👀 ? As we cannot rename usages of this in other files, so I'm not sure if we should do this 🤔.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldNameTest.java
```diff
@@ -0,0 +1,156 @@
...
+import com.google.errorprone.BugCheckerRefactoringTestHelper.TestMode;
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+final class CanonicalConstantFieldNameTest {
+  @Test
+  void identification() {
+    CompilationTestHelper.newInstance(CanonicalConstantFieldName.class, getClass())
+        .addSourceLines(
+            "A.java",
```
rickie, line 13: The reporting here is a bit weird. It's reported on a class level... and only one error per time while we are creating a builder with multiple fixes... I'm not sure what I'm doing wrong or missing here.
 
It should be:
```
"  // BUG: Diagnostic contains:",
"  private static final int foo = 1;",
"  // BUG: Diagnostic contains:",
"  private static final int foobar = 1;",
"  // BUG: Diagnostic contains:",
"  private static final int foobarbaz = 1;",
```

Later we should drop the `foobar` and `foobarbaz` but now it's here to show what is wrong and we should figure out.

rickie (commented):
Made some changes and found some interesting cases 👀.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldNameTest.java
rickie: So it's reported on the class file... It could be that this is a bug in the SuggestedFixes?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
rickie: Yeah let's start with private variables and add an XXX to consider protected and public ones later as well.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldNameTest.java
mohamedsamehsalah: We have a _replacement_ test for this situation; where we rename a variabled referenced in another file. 
See
```
  int referenceToNumberFromAnotherClass = B.numberFromAnotherClass;
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldNameTest.java
mohamedsamehsalah: > It could be that this is a bug in the SuggestedFixes?

I don't think it is a bug, since we explicitly passing the class tree to the `describeMatch` method. If we want to improve on this, I think we can use a TreeScanner. I will look into this 🙏

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldNameTest.java
mohamedsamehsalah: I have a test for this where we replace with suggested fix the variable imported from another class.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldNameTest.java
mohamedsamehsalah: Pushed something 👍

mohamedsamehsalah:
Thanks for review @rickie 🙏 

I pushed a commit to address your comments.

However, I get this dependency analysis error from maven
```
[ERROR] Used undeclared dependencies found:
[ERROR]    org.jetbrains:annotations:jar:24.1.0:provided
```

Any clue why ?

rickie (commented):
As a result of dropping, we can simplify the code, pushing a suggestion!

mohamedsamehsalah:
Thanks @rickie , changes LGMT 🚀

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,153 @@
...
+   */
+  @Inject
+  CanonicalConstantFieldName(ErrorProneFlags flags) {
+    optionalExcludedConstantFieldNames = getAllowedFieldNames(flags);
+  }
+
+  @Override
+  public Description matchVariable(VariableTree tree, VisitorState state) {
+    String variableName = tree.getName().toString();
+    if (IS_CONSTANT.matches(tree, state)
```
rickie: If we flip the conditions here and do an early return, we have one level of indentation less. Additionally, we generally try to have early returns if possible :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,153 @@
...
+        reportConstantRenameBlocker(tree, replacement, state);
+      }
+
+      return describeMatch(tree, fixBuilder.build());
+    }
+
+    return Description.NO_MATCH;
+  }
+
+  private static ImmutableList<VariableTree> getVariablesInCompilationUnit(
```
rickie: Changed the order of static methods in the class based on which one is used first :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
```diff
@@ -0,0 +1,153 @@
...
+    String variableName = tree.getName().toString();
+    if (IS_CONSTANT.matches(tree, state)
+        && !isUpperSnakeCase(variableName)
+        && !isVariableNameExcluded(variableName)) {
+      SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
+
+      ImmutableList<VariableTree> variablesInCompilationUnit =
+          getVariablesInCompilationUnit(state.getPath().getCompilationUnit());
+      String replacement = toUpperSnakeCase(variableName);
+      if (variablesInCompilationUnit.stream()
```
rickie: Just a suggestion but what if we extract this to a method, I think it reads nicely. Additionally if we use `anyMatch` instead of `noneMatch` it reads as "isVariableNameInUse`, instead of the negated variant, could be a bit easier to understand.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldNameTest.java
rickie: Added a few more tests :).

rickie (approved):
Hey @mohamedsamehsalah , thanks a lot for pushing this! It looks really nice and am curious to see this getting rolled out on Picnic's codebase 🔥 !!!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantFieldName.java
mohamedsamehsalah: More readable, I agree 💯

mohamedsamehsalah:
Thanks for suggestions @rickie . LGTM 🚀

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantNaming.java
```diff
@@ -0,0 +1,157 @@
...
+ * private static final int SIMPLE_NUMBER = 1;
+ * }</pre>
+ *
+ * @apiNote This check has an optional flag `AllowedConstantNames` that represents a list of field
+ *     names to exclude from this check.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Constant variables should adhere to the `UPPER_SNAKE_CASE` naming convention",
+    link = BUG_PATTERNS_BASE_URL + "CanonicalConstantNaming",
```
Stephan202: I think `ConstantNaming` or `ConstantName` would be a better name for this check; the word "canonical" has a different meaning in the context of the other two checks named `Canonical*`. (We could say that it's about _idiomatic_ naming, but that applies to quite some other checks as well.)

Between `ConstantNaming` and `ConstantName` I can make an argument either way; we're not consistent in this respect. Will use the former for now.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalConstantNaming.java
```diff
@@ -0,0 +1,157 @@
...
+import com.sun.source.util.TreeScanner;
+import java.util.Locale;
+import java.util.regex.Pattern;
+import javax.inject.Inject;
+import javax.lang.model.element.Modifier;
+import org.jspecify.annotations.Nullable;
+import tech.picnic.errorprone.utils.Flags;
+
+/**
+ * A {@link BugChecker} that flags constant variables that do not follow the upper snake case naming
```
Stephan202:
```suggestion
 * A {@link BugChecker} that flags static constants that do not follow the upper snake case naming
```

Stephan202 (approved):
Rebased and added a commit. Very nice check @mohamedsamehsalah!

Suggested commit message:
```
Introduce `ConstantNaming` check (#794)

This check flags static constants that do not follow the upper snake
case naming convention.

While there, introduce the `Flags#getSet` utility method, and use it to
replace `Flags#getList` in relevant places.
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ConstantNaming.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+   * Instantiates a customized {@link ConstantNaming}.
+   *
+   * @param flags Any provided command line flags.
+   */
+  @Inject
+  ConstantNaming(ErrorProneFlags flags) {
+    exemptedNames =
+        Sets.union(DEFAULT_EXEMPTED_NAMES, Flags.getSet(flags, ADDITIONAL_EXEMPTED_NAMES_FLAG))
```
Stephan202, line 74: Here and below: Pitest correctly flags that the `Sets#union` arguments can be swapped without causing a test failure. That's expected, as set union is a symmetric operation, and this code doesn't care about the created set's iteration order.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ConstantNaming.java
```diff
@@ -0,0 +1,125 @@
...
+  private static String toUpperSnakeCase(String variableName) {
+    return SNAKE_CASE.matcher(variableName).replaceAll("$1_$2").toUpperCase(Locale.ROOT);
+  }
+
+  private static boolean isVariableNameInUse(String name, VisitorState state) {
+    return Boolean.TRUE.equals(
+        new TreeScanner<Boolean, @Nullable Void>() {
+          @Override
+          public Boolean visitVariable(VariableTree tree, @Nullable Void unused) {
+            return ASTHelpers.getSymbol(tree).getSimpleName().toString().equals(name)
```
Stephan202, line 115: Can be simplified to
```suggestion
            return ASTHelpers.getSymbol(tree).contentEquals(name);
```

I noticed a similar discussion [here](https://github.com/PicnicSupermarket/error-prone-support/pull/783#discussion_r1502274717), so decided to create a Refaster rule: #1379.

ben-manes:
I'm not sure what your style guide is, but fwiw Google's is popular and does not follow this rule (e.g. `logger` not `LOGGER` as this suggests). Of course not an issue as trivial to suppress or disable, just fyi that this may be considered overly broad as  broaches into style vs quality. 

https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

Stephan202:
Thanks for sharing @ben-manes! I suppose this one of the (likely few) ways in which inside Picnic we deviate from the Google style guide. In version 0.19.0 we also landed #783, which by default will suggest `LOG` for SL4FJ loggers. In that case we support an override (though while still going UPPER_SNAKE_CASE) for static loggers). For this one a flag would have to unlock significantly more advanced heuristics, and it'd never be perfect. Indeed disabling may be best for some (most?) users. :thinking:

PR #810, opened 2023-10-04
Introduce `{Mono,Flux}OfType` Refaster rules

werli:
Since semantically this should be equivalent.

Suggested commit message:
```
Introduce `{Mono,Flux}OfType` Refaster rules (#810)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1053,6 +1053,40 @@ final class ReactorRules {
     }
   }
 
+  /** Prefer {@link Mono#ofType(Class)} over more contrived alternatives. */
+  static final class MonoOfType<T, S> {
+    @BeforeTemplate
+    Mono<S> before(Mono<T> mono) {
+      return mono.filter(
+              Refaster.anyOf(
+                  Refaster.<S>clazz()::isInstance,
+                  x -> Refaster.<S>clazz().isAssignableFrom(x.getClass())))
+          .cast(Refaster.<S>clazz());
+    }
+
+    @AfterTemplate
+    Mono<S> after(Mono<T> mono) {
+      return mono.ofType(Refaster.<S>clazz());
+    }
+  }
+
+  /** Prefer {@link Flux#ofType(Class)} over more contrived alternatives. */
+  static final class FluxOfType<T, S> {
+    @BeforeTemplate
+    Flux<S> before(Flux<T> flux) {
+      return flux.filter(
+              Refaster.anyOf(
+                  Refaster.<S>clazz()::isInstance,
+                  x -> Refaster.<S>clazz().isAssignableFrom(x.getClass())))
+          .cast(Refaster.<S>clazz());
+    }
+
+    @AfterTemplate
+    Flux<S> after(Flux<T> flux) {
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -794,6 +794,40 @@ Flux<S> after(Flux<T> flux) {
...
   }
 
+  /** Prefer {@link Mono#ofType(Class)} over more contrived alternatives. */
+  static final class MonoOfType<T, S> {
+    @BeforeTemplate
+    Mono<S> before(Mono<T> mono) {
+      return mono.filter(
+              Refaster.anyOf(
+                  Refaster.<S>clazz()::isInstance,
+                  x -> Refaster.<S>clazz().isAssignableFrom(x.getClass())))
```
Stephan202, line 804: We can introduce a separate rule to prefer the first over the second. (I have this change locally already, but need to leave the train now :sweat_smile:.)

Stephan202 (commented):
Nice! Will finalize review later.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: I filed #811 for this.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -794,6 +794,40 @@ Flux<S> after(Flux<T> flux) {
     }
   }
 
+  /** Prefer {@link Mono#ofType(Class)} over more contrived alternatives. */
+  static final class MonoOfType<T, S> {
+    @BeforeTemplate
+    Mono<S> before(Mono<T> mono) {
+      return mono.filter(
+              Refaster.anyOf(
+                  Refaster.<S>clazz()::isInstance,
```
Stephan202: Instead of `Refaster.<S>clazz()` we can use a `Class<S> clazz` parameter, in which case non-literal class references are also captured.

Stephan202 (approved):
Rebased the PR on top of #811 and added a commit. Very nice addition @werli!

werli (commented):
The latest changes LGTM, thanks @Stephan202 :muscle:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -794,6 +794,32 @@ Flux<S> after(Flux<T> flux) {
...
+  /** Prefer {@link Mono#ofType(Class)} over more contrived alternatives. */
+  static final class MonoOfType<T, S> {
+    @BeforeTemplate
+    Mono<S> before(Mono<T> mono, Class<S> clazz) {
+      return mono.filter(clazz::isInstance).cast(clazz);
+    }
+
+    @AfterTemplate
+    Mono<S> after(Mono<T> mono, Class<S> clazz) {
+      return mono.ofType(clazz);
```
Venorcis, line 806: nice, didn't know this existed, will rewrite some uses in our code 😄 🚀

Venorcis (approved):

rickie (approved):
Rebased onto `master`! Looks good 🚀

PR #811, opened 2023-10-04
Introduce `ClassRules` Refaster rule collection

Stephan202:
Proposal based on [this comment](https://github.com/PicnicSupermarket/error-prone-support/pull/810#discussion_r1346203628).

Suggested commit message:
```
Introduce `ClassRules` Refaster rule collection (#811)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ClassRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ClassRules.java
@@ -1,0 +1,65 @@
+package tech.picnic.errorprone.refasterrules;
+
+import com.google.errorprone.refaster.Refaster;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.util.function.Predicate;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
+
+/** Refaster rules related to expressions dealing with classes. */
+@OnlineDocumentation
+final class ClassRules {
+  private ClassRules() {}
+
+  /** Prefer {@link Class#isInstance(Object)} over more contrived alternatives. */
+  static final class ClassIsInstance<T, S> {
+    @BeforeTemplate
+    boolean before(Class<T> clazz, S object) {
+      return clazz.isAssignableFrom(object.getClass());
+    }
+
+    @AfterTemplate
+    boolean after(Class<T> clazz, S object) {
+      return clazz.isInstance(object);
+    }
+  }
+
+  /** Prefer using the {@code instanceof} keyword over less idiomatic alternatives. */
+  static final class Instanceof<T, S> {
+    @BeforeTemplate
+    boolean before(S object) {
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ClassRules.java
```diff
@@ -0,0 +1,65 @@
...
+    @BeforeTemplate
+    boolean before(S object) {
+      return Refaster.<T>clazz().isInstance(object);
+    }
+
+    @AfterTemplate
+    boolean after(S object) {
+      return Refaster.<T>isInstance(object);
+    }
+  }
```
Stephan202, line 38: These rules are split out, because the second rewrite is possible only for literal class references.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ClassRules.java
```diff
@@ -0,0 +1,65 @@
...
+    @BeforeTemplate
+    Predicate<S> before(Class<T> clazz) {
+      return o -> clazz.isInstance(o);
+    }
+
+    @AfterTemplate
+    Predicate<S> after(Class<T> clazz) {
+      return clazz::isInstance;
+    }
+  }
```
Stephan202, line 64: Hmm, I just realized I don't need to split these out; will add another commit.

werli (approved):
Nice!

Venorcis (approved):
LGTM 👍🏻

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ClassRulesTestInput.java
```diff
@@ -0,0 +1,25 @@
...
+import tech.picnic.errorprone.refaster.test.RefasterRuleCollectionTestCase;
+
+final class ClassRulesTest implements RefasterRuleCollectionTestCase {
+  boolean testClassIsInstance() throws IOException {
+    return CharSequence.class.isAssignableFrom("foo".getClass());
+  }
+
+  ImmutableSet<Boolean> testInstanceof() throws IOException {
+    Class<?> clazz = CharSequence.class;
+    return ImmutableSet.of(CharSequence.class.isInstance("foo"), clazz.isInstance("bar"));
```
rickie, line 14: Hmm, that means we can actually drop this second test case right? What's the reason for keeping it?

Stephan202: We could, but I wanted to be explicit about covering both cases. We could keep _just_ the second, if you prefer, cause that one is more generic (but that's only clear in the context of the other rules).

rickie: AHh okay clear, let's keep it 😄.

rickie (approved):

rickie:
Rebased and tweaked suggested commit message.

Issue #655, opened 2023-06-01
It should be possible to change a severity level per rule


vprudnikov:
### Problem

I would like to make some suggestions required for my team by changing the default severity level for them.

### Description of the proposed new feature

- [ ] Avoid a common gotcha, or potential problem.

It would be nice to provide a compiler argument that changes the severity for one rule, for example:
`-XepOpt:Refaster:StreamRules:StreamMapFirst:ERROR`

Stephan202:
Thanks @vprudnikov; this is a nice idea. We'll have to assess which flag/config format works best, but technically this looks very doable :+1:.

(Before making a decision on the flag support we should perhaps assess [to what extent repeated flags can be supported](https://github.com/PicnicSupermarket/error-prone-support/issues/654#issuecomment-1571949284), as that may influence the design space.)

vprudnikov:
thanks @Stephan202! I admit that my example is not that good, however, I hope it outlines the idea :)
And I agree that it would be less confusing if I could repeat the same flag as many times as I want.

jai-ramakrishnan:
I am trying to disable all checks and selectively turn on Raster rules 1 at a time. So passing -XepDisableAllChecks takes care of the former. Now I need to enable say all NullRules https://error-prone.picnic.tech/refasterrules/NullRules/
Is there a way to achieve that with the current setup?

Stephan202:
Hey @jai-ramakrishnan! For this you'd add `-Xep:Refaster:OFF -XepOpt:Refaster:NamePattern=^NullRules\$.*`

(The latter flag is a regex, so later you could e.g. do `-XepOpt:Refaster:NamePattern=^(MapRules|NullRules)\$.*`. Once you're ready to enable all-but-some rules, you can instead start listing the rules to _exclude_. E.g. using `-XepOpt:Refaster:NamePattern=^(?!XRules\$)(?!YRules\$).*`)

jai-ramakrishnan:
Thanks a bunch @Stephan202. I can get it working partially with the negate pattern.

[...]

Issue #654, opened 2023-06-01
Can't disable rule using compiler argument


vprudnikov:
### Minimal Reproducible Example

Here is my configuration in pom.xml

```xml
...
        <error-prone.version>2.19.1</error-prone.version>
        <error-prone-slf4j.version>0.1.18</error-prone-slf4j.version>
        <error-prone-support.version>0.11.1</error-prone-support.version>
...
                <plugin>
                    <groupId>org.apache.maven.plugins</groupId>
                    <artifactId>maven-compiler-plugin</artifactId>
                    <version>${maven-compiler-plugin.version}</version>
                    <configuration>
                        <source>${java.version}</source>
                        <target>${java.version}</target>
                        <encoding>${project.build.sourceEncoding}</encoding>
                        <compilerArgs>
                            <arg>-XDcompilePolicy=simple</arg>
                            <!---DisableWarningsInGeneratedCode because it fails with Lombok-->
                            <!---Disable CheckArgumentWithMessage because we already check arguments with Lombok-->
                            <!---Disable ImmutableMapOf because we prefer using standard Java library even Map.of() looks confusing since it returns an immutable map-->
                            <arg>
                                -Xplugin:ErrorProne
                                -XepDisableWarningsInGeneratedCode
                                -XepOpt:Refaster:NamePattern=^(?!PreconditionsRules\$CheckArgumentWithMessage).*
                                -XepOpt:Refaster:NamePattern=^(?!ImmutableMapRules\$ImmutableMapOf).*
                                -XepOpt:Refaster:NamePattern=^(?!ImmutableMapRules\$ImmutableMapOf1).*
                                -XepOpt:Refaster:NamePattern=^(?!ImmutableMapRules\$ImmutableMapOf2).*
                                -XepOpt:Refaster:NamePattern=^(?!ImmutableMapRules\$ImmutableMapOf3).*
                                -XepOpt:Refaster:NamePattern=^(?!ImmutableMapRules\$ImmutableMapOf4).*
                                -XepOpt:Refaster:NamePattern=^(?!ImmutableMapRules\$ImmutableMapOf5).*
                            </arg>
                        </compilerArgs>
                        <annotationProcessorPaths>
                            <!-- Error Prone. More info: https://errorprone.info/index -->
                            <path>
                                <groupId>com.google.errorprone</groupId>
                                <artifactId>error_prone_core</artifactId>
                                <version>${error-prone.version}</version>
                            </path>
[...]

PR #812, opened 2023-10-06
Introduce assorted `BigDecimal#signum` Refaster rules

Venorcis:
Expands the `BigDecimalRules`, mainly preventing contrived uses of `compareTo(BigDecimal.ZERO)` by using `signum()` instead. 
Also adds some rules surrounding the use of `signum()`.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
@@ -68,8 +68,8 @@ final class BigDecimalRules {
     }
   }
 
-  /** Prefer using {@link BigDecimal#signum()} over comparing to {@link BigDecimal#ZERO}. */
-  static final class BigDecimalCompareToZero {
+  /** Prefer using {@link BigDecimal#signum()} over more contrived alternatives. */
+  static final class BigDecimalSignumIsZero {
     @BeforeTemplate
     boolean before(BigDecimal value) {
       return Refaster.anyOf(
@@ -83,30 +83,38 @@ final class BigDecimalRules {
     }
   }
 
-  /** Prefer using {@link BigDecimal#signum()} over more contrived alternatives. */
-  static final class BigDecimalCompareToZeroPositive {
+  /**
+   * Prefer a {@link BigDecimal#signum()} comparison to 0 over more contrived or less clear
+   * alternatives.
+   */
+  static final class BigDecimalSignumIsPositive {
     @BeforeTemplate
     boolean before(BigDecimal value) {
       return Refaster.anyOf(
           value.compareTo(BigDecimal.ZERO) > 0,
           BigDecimal.ZERO.compareTo(value) < 0,
-          value.signum() > 0,
-          value.signum() >= 1,
-          value.signum() != -1);
+          value.signum() == 1,
+          value.signum() >= 1);
     }
 
     @AfterTemplate
     boolean after(BigDecimal value) {
-      return value.signum() == 1;
+      return value.signum() > 0;
     }
   }
 
-  /** Prefer using {@link BigDecimal#signum()} over comparing to {@link BigDecimal#ZERO}. */
-  static final class BigDecimalCompareToZeroPositiveInclusive {
+  /**
+   * Prefer a {@link BigDecimal#signum()} comparison to 0 over more contrived or less clear
+   * alternatives.
+   */
+  static final class BigDecimalSignumIsNonnegative {
     @BeforeTemplate
     boolean before(BigDecimal value) {
       return Refaster.anyOf(
-          value.compareTo(BigDecimal.ZERO) >= 0, BigDecimal.ZERO.compareTo(value) <= 0);
+          value.compareTo(BigDecimal.ZERO) >= 0,
+          BigDecimal.ZERO.compareTo(value) <= 0,
+          value.signum() != -1,
+          value.signum() > -1);
     }
 
     @AfterTemplate
@@ -115,44 +123,43 @@ final class BigDecimalRules {
     }
   }
 
-  /** Prefer using {@link BigDecimal#signum()} over comparing to {@link BigDecimal#ZERO}. */
-  static final class BigDecimalCompareToZeroNegative {
+  /**
+   * Prefer a {@link BigDecimal#signum()} comparison to 0 over more contrived or less clear
+   * alternatives.
+   */
+  static final class BigDecimalSignumIsNegative {
     @BeforeTemplate
     boolean before(BigDecimal value) {
       return Refaster.anyOf(
-          value.compareTo(BigDecimal.ZERO) < 0, BigDecimal.ZERO.compareTo(value) > 0);
+          value.compareTo(BigDecimal.ZERO) < 0,
+          BigDecimal.ZERO.compareTo(value) > 0,
+          value.signum() == -1,
+          value.signum() <= -1);
     }
 
     @AfterTemplate
     boolean after(BigDecimal value) {
-      return value.signum() == -1;
+      return value.signum() < 0;
     }
   }
 
-  /** Prefer using {@link BigDecimal#signum()} over comparing to {@link BigDecimal#ZERO}. */
-  static final class BigDecimalCompareToZeroNegativeInclusive {
+  /**
+   * Prefer a {@link BigDecimal#signum()} comparison to 0 over more contrived or less clear
+   * alternatives.
+   */
+  static final class BigDecimalSignumIsNonpositive {
     @BeforeTemplate
     boolean before(BigDecimal value) {
       return Refaster.anyOf(
-          value.compareTo(BigDecimal.ZERO) <= 0, BigDecimal.ZERO.compareTo(value) >= 0);
+          value.compareTo(BigDecimal.ZERO) <= 0,
+          BigDecimal.ZERO.compareTo(value) >= 0,
+          value.signum() != 1,
+          value.signum() < 1);
     }
 
     @AfterTemplate
     boolean after(BigDecimal value) {
       return value.signum() <= 0;
-    }
-  }
-
-  /** {@link BigDecimal#signum()} does not produce values lower than {@code -1}. */
-  static final class BigDecimalNegativeSignum {
-    @BeforeTemplate
-    boolean before(BigDecimal value) {
-      return Refaster.anyOf(value.signum() < 0, value.signum() <= -1, value.signum() != 1);
-    }
-
-    @AfterTemplate
-    boolean after(BigDecimal value) {
-      return value.signum() == -1;
     }
   }
 }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
@@ -84,7 +84,7 @@ final class BigDecimalRules {
   }
 
   /**
-   * Prefer a {@link BigDecimal#signum()} comparison to 0 over more contrived or less clear
+   * Prefer a {@link BigDecimal#signum()} comparison to 1 over more contrived or less clear
    * alternatives.
    */
   static final class BigDecimalSignumIsPositive {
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
```diff
@@ -66,4 +66,114 @@ BigDecimal after(double value) {
...
+    }
+
+    @AfterTemplate
+    boolean after(BigDecimal value) {
+      return value.signum() == 0;
+    }
+  }
+
+  /** Prefer using {@link BigDecimal#signum()} over comparing to {@link BigDecimal#ZERO}. */
+  static final class BigDecimalCompareToNonZero {
```
rickie, line 85: Ah, we can use the `@AlsoNegation` and drop this template and get the same result :smile:.
We can move over the test cases though.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
```diff
@@ -66,4 +66,114 @@ BigDecimal after(double value) {
...
+  static final class BigDecimalCompareToZeroPositive {
+    @BeforeTemplate
+    boolean before(BigDecimal value) {
+      return Refaster.anyOf(
+          value.compareTo(BigDecimal.ZERO) > 0, BigDecimal.ZERO.compareTo(value) < 0);
+    }
+
+    @AfterTemplate
+    boolean after(BigDecimal value) {
+      return value.signum() == 1;
```
rickie, line 108: The `BigDecimalPositiveSignum` has the same after template. We can combine them in the `Refaster#anyOf` 😄.

rickie: In that case we also need to update the Javadoc :).

rickie (approved):
Added a commit with some suggestions :).

Suggested commit message (repeating title):
```
Introduce assorted `BigDecimal#signum` Refaster rules (#812)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
```diff
@@ -66,4 +67,92 @@ BigDecimal after(double value) {
...
+  static final class BigDecimalCompareToZeroNegative {
+    @BeforeTemplate
+    boolean before(BigDecimal value) {
+      return Refaster.anyOf(
+          value.compareTo(BigDecimal.ZERO) < 0, BigDecimal.ZERO.compareTo(value) > 0);
+    }
+
+    @AfterTemplate
+    boolean after(BigDecimal value) {
+      return value.signum() == -1;
```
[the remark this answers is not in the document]
Stephan202: (^ With this change we can also further collapse the rules using `@AfterTemplate`.)

Stephan202 (approved):
W.r.t. the incorrect cases: I have on a TODO list somewhere the idea to generate unit tests from these Refaster rules, so that this kind of thing would be automatically flagged.

Rebased and added two commits; lemme know what you think.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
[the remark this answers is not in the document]
Venorcis: > Instead of == 1 and == -1, shall we settle on > 0 and < 0 respectively? 

IMO `== 1` and `== -1` are more readable as meaning 'positive sign' and 'negative sign'. The `signum` function is also [mathematically defined](https://en.wikipedia.org/wiki/Sign_function) like that (unlike comparators):
image

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/BigDecimalRulesTestInput.java
```diff
@@ -20,4 +20,36 @@ ImmutableSet<BigDecimal> testBigDecimalTen() {
...
+
+  ImmutableSet<Boolean> testBigDecimalSignumIsNegative() {
+    return ImmutableSet.of(
+        BigDecimal.ONE.compareTo(BigDecimal.ZERO) < 0,
+        BigDecimal.ZERO.compareTo(BigDecimal.ONE) > 0,
+        BigDecimal.ZERO.signum() == -1,
+        BigDecimal.ZERO.signum() <= -1,
+        BigDecimal.ONE.compareTo(BigDecimal.ZERO) >= 0,
+        BigDecimal.ZERO.compareTo(BigDecimal.ONE) <= 0,
+        BigDecimal.ZERO.signum() != -1,
```
Venorcis, line 52:
```suggestion
        BigDecimal.ZERO.signum() != 1,
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
```diff
@@ -66,4 +67,92 @@ BigDecimal after(double value) {
...
+    }
+
+    @AfterTemplate
+    @AlsoNegation
+    boolean after(BigDecimal value) {
+      return value.signum() == 0;
+    }
+  }
+
+  /** Prefer using {@link BigDecimal#signum()} over more contrived alternatives. */
```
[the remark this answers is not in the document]
Venorcis: copy-pasted too hastily 🙈

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/BigDecimalRulesTestInput.java
Stephan202: This test case is for the negation :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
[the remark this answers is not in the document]
Stephan202: That's fair, but Java's type system still maps those numbers onto larger domain, with the alternative making clear how such won't-happen output would be treated. I.e., a trade-off between mathematical fact and programming pragmatics :shrug:

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/BigDecimalRulesTestInput.java
Venorcis: Guess I don't really follow these last 2 test cases (neither above), as the test titles seem to suggest these are variants of checking if a number is **negative**, but the last two actually do the reverse (and same above)

Stephan202: > as the test titles seem to suggest these are variants of checking if a number is *negative*, but the last two actually do the reverse (and the same applies above).

Correct; the second half of the test cases cover the implicitly-generated `@AlsoNegation` rules.

> Probably the `ZERO` and `ONE` don't really help here either.

Yeah, we generally try to vary the captured sub-expressions; lemme try to push something.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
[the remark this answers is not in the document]
Venorcis: I'm not sure if the type should be leading here; such reasoning could be extrapolated to mean you should be prepared for e.g. `Collection#size` returning something negative 😛 Perhaps this discussion signals that either `< 0` or `== -1` should be acceptable, but that other variants are not (and ofc. we still need to pick a default of how to rewrite those here).

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/BigDecimalRulesTestInput.java
Stephan202: Pushed something.

Venorcis: > the implicitly-generated @AlsoNegation rules

Ah right, that makes it a bit harder. Varying the numbers helps a bit though 👍🏻

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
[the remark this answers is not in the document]
Stephan202: > such reasoning could be extrapolated to mean you should be prepared for e.g. `Collection#size` returning something negative

Luckily we can [side step](https://github.com/PicnicSupermarket/error-prone-support/blob/da5eea832975cb9222cba51935f7a22b4ff35e46/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java#L32-L57) that particular case ;)

> Perhaps this discussion signals that either `< 0` or `== -1` should be acceptable, but that other variants are not (and ofc. we need to pick a default of how to rewrite those here).

Indeed, we could also drop some of the cases from the `@BeforeTemplate` body. Downside of _that_ is that we won't "clean up" mixed-use code-bases. (And it would still require agreement on what the `@AfterTemplate` should contain :sweat_smile:)

Perhaps this is a case where a flag could help, though right now Refaster doesn't support that (but the idea was also floated in #654 and #655), while a `BugChecker` approach would be _way_ more work.

Maybe I'm too a-mathematical on this point; will think about it. Also interested in @rickie's thoughts on this.

NB: Whatever we settle on: we can do the same for `{Integer,Long}#signum`. :smile:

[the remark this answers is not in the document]
Stephan202: Upon further reflection, the signum is often multiplied; an operation that crucially depends on the [-1, 0, 1] domain. That's an argument in favour of just embracing that fact. If we do, however, it does make sense to me that we _always_ compare to these values, so also in favour of the `>= 0` and `<= 0` alternatives proposed earlier.

I pushed a PR in that direction. WDYT?

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/BigDecimalRulesTestOutput.java
```diff
@@ -31,25 +31,25 @@ ImmutableSet<Boolean> testBigDecimalSignumIsZero() {
...
-        BigDecimal.valueOf(4).signum() > 0,
-        BigDecimal.valueOf(5).signum() <= 0,
-        BigDecimal.valueOf(6).signum() <= 0,
-        BigDecimal.valueOf(7).signum() <= 0,
-        BigDecimal.valueOf(8).signum() <= 0);
+        BigDecimal.valueOf(1).signum() == 1,
+        BigDecimal.valueOf(2).signum() == 1,
+        BigDecimal.valueOf(3).signum() == 1,
+        BigDecimal.valueOf(4).signum() == 1,
+        BigDecimal.valueOf(5).signum() != 1,
```
Venorcis, line 38: if I understand correctly how this works, IMO the `AlsoNegation` now doesn't make sense anymore (or will these `!= 1`s be rewritten in a second pass?)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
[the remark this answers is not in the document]
Venorcis: It was my original proposal, so SGTM 😄

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/BigDecimalRulesTestOutput.java
Stephan202: Nope, `!= 1` means "the value is non-positive", while `!= -1` means "the value is nonnegative" (i.e., what was earlier expressed as `<= 0` and `>= 0`, respectively).

(In my most recent commit I only swapped the "canonical presentation"; there's no semantic change. Good to keep in mind that for the second half of the test cases, the test method is misleading (which is unavailable due to our test framework).)

Venorcis: Ah right 👍🏻

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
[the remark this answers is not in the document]
Stephan202: > NB: Whatever we settle on: we can do the same for `{Integer,Long}#signum`. 😄

I filed #822 for this :)

Stephan202:
@rickie this PR changed significantly since you last looked at it; feel free to merge if you're happy with the latest changes.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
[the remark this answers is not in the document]
rickie: I like the argument @Venorcis made [here](https://github.com/PicnicSupermarket/error-prone-support/pull/812#discussion_r1349508629). But then you also came up with the multiplication argument that is strong.

Nice discussion 🚀 !

rickie (approved):

PR #813, opened 2023-10-06
Add another variant to the `StepVerifierLastStepVerifyErrorClass`

Venorcis:
Added `verifyErrorMatches(clazz::isInstance)` as a variant of `verifyErrorSatisfies(t -> assertThat(t).isInstanceOf(clazz))` that we already rewrite.

### DIFF
--- a/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
+++ b/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
@@ -520,7 +520,7 @@ final class ReactorRulesTest implements RefasterRuleCollectionTestCase {
         StepVerifier.create(Mono.empty())
             .verifyErrorMatches(IllegalStateException.class::isInstance),
         StepVerifier.create(Mono.empty())
-            .verifyErrorSatisfies(t -> assertThat(t).isInstanceOf(IllegalStateException.class)));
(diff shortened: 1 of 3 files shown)


rickie (approved):
Nice @Venorcis 😄 ! 

Suggested commit message:
```
Extend `StepVerifierLastStepVerifyErrorClass` Refaster rule (#813)
```

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -440,6 +440,8 @@ Duration testStepVerifierLastStepVerifyError() {
   ImmutableSet<Duration> testStepVerifierLastStepVerifyErrorClass() {
     return ImmutableSet.of(
         StepVerifier.create(Mono.empty()).expectError(IllegalArgumentException.class).verify(),
+        StepVerifier.create(Mono.empty())
+            .verifyErrorMatches(IllegalStateException.class::isInstance),
         StepVerifier.create(Mono.empty())
             .verifyErrorSatisfies(t -> assertThat(t).isInstanceOf(IllegalStateException.class)));
```
Stephan202: For consistency with other test cases, let's vary this exception type. I see that elsewhere in this class we use `AssertionError` as a third type.

Stephan202 (approved):
Rebased and added a tiny commit. This'll work well together with #811 :).

PR #814, opened 2023-10-06
Introduce `ImmutableList.of().stream()` Refaster rules

Venorcis:
Mainly occurs in tests, but making a list for streaming only is unnecessary (and more verbose) 🤷🏻‍♂️


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ImmutableListRules.java
```diff
@@ -263,4 +263,84 @@ ImmutableList<T> after(T e1, T e2, T e3, T e4, T e5) {
       return ImmutableList.of(e1, e2, e3, e4, e5);
     }
   }
+
+  /**
+   * Prefer {@link Stream#of(Object[])} over constructing a list and then directly calling {@link
```
Stephan202:
```suggestion
   * Prefer {@link Stream#of(Object)} over constructing a list and then directly calling {@link
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ImmutableListRules.java
```diff
@@ -263,4 +263,84 @@ ImmutableList<T> after(T e1, T e2, T e3, T e4, T e5) {
...
+   */
+  static final class StreamImmutableListOf1<T> {
+    @BeforeTemplate
+    Stream<T> before(T e1) {
+      return ImmutableList.of(e1).stream();
+    }
+
+    @AfterTemplate
+    Stream<T> after(T e1) {
+      return Stream.of(e1);
```
Stephan202: Just like with the naming, we try to group based on their `@AfterTemplate`. As such I'll move these new rules to the `StreamRules` class.

Stephan202 (approved):
These `ImmutableList#of` overloads aren't very friendly to the way in which Refaster works, thus requiring this kind of duplication in various places. At some point we should write a checker instead, but for now this SGTM :)

```
Introduce `StreamOf{1,2,3,4,5}` Refaster rules (#814)
```

rickie (approved):

PR #744, opened 2023-08-06
Introduce `IsEmpty` matcher for use by Refaster templates

Stephan202:
Suggested commit message:
```
Introduce `IsEmpty` matcher for use by Refaster templates (#744)
```

As discussed offline with @rickie: this matcher could be used for #733 (assuming that it's extended to support `Mono<Void>`; that's not part of this PR.)

PR #815, opened 2023-10-06
Introduce `FluxEmpty` Refaster rule

Venorcis:
This is mainly intended to prevent `Flux.just()`, preferring the more expressive `Flux.empty()`, but while there why not prevent the other `vararg` constructors from being used without arguments 🤷🏻‍♂️

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -426,6 +426,25 @@ final class ReactorRules {
     }
   }
 
+  /** Prefer {@link Flux#empty()} over more contrived ways of creating an empty flux. */
+  static final class FluxEmpty<T> {
+    @BeforeTemplate
+    Flux<T> before() {
+      return Refaster.anyOf(
+          Flux.just(),
+          Flux.concat(),
+          Flux.concatDelayError(),
+          Flux.firstWithSignal(),
+          Flux.merge(),
+          Flux.mergeSequential());
+    }
+
+    @AfterTemplate
+    Flux<T> after() {
+      return Flux.empty();
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -391,6 +391,25 @@ Flux<T> after(Flux<T> flux, T object) {
     }
   }
 
+  /** Prefer {@link Flux#empty()} over more contrived ways of creating an empty flux. */
+  static final class FluxEmpty<T> {
+    @BeforeTemplate
+    Flux<T> before() {
+      return Refaster.anyOf(
+          Flux.just(),
```
rickie: Would like to propose to sort the entries :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -391,6 +391,25 @@ Flux<T> after(Flux<T> flux, T object) {
     }
   }
 
+  /** Prefer {@link Flux#empty()} over more contrived ways of creating an empty flux. */
```
rickie: Usually we just say `more contrived alternatives`, so would say that would suffice here as well 😄.

rickie (approved):
Nice improvements @Venorcis 🚀 !

Stephan202 (approved):
I rebased and added a commit. I.c.w. #744 we could rewrite even more cases. (But we would be entering "just write a `BugChecker`" territory.

Nice addition!

```
Introduce `Flux{Empty,Just}` Refaster rules (#815)
```

rickie:
Will take a look at the new changes Monday and merge it 👍🏻.

rickie (approved):
Didn't review all the extra operators in too much detail, but overall LGTM!

PR #816, opened 2023-10-06
Introduce `JUnitSingleArguments` bug checker

Venorcis:
Adds a bugchecker that flags uses of JUnit `Argument`s when only a single (or in a weird edge case no) argument is used.

An example can be seen in the diff of the `MethodMatcherFactoryTest` 😄

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitSingleArguments.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitSingleArguments.java
@@ -1,0 +1,50 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import static com.google.errorprone.BugPattern.LinkType.CUSTOM;
+import static com.google.errorprone.BugPattern.SeverityLevel.SUGGESTION;
+import static com.google.errorprone.BugPattern.StandardTags.SIMPLIFICATION;
+import static com.google.errorprone.matchers.method.MethodMatchers.staticMethod;
+import static tech.picnic.errorprone.bugpatterns.util.Documentation.BUG_PATTERNS_BASE_URL;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.MethodInvocationTreeMatcher;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import tech.picnic.errorprone.bugpatterns.util.SourceCode;
+
+/**
+ * A {@link BugChecker} that flags uses of {@link
+ * org.junit.jupiter.params.provider.Arguments#arguments(Object...)} with a single (or no) argument;
+ * in such cases the use of {@link org.junit.jupiter.params.provider.Arguments} is not required as a
+ * {@link java.util.stream.Stream} of objects can be directly provided to a {@link
+ * org.junit.jupiter.params.provider.MethodSource}.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "JUnit arguments wrapping a single object are redundant",
+    link = BUG_PATTERNS_BASE_URL + "JUnitSingleArguments",
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = SIMPLIFICATION)
+public final class JUnitSingleArguments extends BugChecker implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<ExpressionTree> ARGUMENTS_ARGUMENTS =
+      staticMethod().onClass("org.junit.jupiter.params.provider.Arguments").named("arguments");
+
+  /** Instantiates a new {@link JUnitSingleArguments} instance. */
+  public JUnitSingleArguments() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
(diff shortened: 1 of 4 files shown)


error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitSingleArgumentsTest.java
```diff
@@ -0,0 +1,30 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+final class JUnitSingleArgumentsTest {
+  @Test
+  void identification() {
+    CompilationTestHelper.newInstance(JUnitSingleArguments.class, getClass())
+        .addSourceLines(
```
Venorcis, line 10: I don't get what's not formatting-compliant here 🤔

Stephan202: It was subtle; ran `./apply-error-prone-suggestions.sh` and pushed :)

Venorcis: Ah thanks, didn't know that existed 😄 Also the command line error just gives 'did you mean' with the first few lines, which is confusing 🤪

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitSingleArguments.java
```diff
@@ -0,0 +1,50 @@
...
+  private static final Matcher<ExpressionTree> ARGUMENTS_ARGUMENTS =
+      staticMethod().onClass("org.junit.jupiter.params.provider.Arguments").named("arguments");
+
+  /** Instantiates a new {@link JUnitSingleArguments} instance. */
+  public JUnitSingleArguments() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (ARGUMENTS_ARGUMENTS.matches(tree, state) && tree.getArguments().size() <= 1) {
+      return describeMatch(tree, SourceCode.unwrapMethodInvocation(tree, state));
```
Stephan202, line 45: This would cause a compilation error w.r.t. the return type, right? :thinking:

Stephan202 (commented):
Nice rule! I suspect we an make it robust in the common case, but will need to find some time to look closer into the least-impactful approach.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitSingleArguments.java
Venorcis: Yeah, I wasn't sure how we deal with the fix suggestions. AFAIK not all suggestions necessarily lead to directly compiling code? We can ofc. remove it here and just flag instead 🤷🏻‍♂️

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/JUnitSingleArgumentsTest.java
Stephan202: Yeah, that output is... "not great". Haven't really found a way yet to make that play nicer with Error Prone.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitSingleArguments.java
Stephan202: We _try_ either produce compilable code _or_ only flag the issue, with a strong preference for the former (as the latter really introduces quite a bit more friction). That said, especially some of the Refaster rules can in special circumstances break the code.

For this case we could:
1. Simplify the factory method if it has a single return statement, using similar logic we use in [`JUnitValueSource`](https://github.com/PicnicSupermarket/error-prone-support/blob/da5eea832975cb9222cba51935f7a22b4ff35e46/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitValueSource.java).
2. Only flag otherwise.

(But that's considerably more complicated than the current proposal, so if you're like "uuuuuh", I get that. Don't mind having a stab; would just take a bit of time.)

Venorcis: Yeah understood. In the most common use case we 'only' need to replace `<Arguments>` with `<T>` in the method definition. Perhaps it's relatively easy to add the suggestion for this use case and still resort to flagging any other occurrences (they would be weird but you never know 😛 ).

Stephan202: Indeed, that's the lines along which I'm thinking :+1:

(Both inside Picnic and for the upcoming integration test framework against open-source repos we run Error Prone in a loop until it no longer makes any changes, which can only work if each intermediate state compiles successfully.)

Stephan202: I've looked a bit closer at this topic, and the logic in `JUnitValueSource` shows that there are quite a lot of details to consider:
1. There are different ways in which the `Arguments` can be wrapped in a top-level collection/stream/array.
2. The factory method can have multiple return statements.
3. The return type could change from `MonadType<Arguments>` or `Arguments[]` to `MonadType<Object>` or `Object[]`, but ideally we're more specific.
4. Only if all `Arguments` instances have exactly one value can we do the unwrapping (in case of variability we should likely emit a separate warning).
5. Ideally we do all this in a way that reuses relevant pieces of logic from `JUnitValueSource`.

I'm likely missing a few other details. Happy to dive deeper into this, but it'll be a while before I find the time to really sit down for this.

PR #817, opened 2023-10-06
Introduce `JUnitParameterizedMethodDeclaration` bug checker

Venorcis:
Flag uses of `@ParameterizedTest` on methods that don't have any parameters.


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitParameterizedMethodDeclaration.java
```diff
@@ -0,0 +1,48 @@
...
+  private static final Matcher<MethodTree> IS_PARAMETERIZED_TEST =
+      annotations(AT_LEAST_ONE, isType("org.junit.jupiter.params.ParameterizedTest"));
+
+  /** Instantiates a new {@link JUnitParameterizedMethodDeclaration} instance. */
+  public JUnitParameterizedMethodDeclaration() {}
+
+  @Override
+  public Description matchMethod(MethodTree tree, VisitorState state) {
+    if (IS_PARAMETERIZED_TEST.matches(tree, state) && tree.getParameters().isEmpty()) {
+      return describeMatch(tree);
```
Venorcis, line 43: I cannot find an option to rename the annotation 🤔

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitParameterizedMethodDeclaration.java
```diff
@@ -0,0 +1,48 @@
...
+ * org.junit.jupiter.params.ParameterizedTest} without actually having any arguments.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "JUnit parameterized test used without arguments",
+    link = BUG_PATTERNS_BASE_URL + "JUnitParameterizedMethodDeclaration",
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = SIMPLIFICATION)
+public final class JUnitParameterizedMethodDeclaration extends BugChecker
```
Stephan202: Proposed name: `JUnitNullaryParameterizedTestDeclaration`. (Because we're not validating other `@ParameterizedTest` constraints (such as that they're at least one argument source) and because in practice the flagged methods aren't parameterized.)

Stephan202 (approved):
Rebased and added a commit; the check now also suggests a fix. Nice idea for a check @Venorcis!

Suggested commit message:
```
Introduce `JUnitNullaryParameterizedTestDeclaration` check (#817)

While there, slightly simplify the `AutowiredConstructor` check.
```

Venorcis:
> Rebased and added a commit; the check now also suggests a fix. Nice idea for a check @Venorcis!

Thanks @Stephan202, all LGTM! 🎉

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitNullaryParameterizedTestDeclaration.java
```diff
@@ -0,0 +1,91 @@
...
+ * A {@link BugChecker} that flags nullary {@link
+ * org.junit.jupiter.params.ParameterizedTest @ParameterizedTest} test methods.
+ *
+ * <p>Such tests are unnecessarily executed more than necessary. This checker suggests annotating
+ * the method with {@link org.junit.jupiter.api.Test @Test}, and to drop all declared {@link
+ * org.junit.jupiter.params.provider.ArgumentsSource argument sources}.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Nullary JUnit test methods need not be parameterized",
```
rickie, line 38: This is probably a personal thingy, but I do not really like "need not be" phrase. Maybe we can change to "should not".

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitNullaryParameterizedTestDeclaration.java
```diff
@@ -0,0 +1,91 @@
...
+    link = BUG_PATTERNS_BASE_URL + "JUnitNullaryParameterizedTestDeclaration",
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = SIMPLIFICATION)
+public final class JUnitNullaryParameterizedTestDeclaration extends BugChecker
+    implements MethodTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final MultiMatcher<MethodTree, AnnotationTree> IS_PARAMETERIZED_TEST =
+      annotations(AT_LEAST_ONE, isType("org.junit.jupiter.params.ParameterizedTest"));
+  private static final Matcher<AnnotationTree> IS_ARGUMENT_SOURCE =
```
rickie, line 48: Makes me think of the `MoreJUnitMatchers#HAS_METHOD_SOURCE` ([link](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/MoreJUnitMatchers.java#L51)) but this usage is slightly different is fine as-is and we don't need to consider any changes right now. Anyway, wanted to highlight it.

rickie (approved):
The changes LGTM! Nice idea for a check 🚀 !

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitNullaryParameterizedTestDeclaration.java
Stephan202: Indeed, fine to move such things only when a second use case comes up :+1:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitNullaryParameterizedTestDeclaration.java
Stephan202: Yes, it's you ;)

But will add a commit.

Stephan202 (commented):
Rebased and added a commit. @rickie feel free to merge once built.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/JUnitNullaryParameterizedTestDeclaration.java
rickie: 😄 haha nice. Will merge once 🟢 !

PR #822, opened 2023-10-07
Introduce `{Integer,Long}SignumIs{Positive,Negative}` Refaster rules

Stephan202:
Suggested commit message:
```
Introduce `{Integer,Long}SignumIs{Positive,Negative}` Refaster rules (#822)
```

These suggestions follow from the discussion in #812.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PrimitiveRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PrimitiveRules.java
@@ -8,7 +8,9 @@ import com.google.common.primitives.Floats;
 import com.google.common.primitives.Ints;
 import com.google.common.primitives.Longs;
 import com.google.common.primitives.Shorts;
+import com.google.errorprone.refaster.Refaster;
 import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.AlsoNegation;
 import com.google.errorprone.refaster.annotation.BeforeTemplate;
 import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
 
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PrimitiveRules.java
```diff
@@ -384,4 +386,98 @@ boolean after(double d) {
...
+
+    @BeforeTemplate
+    boolean before(double v) {
+      return Refaster.anyOf(Math.signum(v) > 0, Math.signum(v) >= 1);
+    }
+
+    @AfterTemplate
+    @AlsoNegation
+    boolean after(double v) {
+      return Math.signum(v) == 1;
```
Stephan202, line 461: Here and in the `@AfterTemplaye` below flags [_S1244: Equality tests should not be made with floating point values_](https://sonarcloud.io/organizations/picnic-technologies/rules?open=java%3AS1244&rule_key=java%3AS1244), which is fair. Options:
1. Omit the `Math.signum` rules.
2. Suppress the warning, since the this will do the right thing, as per the documentation.

Option (2) does force others to suppress the warning too, though, so perhaps (1) is better?

Stephan202 (commented):
Thoughts appreciated.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PrimitiveRules.java
Venorcis: Oof, returning a double/float that is fixed to `1.0 / 0.0 / -1.0` is an interesting design indeed 😿 IMO generally most uses of double/float should be considered very carefully, but more to the point, I guess I'd lean towards the consistency and keep it included.

Venorcis (approved):

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PrimitiveRules.java
rickie: Hmm, IIUC the equality check *will* work right? Because the floating point of 1 equality check is just fine. However, based on the Sonar rule I'd say we should not rewrite this one. It would maybe be better to flag this one as something that is not something one should do in code. 

Additionally, for people outside of Picnic that have setup Sonar with most rules enabled will have to suppress / work around this one after our rewite, if I'm correct.

Stephan202: K, I rebased and dropped the `Math` rules, at least for now. That way we can merge the PR and defer the rest (to potentially never do it).

Stephan202 (commented):
Also rebased.

rickie (approved):
Nice 😃

PR #832, opened 2023-10-11
Build and test against JDK 21 instead of JDK 20

Stephan202:
Suggested commit message:
```
Build and test against JDK 21 instead of JDK 20 (#832)

See https://jdk.java.net/21/release-notes
```


werli (approved):
:tada:

rickie (approved):

PR #850, opened 2023-10-21
Sort arguments of RefasterRuleCollectionTestCase#elidedTypesAndStaticImports

mohamedsamehsalah:
This PR is split into two commits:

- Code implementation.
- Manually applied suggestions

Suggested commit message:
```
Sort arguments of RefasterRuleCollectionTestCase#elidedTypesAndStaticImports (#850)
```


refaster-test-support/src/main/java/tech/picnic/errorprone/refaster/test/RefasterRuleCollection.java
```diff
@@ -173,6 +186,68 @@ private void reportIncorrectClassName(CompilationUnitTree tree, VisitorState sta
...
+          returnTree
+              .map(ReturnTree::getExpression)
+              .map(MethodInvocationTree.class::cast)
+              .ifPresent(
+                  methodInvocationTree -> {
+                    List<? extends ExpressionTree> actualArgumentsOrdering =
+                        methodInvocationTree.getArguments();
+
+                    ImmutableList<? extends ExpressionTree> desiredArgumentOrdering =
+                        actualArgumentsOrdering.stream()
```
mohamedsamehsalah, line 211: Should we group argument types togther before sorting ?
i.e: Group & sort method invocation lexicographically > group & sort identifiers lexicographically, etc...

refaster-test-support/src/main/java/tech/picnic/errorprone/refaster/test/RefasterRuleCollection.java
```diff
@@ -173,6 +186,68 @@ private void reportIncorrectClassName(CompilationUnitTree tree, VisitorState sta
...
+                              .collect(
+                                  joining(
+                                      ", ",
+                                      String.format(
+                                          "%s(",
+                                          state.getSourceForNode(
+                                              methodInvocationTree.getMethodSelect())),
+                                      ")"));
+
+                      fixBuilder.merge(SuggestedFix.replace(methodInvocationTree, suggestion));
```
mohamedsamehsalah, line 231: Curious why running `./apply-error-prone-suggestions.sh` did not fix the *(Input|Otput) tests here 🤔

rickie: So what we are doing in the `RefasterRuleCollection`, is reporting things "in" the output file, that is what we are also verifying in the `refaster-test-support/src/test/resources/tech/picnic/errorprone/refaster/test/NonSortedElidedTypesAndStaticImportsTestRulesTestOutput.java` file. 

So e.g. in the `MatchInWrongMethodRulesTest` input file we try to see what is wrong and we _report_ that in the `MatchInWrongMethodRulesTest` output file, while not doing anything with the content itself, we only add a warning as a comment. 

So in the `NonSortedElidedTypesAndStaticImportsTestRulesTestOutput` file we should not change the output to be correct, we only need a nice comment explaining what is wrong. Instead of 
```
/* ERROR: Unexpected token. */
```
(which also signals something is wrong and the code we add has likely a bug)
It should add something like:
```
/*
   *  ERROR: The entries of the `elidedTypesAndStaticImports` method are not sorted lexicographically. The method should look like this:
   *   public ImmutableSet<Object> elidedTypesAndStaticImports() {
   *     return ImmutableSet.of(
              Assertions.class,
.....  (a method with correct sorting)
*/
```

This thing is not a "regular" BugPattern. It tries to guide a developer when writing TestInput & TestOutput files by giving this warning :).

rickie (commented):
I gave some context that should make some things a bit clearer, if there are any questions let us know 😄.

Nice that you opened another PR @mohamedsamehsalah 🚀 !

refaster-test-support/src/main/java/tech/picnic/errorprone/refaster/test/RefasterRuleCollection.java
mohamedsamehsalah: Thanks for explanation 🙇 🙇

refaster-test-support/src/main/java/tech/picnic/errorprone/refaster/test/RefasterRuleCollection.java
mohamedsamehsalah: > Do you have an example of the case you mean?

My thought was (As an example)
1. Group & sort Objects (Strings for eg).
2. Group & sort Class statemens (`*.class`).
3. Group & sort method arguments.
5. Group & sort primitives.

Essentially turning
```
ImmutableSet.of(List.class, "A", assertDoesNotThrow(() -> null), Iterables.class, "a", 1)
```
to 
```
ImmutableSet.of("A", "a", List.class, Iterables.class, assertDoesNotThrow(() -> null),1)
```

PR #862, opened 2023-10-26
Disallow static import of `BugCheckerRefasctoringTestHelper.TestMode#TEXT_MATCH`

rickie:
I remembered we wanted to fix this one some time ago.

However, it does a non 100% optimal fix for this specific case as it rewrites it to:
```
BugCheckerRefactoringTestHelper.TestMode.TEXT_MATCH
```
Which can be easily fixed by removing the qualifier, but I'd say flagging this is an improvement for now.
Didn't add a test as we already extensively test the members in the `identification` test.

Suggested commit message:
```
Disallow static import of `BugCheckerRefasctoringTestHelper.TestMode#TEXT_MATCH` (#862)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
@@ -78,6 +78,7 @@ public final class NonStaticImport extends BugChecker implements CompilationUnit
   static final ImmutableSetMultimap<String, String> NON_STATIC_IMPORT_CANDIDATE_MEMBERS =
       ImmutableSetMultimap.<String, String>builder()
           .put("com.google.common.base.Predicates", "contains")
+          .put("com.google.errorprone.BugCheckerRefactoringTestHelper.TestMode", "TEXT_MATCH")
           .putAll(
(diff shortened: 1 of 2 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
```diff
@@ -78,6 +78,7 @@ public final class NonStaticImport extends BugChecker implements CompilationUnit
   static final ImmutableSetMultimap<String, String> NON_STATIC_IMPORT_CANDIDATE_MEMBERS =
       ImmutableSetMultimap.<String, String>builder()
           .put("com.google.common.base.Predicates", "contains")
+          .put("com.google.errorprone.BugCheckerRefactoringTestHelper.TestMode", "TEXT_MATCH")
```
Stephan202: It's not like we do want to statically import `AST_MATCH`, so I suppose we can extend `NON_STATIC_IMPORT_CANDIDATE_TYPES` instead.

Stephan202 (approved):
```
Disallow static import of `BugCheckerRefasctoringTestHelper.TestMode` members (#862)
```

cernat-catalin (approved):

PR #865, opened 2023-10-28
Introduce Checkstyle integration test

Stephan202:
Suggested commit message:
```
Introduce Checkstyle integration test (#865)

The new `integration-tests/checkstyle-10.12.4.sh` script can be executed
locally or by adding a comment containing the string `/integration-test`
to a pull request.

The integration test comprises a shell script that:
1. Checks out a Checkstyle release tag.
2. Applies a small patch to enable Error Prone Support.
3. Runs the build in Error Prone patch mode, until no further changes
   are applied.
4. Validates that:
   - The build (including tests) still passes.
   - The set of changes matches a pre-recorded diff.
   - The set of Error Prone Support-emitted warnings matches a
     pre-recorded list.

The script also has a `--sync` mode using which the expected diff and 
warnings can be updated. Combined, the script makes it easy to assess
and track the impact of changes to Error Prone Support in relation to
the Checkstyle code base.

Over time this setup will be generalized to include integration tests
against other code bases.
```


integration-tests/checkstyle-10.12.4.sh
```diff
@@ -0,0 +1,162 @@
...
+  #
+  # Under certain circumstances this does not cause the relevant tag to be
+  # created, so if necessary we manually create it.
+  git -C "${project_root}" checkout --force "${revision}" 2>/dev/null \
+    || (
+         git -C "${project_root}" fetch --depth 1 "${repository}" "${revision}" \
+           && git -C "${project_root}" checkout --force FETCH_HEAD \
+           && git -C "${project_root}" tag "${revision}" || true
+       )
+fi
```
Stephan202, line 77: On GitHub Actions there'll always be a new checkout, but for local development it's very convenient if repeated script runs don't trigger new fetch/clone operations.

integration-tests/checkstyle-10.12.4.sh
```diff
@@ -0,0 +1,162 @@
...
+    # changes. For performance reasons we perform incremental compilation,
+    # enabled using a misleading flag. (See
+    # https://issues.apache.org/jira/browse/MCOMPILER-209 for details.)
+    apply_patch '-Dmaven.compiler.useIncrementalCompilation=false'
+  elif [ "${extra_build_args}" != 'clean' ]; then
+    # No changes were applied. We'll attempt one more round in which all files
+    # are recompiled, because there are cases in which violations are missed
+    # during incremental compilation.
+    apply_patch 'clean'
+  fi
```
Stephan202, line 115: Yes, this is a bit awkward, but the added complexity seems worth the reduced runtime, at least for a build as large as this one. (Current runtime on my machine is ~11 minutes. Don't recall how long it is without this tweak; might do a new timing later.)

integration-tests/checkstyle-10.12.4.sh
```diff
@@ -0,0 +1,162 @@
...
+    clean package \
+    -Derror-prone.flags="${error_prone_validation_flags}" \
+    -Dtest='
+      !MetadataGeneratorUtilTest#metadataFilesGenerationAllFiles,
+      !XdocsJavaDocsTest#allCheckSectionJavaDocs' \
+  | tee "${validation_log_file}"
+
+# Perist or validate the warnings reported by Error Prone Support checks.
+baseline_warnings="${integration_test_root}/${test_name}-expected-warnings.txt"
+generated_warnings="$((grep -oP "(?<=^\\Q[WARNING] ${PWD}/\\E).*" "${validation_log_file}" | grep -P '\] \[' || true) | sort)"
```
Stephan202, line 152: Here and elsewhere, the use of `grep -P` likely doesn't make this script Macos-compatible; that's something to review separately.

Stephan202 (commented):
Added some context. Testing the GitHub action integration is TBD (I expect issues with the `git commit` command). Might also want to run ShellCheck.

Stephan202:
Alright, probably the first of quite a lot of trial-and-error: /integration-test.

Stephan202:
Ah right, this might not work, given that the PR isn't merged yet. Will try something else later.

Stephan202:
I temporarily changed the default branch to be this PR's branch. So, take two: /integration-test

Stephan202:
That [worked](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/6682426284?pr=865+%28comment%29). I restored the default branch.

Stephan202:
Validated in [this build](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/6683968279?pr=865) that artifacts are attached on build failure.

Stephan202:
Alright, I think this PR is ready for review. Once merged we can start working on adding additional projects (something @oxkitsune has already worked on).

integration-tests/checkstyle-10.12.4.sh
```diff
@@ -0,0 +1,183 @@
...
+# Make sure that Git is sufficient configured to enable committing to the
+# project's Git repository.
+git config user.email || git config user.email "integration-test@example.com"
+git config user.name || git config user.name "Integration Test"
+
+# Prepare the code for analysis by (a) applying the minimal set of changes
+# required to run Error Prone with Error Prone Support and (b) formatting the
+# code using the same method by which it will be formatted after each
+# compilation round. The initial formatting operation ensures that subsequent
+# moifications can be rendered in a clean manner.
```
rickie:
```suggestion
# modifications can be rendered in a clean manner.
```

integration-tests/checkstyle-10.12.4.sh
```diff
@@ -0,0 +1,183 @@
...
+    || (
+         git -C "${project_root}" fetch --depth 1 "${repository}" "${revision}" \
+           && git -C "${project_root}" checkout --force FETCH_HEAD \
+           && (git -C "${project_root}" tag "${revision}" || true)
+       )
+fi
+
+pushd "${project_root}"
+
+# Make sure that Git is sufficient configured to enable committing to the
```
rickie:
```suggestion
# Make sure that Git is sufficiently configured to enable committing to the
```
Right?

integration-tests/checkstyle-10.12.4.sh
```diff
@@ -0,0 +1,183 @@
...
+if [ -n "${do_sync}" ]; then
+  echo 'Saving changes...'
+  cp "${actual_changes}" "${expected_changes}"
+  cp "${actual_warnings}" "${expected_warnings}"
+else
+  echo 'Inspecting changes...'
+  # XXX: This "diff of diffs" also contains vacuous sections, introduced due to
+  # line offset differences. Try to omit those from the final output.
+  if ! diff -u "${expected_changes}" "${actual_changes}"; then
+    echo 'There are unexpected changes.'
```
rickie, line 171: Wouldn't it be great to output that to a file and upload that? I'd say that is of great value.

integration-tests/checkstyle-10.12.4.sh
```diff
@@ -0,0 +1,183 @@
...
+git config user.name || git config user.name "Integration Test"
+
+# Prepare the code for analysis by (a) applying the minimal set of changes
+# required to run Error Prone with Error Prone Support and (b) formatting the
+# code using the same method by which it will be formatted after each
+# compilation round. The initial formatting operation ensures that subsequent
+# moifications can be rendered in a clean manner.
+git clean -fdx
+git apply < "${integration_test_root}/${test_name}-init.patch"
+git commit -m 'dependency: Introduce Error Prone Support' .
```
rickie, line 102: I see we are adhering to the guidelines of Checkstyle here, which makes me wonder, how do we want to generalize this script for the next runs? The setup is nice and seems to be written with a generalization in mind.

The reason I'm asking is that I started wondering if we always want to use the `minor` and `dependency` in the commit messages.

rickie (commented):
Added a small commit. 

How will the structure in `integration-tests` look like when we add more? And more versions? Do we want a directory per library? I see you changed some names and based on that I kind of assume you have already something in mind. Happy to think along, but for now going to call it a day :).

rickie:
Forgot to say something important; man this setup is amazing, cool stuff ! ✨ 💖 This will be so useful!

integration-tests/checkstyle-10.12.4.sh
Stephan202: Likely we'll want to make some compromises. Either by disabling additional checks or by making this parametric. While I already spent significant time on making the script reusable, there is certainly more to do. That's why the file is simply named `integration-tests/checkstyle-10.12.4.sh` ;)

integration-tests/checkstyle-10.12.4.sh
Stephan202: That already happens; see the `actions/upload-artifact` goal. (And [this comment](https://github.com/PicnicSupermarket/error-prone-support/pull/865#issuecomment-1784136320).)

Stephan202 (commented):
> How will the structure in `integration-tests` look like when we add more? And more versions? Do we want a directory per library? I see you changed some names and based on that I kind of assume you have already something in mind. Happy to think along, but for now going to call it a day :).

Good questions. I'm undecided about the value of tracking multiple versions per library, so _perhaps_ dropping the versions from the file names make sense. Something to decide once we wish to update the tag against which we tag. W.r.t. whether to introduce subdirectories: also not sure; the current shared prefixes may suffice. The biggest thing to figure out is how to maximally reuse test script logic. I guess we'll figure it out as we go.

integration-tests/checkstyle-10.12.4.sh
rickie: I meant the "diff of the diffs". IIUC only the actual changes, actual logs and actual warnings are logged. While in cases of new changes you really want to see what the diff of the diffs is and inspect that.

integration-tests/checkstyle-10.12.4.sh
rickie: Ah that makes sense, sounds good for now :).

integration-tests/checkstyle-10.12.4-init.patch
```diff
@@ -0,0 +1,185 @@
...
+     public static DetailNode parseJavadocAsDetailNode(DetailAST blockComment) {
+         final JavadocDetailNodeParser parser = new JavadocDetailNodeParser();
+         final ParseStatus status = parser.parseJavadocAsDetailNode(blockComment);
+--- a/src/main/java/com/puppycrawl/tools/checkstyle/Main.java
++++ b/src/main/java/com/puppycrawl/tools/checkstyle/Main.java
+@@ -626,6 +626,8 @@ public final class Main {
+             + "reported to standard out in plain format. Checkstyle requires a configuration "
+             + "XML file that configures the checks to apply.",
+             mixinStandardHelpOptions = true)
++    // XXX: Don't reorder arguments to `picocli.CommandLine.Option#names`.
```
rickie, line 92: Or I don't get this sentence or it should be "from" instead of "to" 😋.

integration-tests/checkstyle-10.12.4-init.patch
```diff
@@ -0,0 +1,189 @@
...
+--- a/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
++++ b/src/main/java/com/puppycrawl/tools/checkstyle/site/SiteUtil.java
+@@ -419,6 +419,10 @@ public final class SiteUtil {
+      * @return a set of properties for the given class.
+      */
+     public static Set<String> getPropertiesForDocumentation(Class<?> clss, Object instance) {
++        // XXX: File PR to replace `.collect(toSet())` with `.collect(toCollection(HashSet::new))`.
++        // XXX: Update `CollectorMutability` to recognize cases such as this one, where the created
++        // collection is clearly modified.
++        @SuppressWarnings("CollectorMutability")
```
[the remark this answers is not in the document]
rickie: @oxkitsune are you interested?

integration-tests/checkstyle-10.12.4.sh
Stephan202: That diff-of-diffs is in the build log, and also easy to reproduce. Not opposed to adding this, but perhaps something for another PR. (I can't justify spending time on this right now.)

rickie: Fine with deferring but I think its nice to add :).

rickie (approved):
It'll take a while before I'll understand all specifics of the bash script, but overall it looks good to me and I'm looking forward to start using this!

rickie:
I'd say we can use "new feature" for this, or add a new label "integration-tests"/"integration-tests-framework"? I can file a PR such that it will get it's own section in the release notes.

Stephan202:
Good point. This is primarily relevant for contributors, with as a secondary benefit that hopefully the average quality of releases will go up. Perhaps the existing `improvement` label suffices?

rickie:
Fine by me, I just imagine that we will have quite some PRs related to this test framework so it could be convenient to have a label for it. Having it under improvement also works for now.

oxkitsune (approved):
lgtm! 
On the topic of making it more generic for other repos, I think we can extract a large part of the script to a generic `run-integration-tests.sh`, and have the more repo specific stuff in a `<repo>-<tag>.sh` script :smile:

PR #880, opened 2023-11-12
Introduce `UnqualifiedSuggestedFixImport` check

Stephan202:
Suggested commit message:
```
Introduce `UnqualifiedSuggestedFixImport` check (#880)

Usage of `SuggestedFix.Builder#add{,Static}Import` does not always yield
valid code, so this check suggests alternatives instead.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImportSuggestion.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImportSuggestion.java
@@ -44,7 +44,7 @@ public final class ImportSuggestion extends BugChecker implements MethodInvocati
       return Description.NO_MATCH;
     }
 
-    switch (ASTHelpers.getSymbol(tree).name.toString()) {
+    switch (ASTHelpers.getSymbol(tree).getSimpleName().toString()) {
       case "addImport":
         return createDescription(
             tree, "SuggestedFix.Builder#addImport", "SuggestedFixes#qualifyType");
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImportSuggestion.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImportSuggestion.java
@@ -1,0 +1,66 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import static com.google.errorprone.BugPattern.LinkType.CUSTOM;
+import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
+import static com.google.errorprone.BugPattern.StandardTags.FRAGILE_CODE;
+import static com.google.errorprone.matchers.Matchers.instanceMethod;
+import static tech.picnic.errorprone.bugpatterns.util.Documentation.BUG_PATTERNS_BASE_URL;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.MethodInvocationTreeMatcher;
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+
+/**
+ * A {@link BugChecker} that flags suggested fixes that involve unconditional imports.
+ *
+ * <p>Such unconditional imports may clash with other imports of the same identifier.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Avoid direct invocation of `SuggestedFix#add{,Static}Import`",
+    link = BUG_PATTERNS_BASE_URL + "ImportSuggestion",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = FRAGILE_CODE)
+public final class ImportSuggestion extends BugChecker implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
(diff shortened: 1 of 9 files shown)


Stephan202:
Added one more commit with some cleanup that is enabled by this check; that code will be further improved by #882 once it lands.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImportSuggestion.java
```diff
@@ -0,0 +1,66 @@
...
+  /** Instantiates a new {@link ImportSuggestion} instance. */
+  public ImportSuggestion() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!FIX_BUILDER_METHOD.matches(tree, state)) {
+      return Description.NO_MATCH;
+    }
+
+    switch (ASTHelpers.getSymbol(tree).name.toString()) {
```
rickie, line 47:
```suggestion
    switch (ASTHelpers.getSymbol(tree).getSimpleName().toString()) {
```
What do we prefer in these cases?

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImportSuggestion.java
```diff
@@ -0,0 +1,66 @@
...
+ * <p>Such unconditional imports may clash with other imports of the same identifier.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Avoid direct invocation of `SuggestedFix#add{,Static}Import`",
+    link = BUG_PATTERNS_BASE_URL + "ImportSuggestion",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = FRAGILE_CODE)
+public final class ImportSuggestion extends BugChecker implements MethodInvocationTreeMatcher {
```
rickie, line 33: I'm thinking that `ImportSuggestion` sounds a bit "generic" for its goal. Maybe one doesn't directly link `Suggestion` to the `SuggestedFix` part. Perhaps the word `SuggestedFix` should be in  there as it is specific to that.

rickie (approved):
Only two considerations :). Approving nonetheless.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpringMvcAnnotation.java
```diff
@@ -114,9 +115,8 @@ private static Fix replaceAnnotation(
             .map(arg -> SourceCode.treeToString(arg, state))
             .collect(joining(", "));
 
-    return SuggestedFix.builder()
-        .addImport(ANN_PACKAGE_PREFIX + newAnnotation)
-        .replace(tree, String.format("@%s(%s)", newAnnotation, newArguments))
-        .build();
+    SuggestedFix.Builder fix = SuggestedFix.builder();
+    String annotation = SuggestedFixes.qualifyType(state, fix, ANN_PACKAGE_PREFIX + newAnnotation);
```
mohamedsamehsalah, line 119: To be consistent with namings of "replacement" variables inside the method.
```suggestion
    String newAnnotation = SuggestedFixes.qualifyType(state, fix, ANN_PACKAGE_PREFIX + annotation);
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/MethodReferenceUsage.java
```diff
@@ -196,6 +196,8 @@ private static ImmutableList<Name> getVariables(LambdaExpressionTree tree) {
     return tree.getParameters().stream().map(VariableTree::getName).collect(toImmutableList());
   }
 
+  // XXX: Resolve this suppression.
```
mohamedsamehsalah, line 199: I also tried 😓 

Do we usually create tickets for these ?

rickie: The `MethodReferenceUsage` check is not enabled currently. It requires some more big hurdles before we can enable this. So we probably want to defer this task till we start working on this check again :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImportSuggestion.java
Stephan202: We should use the method name indeed :+1:

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/SpringMvcAnnotation.java
Stephan202: This would require changing the method parameter name, which currently service as a bit of "self-documenting code" for the caller. At the same time, `annotation` is referenced only on the next line, which should make its semantics clear. So I'd prefer to keep it as-is.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImportSuggestion.java
Stephan202: Hmm, hard to come up with another name. Perhaps `UnqualifiedSuggestedFixImport`?

Stephan202 (commented):
Tnx for the reviews! Added a small commit.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ImportSuggestion.java
rickie: Sounds good IMO, I'll commit the change.

Stephan202:
Rename LGTM. I updated the PR title and suggested commit message. Now resolving conflict.

rickie:
>  Now resolving conflict. 

Oh I missed something? 🤔

Stephan202:
Nope, unrelated conflict. Will merge :)

PR #881, opened 2023-11-12
Introduce `CanonicalClassNameUsage` check

Stephan202:
Suggested commit message:
```
Introduce `CanonicalClassNameUsage` check (#881)

Error Prone checks deal with source code and type matchers, both of
which generally involve canonical type names, rather than the strings
produced by `Class#getName()`. This distinction is particularly relevant
when dealing with nested types.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalClassNameUsage.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalClassNameUsage.java
@@ -1,0 +1,80 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import static com.google.errorprone.BugPattern.LinkType.CUSTOM;
+import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
+import static com.google.errorprone.BugPattern.StandardTags.FRAGILE_CODE;
+import static com.google.errorprone.matchers.Matchers.instanceMethod;
+import static tech.picnic.errorprone.bugpatterns.util.Documentation.BUG_PATTERNS_BASE_URL;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.annotations.Var;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.MethodInvocationTreeMatcher;
+import com.google.errorprone.fixes.SuggestedFixes;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.BinaryTree;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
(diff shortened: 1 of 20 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalClassNameUsage.java
```diff
@@ -0,0 +1,80 @@
...
+      instanceMethod().onExactClass(Class.class.getCanonicalName()).named("getName");
+  private static final Pattern CANONICAL_NAME_USING_TYPES =
+      Pattern.compile("(com\\.google\\.errorprone|tech\\.picnic\\.errorprone)\\..*");
+
+  /** Instantiates a new {@link CanonicalClassNameUsage} instance. */
+  public CanonicalClassNameUsage() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!GET_NAME_INVOCATION.matches(tree, state) || !isPassedToCanonicalNameUsingType(state)) {
```
mohamedsamehsalah, line 54: In the `Class#getCanonicalName` docs, the method can return null for local, anonymous & hidden classes.

I wonder do we need to add a `CanonicalNameIsNotNull` check here ? 

But I guess the idea is that methods that accept canonical names are aware of the API...

mohamedsamehsalah (approved):
LGTM 🚀 

Left one question.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalClassNameUsage.java
```diff
@@ -0,0 +1,80 @@
...
+@BugPattern(
+    summary = "This code should likely use the type's canonical name",
+    link = BUG_PATTERNS_BASE_URL + "CanonicalClassNameUsage",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = FRAGILE_CODE)
+public final class CanonicalClassNameUsage extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<ExpressionTree> GET_NAME_INVOCATION =
```
rickie, line 44: For these situations we use both `_INVOCATION` and `_METHOD`. I feel `INVOCATION` fits better here. Maybe we should apply this consistently 🤔.

rickie (approved):
Code looks nice! Good improvement :rocket:!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CanonicalClassNameUsage.java
rickie: Oeh nice that you think about such edge cases 🚀 !

Interesting it will indeed not work if we rewrite `new Object(){}.getClass().getName()` as it will return `null`. However, it does not really make sense to write such code in one of the methods that we flag, which are specific methods from `tech.picnic.errorprone` or `com.google.errorprone`, which means a . So I think for now it's fine to not flag this 🤔.

Stephan202 (commented):
Tnx for the reviews @mohamedsamehsalah and @rickie! I rebased and added a commit.

PR #882, opened 2023-11-13
Introduce ErrorProneRuntimeClasspath check

Stephan202:
~:exclamation: This PR is on top of #881. :exclamation:~

Suggested commit message:
```
Introduce `ErrorProneRuntimeClasspath` check (#882)

Prefer "type-safe" type references were possible, but use string
literals if the references type may not be available at runtime.
```

The PR is split into three commits that can best be reviewed one-by-one: the check, its automated suggestions, and manual tweaks.


Stephan202:
I've started to resolve the conflicts here. Almost done; will finalize after some meetings.

Stephan202:
Rebased; the PR now targets `master` and is ready for review.

Stephan202:
Once again rebased and resolved a conflict. @rickie and @mohamedsamehsalah I suspect this'll start happening more ;)

mohamedsamehsalah (approved):
Nothing stands out for me :octocat: 

Is there a general rule to follow when determining if a class is available on run-time or not ?

Stephan202 (commented):
I added one more commit, in which I resorted a few things. The only place I didn't reorder is the listing in `JUnitValueSource`, as there the existing order seems more intuitive.

> Is there a general rule to follow when determining if a class is available on run-time or not ?

Thanks for reviewing @mohamedsamehsalah! I'm not sure about a "general rule", but we use [this code](https://github.com/PicnicSupermarket/error-prone-support/blob/cc36aa993c7c7aedb5fc5c18dc829bd7a50ca9b3/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java#L73-L100) to check whether a given class is "currently" on the classpath. That of course doesn't mean that a class will also be on the classpath when the code is executed at a later point in time. For example, we have a bunch of `provided` dependencies that may not _actually_ be "provided" by downstream users. That's why the `ErrorProneRuntimeClasspath` check restricts itself to types that match the `CLASSPATH_TYPES` regex: those are known to be dependencies of Error Prone, and thus ~guaranteed to be present. I've also added a comment making this more explicit :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ErrorProneRuntimeClasspath.java
```diff
@@ -59,16 +59,22 @@
...
-      Pattern.compile(
-          "^(com\\.google\\.common\\..*|com\\.google\\.errorprone\\.([^\\.]+(?<!TestHelper)(\\..*)?)|java\\..*)$");
   private static final Matcher<ExpressionTree> GET_CANONICAL_NAME_INVOCATION =
       instanceMethod().onExactClass(Class.class.getCanonicalName()).named("getCanonicalName");
   private static final ImmutableSet<String> PRIMITIVE_TYPES =
       Primitives.allPrimitiveTypes().stream()
           .map(Class::getCanonicalName)
           .collect(toImmutableSet());
 
+  /**
```
mohamedsamehsalah, line 69: ❤️

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ErrorProneRuntimeClasspath.java
```diff
@@ -59,16 +59,22 @@
...
+   * A pattern that matches fully qualified type names that are expected to be runtime dependencies
```
mohamedsamehsalah, line 70: ❤️

rickie (commented):
Cool PR, sorry for delay in review :). 🚀

rickie (approved):
Added the commit as it can probably be improved. Other issue is not a blocker for me so already approving.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ErrorProneRuntimeClasspath.java
```diff
@@ -0,0 +1,165 @@
...
+    return fix.replace(
+            original,
+            identifier
+                + ".class.getCanonicalName()"
+                + (suffix.isEmpty() ? "" : (" + " + Constants.format(suffix))))
+        .build();
+  }
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
```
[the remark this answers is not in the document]
Stephan202: Indeed. Here I think the idea was to have two "groups" methods, one for each "direction", but I see now that this is not how it turned out. Will move :+1:

Stephan202 (commented):
Rebased and added a commit.

Ideally we move this check, and some like it, to a separate Maven module, so that downstream users won't have to disable it. But that's something we can/should pick up separately.

PR #883, opened 2023-11-13
Introduce `ConstantsFormat` Refaster rule

Stephan202:
Suggested commit message:
```
Introduce `ConstantsFormat` Refaster rule
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BugCheckerRules.java
```diff
@@ -52,4 +54,16 @@ BugCheckerRefactoringTestHelper after(
       return helper.addInputLines(path, source).expectUnchanged();
     }
   }
+
+  static final class ConstantsFormat {
```
rickie, line 59: Let's add a Javadoc? Will push a proposal.

rickie (approved):
Nice one!

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BugCheckerRules.java
Stephan202: Suggestion LGTM :+1:

PR #894, opened 2023-11-23
Introduce `dropwizard/metrics` integration test

oxkitsune:


integration-tests/metrics-v4.2.19.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
rickie, line 1: To ensure the `integration-test.sh` is working for both cases, it'd be nice to also rewrite the content of the `checkstyle-10.12.4.sh` file to sue the script.

integration-tests/integration-test.sh
```diff
@@ -0,0 +1,186 @@
...
+
+set -e -u -o pipefail
+
+integration_test_root="$(readlink -f "$(dirname "${0}")")"
+error_prone_support_root="${integration_test_root}/.."
+repos_root="${integration_test_root}/.repos"
+
+if [ "${#}" -ne 10 ]; then
+  >&2 echo "Usage $(basename "${0}") <test-name> <project> <repository> <revision> <build-flags> <patch-flags> <validation-ep-flags> <validation-mvn-flags> <do-sync> <report-directory>"
+
```
rickie, line 11:
```suggestion
```

integration-tests/metrics-v4.2.19-expected-changes.patch
```diff
@@ -0,0 +1,6660 @@
...
+-import java.util.Collections;
+ import java.util.Map;
+ import java.util.Set;
+ import java.util.SortedMap;
+@@ -80,7 +78,7 @@ public class CollectdReporter extends ScheduledReporter {
+     private SecurityLevel securityLevel = SecurityLevel.NONE;
+     private String username = "";
+     private String password = "";
+-    private Set<MetricAttribute> disabledMetricAttributes = Collections.emptySet();
++    private Set<MetricAttribute> disabledMetricAttributes = emptySet();
```
rickie, line 65: Interesting, are we not rewriting this to `ImmutableSet.of`?

integration-tests/metrics-v4.2.19-expected-warnings.txt
```diff
@@ -0,0 +1,8 @@
+metrics-core/src/main/java/com/codahale/metrics/CsvReporter.java:[382,36] [FormatStringConcatenation] Defer string concatenation to the invoked method
+metrics-graphite/src/main/java/com/codahale/metrics/graphite/GraphiteReporter.java:[429,18] [Slf4jLogStatement] Log statement contains 0 placeholders, but specifies 1 matching argument(s)
```
rickie, line 2: Oh nice, this will be interesting to fix upstream! As this is literally a bug!

integration-tests/integration-test.sh
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env bash
+
+set -e -u -o pipefail
+
+integration_test_root="$(readlink -f "$(dirname "${0}")")"
+error_prone_support_root="${integration_test_root}/.."
+repos_root="${integration_test_root}/.repos"
+
+if [ "${#}" -ne 10 ]; then
+  >&2 echo "Usage $(basename "${0}") <test-name> <project> <repository> <revision> <build-flags> <patch-flags> <validation-ep-flags> <validation-mvn-flags> <do-sync> <report-directory>"
```
rickie, line 10: For this, we should take a look at the other `.sh` files in this repository. I believe we have a slightly different naming situation here (which is closer to the official conventions I believe). 

So I _think_ it should be `[TestName] [Project] ....`

rickie (approved):
Try to make sure it runs for Java 11 and 17. We discussed this offline :).

Some comments but in general this looks really nice!

rickie:
(Sorry didn't mean to approve yet 😬.)

integration-tests/metrics-v4.2.19-expected-changes.patch
oxkitsune: `CollectorMutability` is disabled, because there's a few edge cases iirc. I can supress those edge cases :+1: and re-enable.

integration-tests/metrics-v4.2.19-init.patch
```diff
@@ -0,0 +1,203 @@
...
++        <profile>
++            <id>error-prone-compile</id>
++            <build>
++                <plugins>
++                    <plugin>
++                        <groupId>org.apache.maven.plugins</groupId>
++                        <artifactId>maven-compiler-plugin</artifactId>
++                        <version>${maven-compiler-plugin.version}</version>
++                        <executions>
++                            <!-- Skip the default-compile execution as we don't want to execute the compile goal twice -->
```
rickie, line 76:
```suggestion
+                            <!-- Skip the `default-compile` execution as we don't want to execute the compile goal twice. -->
```

Same below :).

integration-tests/metrics-v4.2.19-init.patch
```diff
@@ -0,0 +1,203 @@
...
+                                 <arg>-XDcompilePolicy=simple</arg>
+-                                <arg>-Xplugin:ErrorProne -XepExcludedPaths:.*/target/generated-sources/.*</arg>
+                                 <arg>-J--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED</arg>
+                                 <arg>-J--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED</arg>
+                                 <arg>-J--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED</arg>
+@@ -181,18 +181,129 @@
+                                 <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED</arg>
+                                 <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED</arg>
+                             </compilerArgs>
+-                            <annotationProcessorPaths>
```
rickie, line 55: It doesn't work if we leave this one here combined with the "append" usage in the other profile?

integration-tests/metrics-v4.2.19-init.patch
```diff
@@ -0,0 +1,203 @@
...
++            </build>
++        </profile>
+         <profile>
+             <id>release-sign-artifacts</id>
+             <activation>
+@@ -323,15 +434,7 @@
+                     <compilerArgs>
+                         <arg>-Xlint:all</arg>
+                         <arg>-XDcompilePolicy=simple</arg>
+-                        <arg>-Xplugin:ErrorProne -XepExcludedPaths:.*/target/generated-sources/.*</arg>
```
rickie, line 192: Do we use this? Otherwise I'd say we can leave this untouched to minimize the diff.

integration-tests/integration-test.sh
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env bash
+# This script is not meant to be invoked manually, instead it should be invoked
```
rickie, line 2:
```suggestion

# This script is not meant to be invoked manually, instead it should be invoked
```

rickie: Should we rename to `run-integration-test`? To make clearer what it does.

integration-tests/checkstyle-10.12.4.sh
```diff
@@ -1,11 +1,6 @@
 #!/usr/bin/env bash
-
```
rickie, line 2: Let's not drop this line.

.github/workflows/run-integration-tests.yml
```diff
@@ -32,7 +35,7 @@ jobs:
       - name: Install project to local Maven repository
         run: mvn -T1C install -DskipTests -Dverification.skip
       - name: Run integration test
-        run: xvfb-run ./integration-tests/checkstyle-10.12.4.sh "${{ runner.temp }}/artifacts"
+        run: xvfb-run ./integration-tests/${{ matrix.integration-test }} "${{ runner.temp }}/artifacts"
```
rickie, line 38: There is another mention of `checkstyle-10.12.4` below, so this will not work :).

rickie: If it's easy we should fix it, but if it's too involved we can defer.

integration-tests/integration-test.sh
```diff
@@ -0,0 +1,188 @@
...
+report_directory="${11}"
+
+if [ -n "${report_directory}" ]; then
+  mkdir -p "${report_directory}"
+else
+  report_directory="$(mktemp -d)"
+  trap 'rm -rf -- "${report_directory}"' INT TERM HUP EXIT
+fi
+
+# XXX: Configure Renovate to manage the AssertJ version declared here.
```
rickie, line 35: There is no AssertJ version here 😉, it's in the other file.

integration-tests/metrics-v4.2.19.sh
```diff
@@ -0,0 +1,21 @@
...
+
+if [ "${#}" -gt 2 ] || ([ "${#}" = 2 ] && [ "${1:---sync}" != '--sync' ]); then
+  echo "Usage: ${0} [--sync] [<report_directory>]"
+  exit 1
+fi
+do_sync="$([ "${#}" = 0 ] || [ "${1:-}" != '--sync' ] || echo 1)"
+report_directory="$([ "${#}" = 0 ] || ([ -z "${do_sync}" ] && echo "${1}") || ([ "${#}" = 1 ] || echo "${2}"))"
+
+# XXX: We exclude the CollectorMutability and the Immutable*|Preconditions*|StringRules.StringIsNullOrEmpty refaster rules
+# as they introduce changes that expect guava to be on the classpath.
```
rickie, line 18:
```suggestion
# as they introduce changes that expect Guava to be on the classpath.
```

integration-tests/metrics-v4.2.19.sh
```diff
@@ -0,0 +1,21 @@
...
+revision=v4.2.19
+
+if [ "${#}" -gt 2 ] || ([ "${#}" = 2 ] && [ "${1:---sync}" != '--sync' ]); then
+  echo "Usage: ${0} [--sync] [<report_directory>]"
+  exit 1
+fi
+do_sync="$([ "${#}" = 0 ] || [ "${1:-}" != '--sync' ] || echo 1)"
+report_directory="$([ "${#}" = 0 ] || ([ -z "${do_sync}" ] && echo "${1}") || ([ "${#}" = 1 ] || echo "${2}"))"
+
+# XXX: We exclude the CollectorMutability and the Immutable*|Preconditions*|StringRules.StringIsNullOrEmpty refaster rules
```
rickie, line 17:
```suggestion
# XXX: We exclude the `CollectorMutability` and the `Immutable*|Preconditions*|StringRules.StringIsNullOrEmpty` refaster rules
```

rickie: Instead, we also have this flag: https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java#L43 which is used in the CollectorMutability: https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CollectorMutability.java#L61 😉.

rickie (commented):
Some comments 😄.

integration-tests/metrics-v4.2.19-init.patch
oxkitsune: Leaving this here results in a ` error: plug-in not found: ErrorProne`

integration-tests/metrics-v4.2.19-init.patch
oxkitsune: Same as above

rickie:

rickie:
Rebased.

oxkitsune:
Checkstyle needs to be re-run, I will push changes when it's complete.

oxkitsune:
I'm unable to figure out why the refaster options aren't being used in the metrics integration test.
I've tried JDK11 to JDK21, and the exact options given in #654, but I cannot get the `ImmutableSet` rules to disable.

rickie:
Closing in favor of: https://github.com/PicnicSupermarket/error-prone-support/pull/1426.

PR #910, opened 2023-12-07
StaticImport: add various `com.fasterxml.jackson.annotation` values

arichapn:
Added are:

* `JsonCreator.Mode`
* `JsonFormat.Shape`
* `JsonInclude.Include`
* `JsonProperty.Access`


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
```diff
@@ -57,6 +57,7 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
   @VisibleForTesting
   static final ImmutableSet<String> STATIC_IMPORT_CANDIDATE_TYPES =
       ImmutableSet.of(
+          "com.fasterxml.jackson.annotation.JsonCreator.Mode",
```
EnricSala, line 60: There's more cases like:
- `@JsonFormat.Shape`.
- `@JsonTypeInfo.As`.
- `@JsonTypeInfo.Id`.
- `@JsonSubTypes.Type`.
- (probably more)

Do you think we could find a way to handle these in bulk @rickie? :)

Or would you recommend listing (and testing) each one individually.

rickie: No we do not need to test all of them. The check is already sufficiently tested.

We can simply add the members to the list. I'm not sure if we should just add all of them. We only add the ones where statically importing doesn't make the code "less readable/understandable", as in, we do not lose any context on what is happening. 

- The `JsonFormat.Shape` seems like a no-brainer that we should add. 
- For `.Id` I"m not sure: `@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS,` I'd say, `JsonTypeInfo` can be omitted but the `Id` part is maybe still "fine" to have? Same holds for:
```
@JsonTypeInfo(
    use = NAME,
    include = JsonTypeInfo.As.EXISTING_PROPERTY,
``` 
We can drop `JsonTypeInfo` but having `.As` is maybe better? This is a bit up for discussion I feel. WDYT?
- I'd argue that `@Type` is not that clear where it comes from. On the other hand, all usages are only nested in the `@JsonSubTypes` annotation (I think) and in that case it would make sense.

EnricSala: I agree with your proposal :+1:

arichapn: Just saw Enric's emoji response. 😄

Curious to hear your thoughts too @rickie, if this all sounds good then I'll proceed. 👍

rickie: I completely agree with the proposal of which ones to statically import. W.r.t. the ones that are still qualified but not statically imported, I'm not sure if that would fit in this check TBH. I think that should go in the `NonStaticImport` as there we qualify which types to qualify instead of statically import. Although I'm not sure if that also does this rewrite: `JsonTypeInfo.Id.X` to `Id.X` but it does make `X` -> `Id.X`. 😄

Nonetheless, I'd say it's worth checking out.

arichapn: Gotcha, in that case I'll just add the static import options to this PR. 

For the imports which remain (at least partially) qualified, I can keep it on my TODO list to check those out. 🙂

rickie (approved):
Thanks for filing a PR @arichapn :rocket:!

rickie:
Suggested commit message:
```
Require static import of various `com.fasterxml.jackson.annotation` members (#910)
```

Stephan202 (approved):
Rebased and added some tweaks.

Suggested commit message:
```
Require static importing of some `com.fasterxml.jackson.annotation` enums (#910)

Types to be imported statically:
- `com.fasterxml.jackson.annotation.JsonCreator.Mode`
- `com.fasterxml.jackson.annotation.JsonFormat.Shape`
- `com.fasterxml.jackson.annotation.JsonInclude.Include`
- `com.fasterxml.jackson.annotation.JsonProperty.Access`
```

PR #922, opened 2023-12-15
Disallow static import of `com.google.errorprone.{VisitorState,util.ASTHelpers}` members

rickie:
Saw this during a code review 😄.

Probably we can add more from Error Prone later. I checked some other classes from Error Prone but these are clear improvements that we also have been discussing before. Therefore, I decided to start with these two.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
@@ -61,8 +61,8 @@ public final class NonStaticImport extends BugChecker implements CompilationUnit
       ImmutableSet.of(
           "com.google.common.base.Strings",
           "com.google.errorprone.BugCheckerRefactoringTestHelper.TestMode",
-          "com.google.errorprone.util.ASTHelpers",
           "com.google.errorprone.VisitorState",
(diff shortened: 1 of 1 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
```diff
@@ -61,6 +61,8 @@ public final class NonStaticImport extends BugChecker implements CompilationUnit
       ImmutableSet.of(
           "com.google.common.base.Strings",
           "com.google.errorprone.BugCheckerRefactoringTestHelper.TestMode",
+          "com.google.errorprone.util.ASTHelpers",
+          "com.google.errorprone.VisitorState",
```
Stephan202: To match the GJF's import order:
```suggestion
          "com.google.errorprone.VisitorState",
          "com.google.errorprone.util.ASTHelpers",
```

Stephan202 (approved):
Rebased and added a commit. At least one Error Prone check has a suggested fix in which an `ASTHelpers` method is statically imported, so it's nice that this is now auto-fixed.

Suggested commit message:
```
Disallow static import of `com.google.errorprone.{VisitorState,util.ASTHelpers}` members (#922)
```

benhalasi (approved):

PR #937, opened 2023-12-23
Make `BugPattern{,Test}Extractor` tests more maintainable

Stephan202:
These change are backported from the `website` branch.

Suggested commit message:
```
Make `BugPattern{,Test}Extractor` tests more maintainable (#937)

As we're moving to a Java-based website generator located in the same
package as the `Extractor` implementations, there is no need to validate
the exact format of generated files; only that the data can be
deserialized again.

While there, track the source file from which data is extracted.
```


documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternExtractor.java
```diff
@@ -91,7 +94,36 @@ private static <T extends AnnotationValue> T doCast(AnnotationValue value, Class
...
+          name,
+          altNames,
+          link,
+          tags,
+          summary,
+          explanation,
+          severityLevel,
+          canDisable,
+          suppressionAnnotations);
+    }
```
Stephan202, line 123: This is boilerplate, but Error Prone flags `new AutoValue_BugPatternExtractor_BugPatternDocumentation` (and similar) calls outside the class in which the immutable type is defined.

documentation-support/src/main/java/tech/picnic/errorprone/documentation/Json.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Utility class that offers mutually consistent JSON serialization and deserialization operations,
+ * without further specifying the exact schema used.
+ */
+final class Json {
+  private static final ObjectMapper OBJECT_MAPPER =
+      new ObjectMapper()
+          .setVisibility(PropertyAccessor.FIELD, Visibility.ANY)
+          .registerModules(new GuavaModule(), new ParameterNamesModule());
```
Stephan202, line 21: This logic is moved to a separate class, not only because the tests use it, but also because other yet-to-backported code relies on it.

documentation-support/src/test/java/tech/picnic/errorprone/documentation/BugPatternExtractorTest.java
```diff
@@ -36,11 +40,25 @@ void minimalBugPattern(@TempDir Path outputDirectory) throws IOException {
...
+            "pkg.MinimalBugChecker",
+            "MinimalBugChecker",
+            ImmutableList.of(),
+            "",
+            ImmutableList.of(),
+            "MinimalBugChecker summary",
+            "",
+            ERROR,
+            /* canDisable= */ true,
+            ImmutableList.of(SuppressWarnings.class.getCanonicalName())));
```
Stephan202, line 57: Here and below: these objects replace the JSON resource files.

Stephan202 (commented):
Some context.

documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternExtractor.java
```diff
@@ -91,7 +94,36 @@ private static <T extends AnnotationValue> T doCast(AnnotationValue value, Class
   }
 
   @AutoValue
+  @JsonDeserialize(as = AutoValue_BugPatternExtractor_BugPatternDocumentation.class)
   abstract static class BugPatternDocumentation {
+    static BugPatternDocumentation create(
```
mohamedsamehsalah, line 99: Is there an order we are following for these parameters ?

Is it worth the hassle to find a better way to sort them ? Orthographicall or Objects > primities > lists for example ?

mohamedsamehsalah (approved):
Nothing stands out to me 🚀 

Only one comment about field order, but I am fine as-is ✅

documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternExtractor.java
Stephan202: The order we're following here matches the order in which the properties are defined. Those in turn are from "high level"/"most important" to "minor"/"more of a detail". It's not an exact science :smile:. Ordering them by type or name would IMHO not improve readability.

Stephan202 (commented):
Tnx for the quick review @mohamedsamehsalah! :rocket:

I added one more commit, as I forgot to include a test class.

rickie (approved):
💯

documentation-support/src/main/java/tech/picnic/errorprone/documentation/Json.java
```diff
@@ -0,0 +1,45 @@
...
+  static <T> void write(Path path, T object) {
+    try {
+      OBJECT_MAPPER.writeValue(path.toFile(), object);
+    } catch (IOException e) {
+      throw failure(e, "Failure writing to '%s'", path);
+    }
+  }
+
+  @FormatMethod
+  private static UncheckedIOException failure(IOException cause, String format, Object... args) {
```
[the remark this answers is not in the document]
Stephan202: Well, Java requires that the varargs argument is last (to avoid ambiguity). :)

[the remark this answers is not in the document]
rickie: 😂 oh ofcourse, forgot that. Let's merge!

PR #939, opened 2023-12-24
Prevent likely static import class with `FailWithMessage{,AndThrowable}` Refaster rules

Stephan202:
This resolves an issue observed internally.

Suggested commit message:
```
Prevent likely static import class with `FailWithMessage{,AndThrowable}` Refaster rules (#939)

This is a workaround for google/error-prone#3584. While there, drop an
unused method from `JUnitToAssertJRules`.
```

(It'd be better to fix google/error-prone#3584, but that looks to be (very) complicated; no time for that right now.)

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
@@ -16,7 +16,6 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertThrowsExactly;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import com.google.common.collect.ImmutableSet;
 import com.google.errorprone.annotations.DoNotCall;
 import com.google.errorprone.refaster.annotation.AfterTemplate;
 import com.google.errorprone.refaster.annotation.BeforeTemplate;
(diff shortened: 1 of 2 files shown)


rickie (approved):

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
```diff
@@ -78,8 +62,13 @@ T before(String message) {
       return Assertions.fail(message);
     }
 
+    // XXX: Add `@UseImportPolicy(STATIC_IMPORT_ALWAYS)` once
+    // https://github.com/google/error-prone/pull/3584 is resolved. Until that time, statically
+    // importing AssertJ's `fail` is likely to clash with an existing static import of JUnit's
+    // `fail`. Note that combining Error Prone's `RemoveUnusedImports` and
+    // `UnnecessarilyFullyQualified` checks and our `StaticImport` check will anyway cause the
```
Badbond, line 69: I was worried about the order of checks and potential clashes still, but luckily the `StaticImport` check [guards](https://github.com/PicnicSupermarket/error-prone-support/blob/12f2feaacfc819cd8ae2733396f933a5c155f67f/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java#L250) against that.

Badbond (approved):
LGTM :+1:

PR #963, opened 2024-01-12
Introduce `InputStreamRules` Refaster rule collection

Stephan202:
Suggested commit message:
```
Introduce `InputStreamRules` Refaster rule collection (#963)
```

Kudos to @koppor for suggesting this in openrewrite/rewrite-migrate-java#390.


Stephan202:
I also filed google/guava#6915.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/InputStreamRules.java
```diff
@@ -0,0 +1,50 @@
...
+    byte[] after(InputStream in) throws IOException {
+      return in.readAllBytes();
+    }
+  }
+
+  // XXX: Enable once we have a way to target JDK 12+ APIs.
+  // static final class InputStreamSkipNBytes {
+  //  @BeforeTemplate
+  //  void before(InputStream in, long n) throws IOException {
+  //    ByteStreams.skipFully(in, n);
```
Venorcis, line 42: `CharStreams` also features this (`Reader#skip` would be equivalent)

Venorcis (approved):
Nice 👍🏻

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/InputStreamRules.java
Stephan202: Good one!

mohamedsamehsalah (approved):
LGTM 👍

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/InputStreamRules.java
rickie: Not sure if I'm a fan of having Refaster rules like this commented out in the collection. A simple sentence that explains the rewrite we want to makes a bit more sense to me. 🤷🏻

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/InputStreamRules.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.google.common.io.ByteStreams;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/** Refaster rules related to expressions dealing with {@link InputStream}s. */
+final class InputStreamRules {
```
rickie, line 11: We are missing an `@OnlineDocumentation` annotation :).

rickie (changes requested):
Added a commit with some fixes. Its not enough for now, but I have to go.

We are missing the entry in `RefasterRulesTest`, so the tests are not executed.

 I found it weird that the output was before the input haha, so noticed something was wrong.

Sorry for not finalizing the review right now.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/InputStreamRules.java
Stephan202: Sure, can do :+1:

Stephan202 (commented):
Tnx for noticing @rickie! Pushed the fixes.

(And yes, I will use this opportunity once again to point out that we should prioritize automatic detection of these things. ;).)

rickie (approved):

koppor:
Thank you for implementing @Stephan202. I got some other feedback that using these methods on Android 9 and 10 could cause issues. Does anyone have test environments available? Maybe, this could be mentioned in JavaDocs?

Stephan202:
Hey @koppor! None of the code in this repo aims to be compatible with Android. We aim for (and [test](https://github.com/PicnicSupermarket/error-prone-support/blob/dff67fecbceccde17c4839e37c73dc66a9c65c0f/.github/workflows/build.yml#L11-L24) for) compatibility with JDK 11, 17 and 21. I suppose we could make this more explicit by adding a note to the README or extending the [compatibility matrix](https://error-prone.picnic.tech/compatibility/). I'll make a note to do so :)

PR #964, opened 2024-01-12
Build with Maven 3.9.6

Stephan202:
This should unblock PRs #434 and #491.

Suggested commit message:
```
Build with Maven 3.9.6 (#964)

Using the `setup-maven-action` GitHub action we can both simplify the
build configuration and configure the version of Maven to use.

See https://github.com/s4u/setup-maven-action
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,16 +30,12 @@ jobs:
       # on Maven Central, and once against the Picnic Error Prone fork,
       # additionally enabling all checks defined in this project and any Error
       # Prone checks available only from other artifact repositories.
-      - name: Check out code
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-        with:
-          persist-credentials: false
-      - name: Set up JDK
-        uses: actions/setup-java@387ac29b308b003ca37ba93a6cab5eb57c8f5f93 # v4.0.0
+      - name: Check out code and set up JDK and Maven
+        uses: s4u/setup-maven-action@fa2c7e4517ed008b1f73e7e0195a9eecf5582cd4 # v1.11.0
         with:
           java-version: ${{ matrix.jdk }}
-          distribution: ${{ matrix.distribution }}
-          cache: maven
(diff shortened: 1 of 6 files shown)


rickie (approved):
Niceeee, this solution is way better 🚀 !

.github/workflows/build.yml
```diff
@@ -30,16 +30,12 @@ jobs:
...
-        with:
-          persist-credentials: false
-      - name: Set up JDK
-        uses: actions/setup-java@387ac29b308b003ca37ba93a6cab5eb57c8f5f93 # v4.0.0
+      - name: Check out code and set up JDK and Maven
+        uses: s4u/setup-maven-action@fa2c7e4517ed008b1f73e7e0195a9eecf5582cd4 # v1.11.0
         with:
           java-version: ${{ matrix.jdk }}
-          distribution: ${{ matrix.distribution }}
-          cache: maven
```
mohamedsamehsalah, line 42: Why did we drop the cache ?

Did not find [anything](https://www.google.com/search?q=codeql+analysiss+and+cache&rlz=1C5CHFA_enNL1017NL1017&oq=codeql+analysiss+and+cache&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIJCAEQIRgKGKAB0gEINDc1OWowajeoAgCwAgA&sourceid=chrome&ie=UTF-8) that says this is enabled by default 🤔

mohamedsamehsalah (approved):
🚀 🚀

.github/workflows/build.yml
Stephan202: Good question. The implementation shows that the check [includes](https://github.com/s4u/setup-maven-action/blob/9337884575c7c64d6ed20d9ab3ee943658d49336/action.yml#L112-L119) a caching step, which is [enabled by default](https://github.com/s4u/setup-maven-action/blob/9337884575c7c64d6ed20d9ab3ee943658d49336/action.yml#L50-L54).

mohamedsamehsalah: Screenshot 2024-01-13 at 17 15 49
Sorry I was being blind 😮‍💨

PR #969, opened 2024-01-13
Introduce additional Reactor Refaster rules

Stephan202:
Suggested commit message:
```
Introduce additional Reactor Refaster rules (#969)

In various contexts, suggest more efficient and/or less verbose
constructs.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -5,7 +5,11 @@ import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.collect.MoreCollectors.toOptional;
 import static com.google.errorprone.BugPattern.SeverityLevel.WARNING;
 import static com.google.errorprone.refaster.ImportPolicy.STATIC_IMPORT_ALWAYS;
+import static java.util.Comparator.naturalOrder;
+import static java.util.Comparator.reverseOrder;
 import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.maxBy;
+import static java.util.stream.Collectors.minBy;
 import static java.util.stream.Collectors.toCollection;
 import static org.assertj.core.api.Assertions.assertThat;
 import static reactor.function.TupleUtils.function;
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1579,6 +1590,107 @@ Mono<ImmutableSet<T>> after(Flux<T> flux) {
...
+  /** Prefer {@link MathFlux#max(Publisher)} over less efficient alternatives. */
+  static final class FluxTransformMax<T extends Comparable<? super T>> {
+    @BeforeTemplate
+    Mono<T> before(Flux<T> flux) {
+      return flux.sort().last();
+    }
+
+    @AfterTemplate
+    Mono<T> after(Flux<T> flux) {
+      return flux.transform(MathFlux::max).next();
```
werli, line 1645: Food for thought: `MathFlux#max` returns a `Mono`, so the resulting `Flux` has either zero or one element. Should we make this more explicit but using `Flux#singleOrEmpty()`? This comment applies for more parts of the PR.

```suggestion
      return flux.transform(MathFlux::max).singleOrEmpty();
```

werli (commented):
Cool stuff

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: Nice suggestion; applied!

werli (approved):

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
werli: Maybe this opens up a box of other possible improvements.

Any conversion `Mono -> Flux -> Mono` for which we know at most one element emitted and that's currently using `Flux#next` should be rewritten? E.g. `flux.transform(xyz -> mono).next()` could call `Flux#singleOrEmpty()`.

Venorcis (approved):
LGTM 👍🏻

rickie (approved):
LGTM!

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -608,7 +618,7 @@ Flux<S> before(
     }
 
     @AfterTemplate
-    Flux<S> after(Mono<T> mono, Function<? super T, ? extends Iterable<? extends S>> function) {
```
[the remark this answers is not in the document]
Stephan202: Hmm, it could make sense, but using `I` instead of `? extends Iterable<? extends S>` would impose a greater restriction than necessary or desired: the rule should also work for any `Iterable` that is not a sub- _or_ super-type of `I`.

PR #972, opened 2024-01-13
Have `ErrorProneRuntimeClasspath` ignore non-public types

Stephan202:
~:exclamation: This PR is on top of #976. :exclamation:~

Suggested commit message:
```
Have `ErrorProneRuntimeClasspath` ignore non-public types (#972)
```

### DIFF
--- a/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibraryTest.java
+++ b/error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibraryTest.java
@@ -4,6 +4,7 @@ import static com.google.errorprone.BugPattern.SeverityLevel.ERROR;
 import static java.util.stream.Collectors.joining;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.errorprone.BugPattern;
 import com.google.errorprone.CompilationTestHelper;
 import com.google.errorprone.VisitorState;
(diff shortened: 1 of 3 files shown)


Stephan202:
Okay, the Pitest analysis now fails. Seems to be due to `junit-jupiter-engine` dependencies being dropped in #973. Two thoughts:
1. Would have been nice if CI had caught this, but that's perhaps too expensive to achieve.
2. I guess I should report this upstream, as it's counter-intuitive that Surefire doesn't require this dependency to be specified explicitly, while Pitest does.

Stephan202:
I filed pitest/pitest-junit5-plugin#105. Let's see whether there's a fast response; I can open a PR to revert most of #973 tomorrow.

Stephan202:
I rebased the branch on top of #976; the build should now pass :eyes:

Venorcis (approved):

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibraryTest.java
```diff
@@ -86,23 +87,64 @@ void isIntroductionAllowedIgnoreClasspathCompat(boolean ignoreClassPath) {
...
+   */
+  @BugPattern(severity = ERROR, summary = "Interacts with `ThirdPartyLibrary` for testing purposes")
+  public static final class CanIntroduceUsageTestChecker extends BugChecker
+      implements ClassTreeMatcher {
+    private static final long serialVersionUID = 1L;
+    private static final ImmutableMap<String, String> TYPES =
+        ImmutableMap.of(
+            "GUAVA_PUBLIC",
+            ImmutableList.class.getCanonicalName(),
+            "GUAVA_PRIVATE",
```
rickie, line 133: Depends on how you look at it, but strictly speaking this type is not `private`, right?

rickie (approved):
Nice improvement! 🚀

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibraryTest.java
Stephan202: Indeed, it's package-private, but here I meant private-to-the-library/implementation.

PR #979, opened 2024-01-16
Use `plus###` and `minus###` shorthands when possible for dates/times

Venorcis:
Uses e.g. `LocalDate#plusDays` instead of more contrived alternatives.
Quite a big PR as sadly none of the involved classes share a common interface for the shorthand methods.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
@@ -582,4 +582,898 @@ final class TimeRules {
       return Period.ZERO;
     }
   }
+
+  /** Prefer {@link LocalDate#plusDays} (and variants) over more contrived alternatives. */
+  static final class LocalDatePlusDays {
+    @BeforeTemplate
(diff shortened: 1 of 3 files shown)


Stephan202 (commented):
When you asked me offline whether we could do this better, I didn't expect there to be _this many_ cases, haha. Perhaps an Error Prone check _would_ make sense, but it might be nice to get this merged first. (There are also other Refaster rules that may benefit from generalization.)

Stephan202:
Added two commits.

Suggested commit message:
```
Extend `TimeRules` Refaster rule collection (#979)

By introducing `SomeDateType{Plus,Minus}SomeUnit` Refaster rules, that suggest
e.g. `LocalDate#plusDays(long)` over more contrived alternatives.
```

Stephan202 (approved):

Venorcis:
> Added two commits.
> 
> Suggested commit message:

Thanks @Stephan202! 🦸🏻‍♂️

rickie (approved):
Used my brain to do a lot of pattern matching during the review 🧠 .

Nice work guys 😄.

rickie:
Nice suggested commit message @Stephan202 !

PR #986, opened 2024-01-22
Actually skip test failures during SonarCloud analysis

Stephan202:
Extracted from #603.

Suggested commit message:
```
Actually skip test failures during SonarCloud analysis (#986)
```

To test, run:
```sh
vimdiff \
  <(mvn help:effective-pom) \
  <(mvn help:effective-pom -Dsonar.projectKey=PicnicSupermarket_error-prone-support)
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1981,6 +1981,11 @@
         </profile>
         <profile>
             <id>sonar</id>
+            <activation>
+                <property>
+                    <name>sonar.projectKey</name>
+                </property>
+            </activation>
             <properties>
                 <maven.test.failure.ignore>true</maven.test.failure.ignore>
             </properties>


mohamedsamehsalah (approved):
> To test, run:
```sh
vimdiff \
  <(mvn help:effective-pom) \
  <(mvn help:effective-pom -Dsonar.projectKey=PicnicSupermarket_error-prone-support)
```

Don't have experience with `vimdiff` so not sure what I am supposed to be looking at here 🙈

Stephan202:
It's not a question of `vimdiff`; any diff would suffice. It should should that on this branch `maven.test.failure.ignore` is set only in the latter case. (While without this PR there'd be no diff.)

I do see that I failed to add a backslash to the command; try again :)

rickie (approved):

PR #992, opened 2024-01-25
Disallow static import of `{Mono,Flux}` and partially allow `empty`

rickie:
Suggested commit message:
```
Disallow static import of `{Mono,Flux}` and partially allow `empty` (#XXX) 

As exception, disallow static imports of `Optional#empty`. 
```

Slightly different then what we discussed. Instead of only adding `{Mono,Flux}#empty`, I figured we want to basically never import something from `Mono` and `Flux`. For `Mono` we have ~26 static imports internally.


error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/StaticImportTest.java
```diff
@@ -99,6 +100,8 @@ void identification() {
             "    // BUG: Diagnostic contains:",
             "    MediaType t2 = MediaType.APPLICATION_JSON;",
             "",
+            "    // BUG: Diagnostic contains:",
+            "    Filters.empty();",
             "    Optional.empty();",
```
mohamedsamehsalah, line 105: Tests still pass if I remove line 105, what is the purpose of this line ?
```suggestion
```

mohamedsamehsalah (commented):
Thanks 💪

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/StaticImportTest.java
rickie: To show that we are not changing the `Optional.empty()` 😄, so this is expected.

Venorcis (approved):
👍🏻

cernat-catalin:
Changes look good but are we sure we want we want to disallow all static imports for `Mono` and `Flux` (not saying I disagree, just want to make sure we make the correct decision).

Most of the present use cases seem to fall into this pattern
```java
  public Mono<BankAccountDto> findBankAccountByPicnicId(@PathVariable String picnicId) {
    return bankAccountService
        .findByEmployeePicnicId(picnicId)
        .flatMap(this::mapBankAccount)
        .switchIfEmpty(error(itemNotFound("Bank Data of Employee", picnicId)));
  }
```

Which doesn't look that bad to me (but consistency might justify disallowing it).

rickie:
Good point, when I was looking to usages in our codebase I couldn't find many other edge cases. 

We could of course decide to only allow `error` and not the rest 🤔.

cernat-catalin:
> We could of course decide to only allow error and not the rest

Feels a bit random. In the example with the controller it is clear that `error` refers to a `Mono` since you can clearly see the whole function signature and that is the last step.

But now that I think more about it maybe sometimes it is not clear at all

```java
    Flux.range(0, 100)
            .map(x -> error(...));
```

Is that `Mono.error()`, `Flux.error`, something else? I'm not sure it makes a difference but still.

cernat-catalin (approved):

Stephan202 (approved):
Rebased. Alternative suggested commit message:
```
Update `{,Non}StaticImport` checks (#992)

Summary of changes:
- Disallow static import of `Mono` and `Flux` members.
- Allow static import of `empty` methods.
- Disallow static import of `Optional#empty`.
```

@mohamedsamehsalah anything else from your side?

mohamedsamehsalah (approved):
Nothing else from my side ✅ 

Thanks for checking @Stephan202 🙏

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/StaticImportTest.java
mohamedsamehsalah: I also realise how dumb this comment was 🤦 (Haven't written EPS tests in a while)

PR #996, opened 2024-01-26
Allow static imports of `{ZERO,ONE}` identifiers

rickie:
We think that in the contexts where this happen it usually _does_ make sense to statically import these identifiers.


mohamedsamehsalah (approved):
Makes sense 🔢

Stephan202 (approved):
Suggested commit message:
```
Allow static imports of `ZERO` and `ONE` identifiers (#996)
```

PR #999, opened 2024-01-27
Sync Checkstyle integration test

Stephan202:
The two commits are independently reviewable.

Suggested commit message:
```
Sync Checkstyle integration test (#999)

Summary of changes:
- Update the set of expected changes for compatibility with the latest
  features, except for the `ErrorProneRuntimeClasspath` check.
- Test against version 10.12.7 rather than 10.13.0.
- Omit the targeted tag from file names, so that similar upgrade PRs can be
  tested using an `/integration-test` GitHub comment.
```


Stephan202:
The `/integration-test` build [fails](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/7678996750/job/20929421466?pr=999), because (IIUC) the GitHub action config is derived from `master`. Perhaps we should rename the scripts to omit the exact tag used. Assuming we don't want to track multiple tags per project, this will simplify things.

Stephan202:
Added a commit and updated the suggested commit message. Do still need to validate these changes locally.

rickie:
Cool stuff! At least locally it now runs for me 😄. Already updated #867 with the latest state of this PR.

Stephan202:
I noticed that with my changes at some point the build would fail with:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.12.1:compile (error-prone-compile) on project checkstyle: Fatal error compiling: AmbiguousJsonCreator is not a valid checker name -> [Help 1]
```

I added a commit to fix that.

Stephan202:
K, validated the latest changes. PR is now properly ready for review.

Stephan202:
Rebased and updated the PR to target version 10.13.0. Suggested commit message updated.

rickie (approved):
Thanks for bringing this one up-to-date! 

Smart move to drop the version from the file names :).

Ran it locally and it works :).

mohamedsamehsalah (commented):
Is the goal here to run `/integration-tests/checkstyle.sh` succesfully ?
Because It is not running for me (macOs).
```
usage: paste [-s] [-d delimiters] file ...
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
```

also fails with `--sync` flag. What am I missing ?

mohamedsamehsalah (approved):
Clear 👍

rickie:
Rebased,will merge!

PR #1000, opened 2024-01-27
Introduce `error-prone-utils` module

rickie:
I think I like `utilities` a bit more as it sounds a bit more "official" as module name. However, that's maybe not too strong of a reasoning 😄. Though, I'm also fine with this.

I'm doubting if this would benefit from a README, or not really 🤔.


error-prone-utils/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+        <dependency>
+            <groupId>${groupId.error-prone}</groupId>
+            <artifactId>error_prone_check_api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>${groupId.error-prone}</groupId>
+            <artifactId>error_prone_core</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
Stephan202: Surprisingly, it seems that this one isn't necessary.

Stephan202 (approved):
Rebased and added a small commit. I like `utils`, because (anecdotally) it's a more common Maven artifactId suffix than `utilities`.

Suggested commit message:
```
Introduce `error-prone-utils` module (#1000)

By extracting the `tech.picnic.errorprone.bugpatterns.util` package into a new 
Maven module (and renaming the package to `tech.picnic.errorprone.utils`), the 
utility classes it contains will also become usable by other Maven modules.
```

Stephan202:
W.r.t. the label: seems `chore` to me; there's no value unlocked by this PR just yet.

mohamedsamehsalah (approved):
🚀 🚀

PR #1002, opened 2024-01-28
Introduce `error-prone-guidelines` module

rickie:
:exclamation: This PR is on top of #1000. :exclamation:

In the end I decided to defer the Refaster rules for a bit. Let's get this first step in 💪🏻.


error-prone-guidelines/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+    <parent>
+        <groupId>tech.picnic.error-prone-support</groupId>
+        <artifactId>error-prone-support</artifactId>
+        <version>0.14.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>error-prone-guidelines</artifactId>
+
+    <name>Picnic :: Error Prone Support :: Error Prone Guidelines</name>
+    <description>Error Prone checks defining coding guidelines for the Error Prone project.</description>
```
Stephan202:
```suggestion
    <description>Coding guidelines for the Error Prone Support project.</description>
```
(This to (a) prepare for moving Refaster rules here and (b) because effectively these rules only apply to this project.)

error-prone-guidelines/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>error-prone-utils</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.auto.service</groupId>
+            <artifactId>auto-service-annotations</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
Stephan202, line 60: We can drop a few more dependencies; will push.

error-prone-guidelines/README.md
```diff
@@ -0,0 +1,5 @@
+# Error Prone coding guidelines
+
+This module provides Error Prone checks that describe and suggest coding
+guidelines for both Error Prone checks and Refaster rules. The rules should
+only be applied on Error Prone code.
```
Stephan202: I'll make some suggestions.

error-prone-guidelines/src/main/java/tech/picnic/errorprone/bugpatterns/AssociativeMethodInvocation.java
Stephan202, line 1: Strictly speaking this check didn't need to be moved; will add a comment.

Stephan202: Just like in the other PR we should selected a new package. Proposal: `tech.picnic.errorprone.guidelines.bugpatterns`.

Stephan202 (approved):
Rebased and added a commit. One could argue that `error-prone-support-guidelines` would be a better module name, but somehow it feels off. Another option is to go for just `coding-guidelines`, but I'm not sold on that either :shrug:

Suggested commit message:
```
Introduce `error-prone-guidelines` module (#1002)

And move all applicable checks to this new module.
```

mohamedsamehsalah (approved):
🤓 🤓 🤓

Stephan202:
Rebased and resolved conflicts. @rickie I think we can merge this one, once built.

PR #1003, opened 2024-01-28
Introduce `error-prone-experimental` module

rickie:
:exclamation: This PR is on top of #1000. :exclamation:


error-prone-experimental/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+    <parent>
+        <groupId>tech.picnic.error-prone-support</groupId>
+        <artifactId>error-prone-support</artifactId>
+        <version>0.14.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>error-prone-experimental</artifactId>
+
+    <name>Picnic :: Error Prone Support :: Experimental</name>
+    <description>Experimental Error Prone checks.</description>
```
rickie, line 14: Had no better idea as my alternatives all got a bit too long.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/IsInstanceLambdaUsage.java
```diff
@@ -24,7 +24,7 @@
  * A {@link BugChecker} that flags lambda expressions that can be replaced with a method reference
  * of the form {@code T.class::isInstance}.
  *
- * @see MethodReferenceUsage
+ * <p>See the `MethodReferenceUsage` check in the `error-prone-experimental` module.
```
Stephan202: Here and below: instead of backticks we should use `{@code ..}`. But since there's a corresponding `XXX` comment: let's drop.

error-prone-experimental/README.md
```diff
@@ -0,0 +1,6 @@
+# Experimental Error Prone checks
+
+This module contains Error Prone checks that are currently under development or
+evaluation. These checks may be works-in-progress or have uncertain impacts on
+code. Having this module allows for controlled experimentation and refinement
+before integration into production environments.
```
Stephan202:
```suggestion
This module contains Error Prone checks that are currently under development or
evaluation. These checks may be works-in-progress or have uncertain impact on
code. Having this module allows for controlled experimentation and refinement
before integration into production environments.
```

error-prone-experimental/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jspecify</groupId>
+            <artifactId>jspecify</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
Stephan202: There are a few more unused dependencies here; will drop.

error-prone-experimental/src/test/java/tech/picnic/errorprone/bugpatterns/MethodReferenceUsageTest.java
Stephan202, line 1: We should change the package name, in preparation for one day adopting the Java module system (which doesn't allow packages to be split between modules).

I'm proposing `tech.picnic.errorprone.experimental.bugpatterns`.

Stephan202 (approved):
```
Introduce `error-prone-experimental` module (#1003)

And move the known-incomplete `MethodReferenceUsage` check to it.
```

Issue #1001, opened 2024-01-28
Rewrite `String#copyValueOf(char[])` to `new String(char[])`


rickie:
### Problem

Rewrite `String#copyValueOf(char[])` to `new String(char[])`.

### Description of the proposed new feature

- [x] Support a stylistic preference.
- [ ] Avoid a common gotcha, or potential problem.
- [ ] Improve performance.

The Refaster rule should live in the `StringRules` file. 
Make sure to add a test entry in the `StringRulesTestInput` and StringRulesTestOutput 😄.

I would like to rewrite the following code:
```java
char[] foo = new char[0]; 
String s = String.copyValueOf(foo);
```

to:
```java
char[] foo = new char[0]; 
String s = new String(foo);
```

BLasan:
Hi,

Will start working on this.

Thanks,
Benura

BLasan:
HI @rickie ,

Is there a slack channel or any other way to communicate installation issues? I'm facing some issues when running `mvn clean install` regarding some jars not found.

Thanks,
Benura

BLasan:
Hi @rickie ,

When running `mvn clean install`, it gives the below error.
`
Caused by: org.eclipse.aether.resolution.ArtifactResolutionException: The following artifacts could not be resolved: tech.picnic.error-prone-support:error-prone-contrib:jar:2.24.1 (absent), tech.picnic.error-prone-support:refaster-runner:jar:2.24.1 (absent): tech.picnic.error-prone-support:error-prone-contrib:jar:2.24.1 was not found in https://repo.maven.apache.org/maven2 during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of central has elapsed or updates are forced`

I'm using maven `3.9.5` and java `11.0.16`.

Thanks,
Benura

Stephan202:
Hey @BLasan! Unfortunately we don't have a Slack channel. But we can use this PR to sync :)

W.r.t. your question: it looks like you mixed up some version numbers: version `2.24.1` is the version of Error Prone, while the `tech.picnic.error-prone-support:*` artifacts for which you get are error are for Error Prone **Support** (this project), of which the most recent version is `0.14.0`.

That said: if you tried to build an unmodified version of this project, then it shouldn't be necessary to modify `pom.xml` at all. If you're still stuck, feel free to share the code you have so far :)

BLasan:
Hi @Stephan202 ,

[...]

PR #1012, opened 2024-02-03
Introduce `NewStringFromCharArray{,SubSequence}` Refaster rules

BLasan:
fixes: https://github.com/PicnicSupermarket/error-prone-support/issues/1001

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
@@ -163,18 +163,35 @@ final class StringRules {
   }
 
   /**
-   * Prefer direct invocation of {@link String#copyValueOf(char[])} over the indirection introduced
-   * by {@link String#copyValueOf(char[])}.
+   * Prefer direct invocation of {@link String#String(char[], int, int)} over the indirection
+   * introduced by alternatives.
    */
-  static final class StringCopyValueOf {
+  static final class NewStringFromCharArraySubSequence {
     @BeforeTemplate
     String before(char[] data, int offset, int count) {
-      return String.copyValueOf(data, offset, count);
+      return Refaster.anyOf(
+          String.valueOf(data, offset, count), String.copyValueOf(data, offset, count));
     }
 
     @AfterTemplate
     String after(char[] data, int offset, int count) {
       return new String(data, offset, count);
+    }
+  }
+
+  /**
+   * Prefer direct invocation of {@link String#String(char[])} over the indirection introduced by
+   * alternatives.
+   */
+  static final class NewStringFromCharArray {
+    @BeforeTemplate
+    String before(char[] data) {
+      return Refaster.anyOf(String.valueOf(data), new String(data, 0, data.length));
+    }
+
+    @AfterTemplate
+    String after(char[] data) {
+      return new String(data);
     }
   }
 
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
@@ -163,6 +163,22 @@ final class StringRules {
   }
 
   /**
+   * Prefer direct invocation of {@link String#copyValueOf(char[])} over the indirection introduced
+   * by {@link String#copyValueOf(char[])}.
+   */
+  static final class StringCopyValueOf {
+    @BeforeTemplate
+    String before(char[] data, int offset, int count) {
+      return String.copyValueOf(data, offset, count);
+    }
+
+    @AfterTemplate
+    String after(char[] data, int offset, int count) {
+      return new String(data, offset, count);
+    }
+  }
+
+  /**
    * Prefer direct delegation to {@link String#valueOf(Object)} over the indirection introduced by
    * {@link Objects#toString(Object)}.
    */
--- a/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StringRulesTestInput.java
+++ b/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StringRulesTestInput.java
@@ -73,6 +73,18 @@ final class StringRulesTest implements RefasterRuleCollectionTestCase {
     return Objects.toString("foo");
   }
 
+  ImmutableSet<String> testNewStringFromCharArraySubSequence() {
+    return ImmutableSet.of(
+        String.valueOf(new char[] {'f', 'o', 'o'}, 0, 1),
+        String.copyValueOf(new char[] {'b', 'a', 'r'}, 2, 3));
+  }
+
+  ImmutableSet<String> testNewStringFromCharArray() {
+    return ImmutableSet.of(
+        String.valueOf(new char[] {'f', 'o', 'o'}),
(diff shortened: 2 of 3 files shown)


BLasan:
@rickie @Stephan202 Please review. I have added the rule to copy string characters with an offset and a count. I tried the below way to support both copying all characters and copying characters with an offset[1]. But it gives the following error[2]

[1]     ```
@AfterTemplate
    String after(char[] data, int offset, int count) {
      return Refaster.anyOf(new String(data, offset, count), new String(data));
    }
```
[2] `Caused by: java.lang.IllegalArgumentException: @AfterTemplate of tech.picnic.errorprone.refasterrules.StringRules.StringCopyValueOf defines arguments that are not present in all @BeforeTemplate methods: [offset, count]`

Is there a specific way to overcome this? Or do we need another class to be initiated?

Thanks,
Benura

**Comment by Stephan202** [comment]
Hey @BLasan! Thanks; I'll have a look this weekend.

To answer your question: Refaster has some constraints. Relevant to your example are:
1. `Refaster.anyOf` can only be used in `@BeforeTemplate` methods.
2. Any parameter to an `@AfterTemplate` method must also be specified (and bound) by all `@BeforeTemplate` methods.

The `new String(data)` case would require a separate rule that rewrites (off the top of my head) `String.copyValueOf(data, 0, data.length)`. I think that makes sense to do!

File: error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java [hunk 1]
```diff
@@ -162,6 +162,22 @@ String after(Object object) {
     }
   }
 
+  /**
+   * Prefer direct invocation of {@link String#copyValueOf(char[])} over the indirection introduced
+   * by {@link String#copyValueOf(char[])}.
+   */
```
Stephan202: This sentence contains a copy-paste error :eyes:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
```diff
@@ -162,6 +162,22 @@ String after(Object object) {
...
+  static final class StringCopyValueOf {
+    @BeforeTemplate
+    String before(char[] data, int offset, int count) {
+      return String.copyValueOf(data, offset, count);
```
Stephan202: I see that we can also rewrite a `String.valueOf` variant; will propose something.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StringRulesTestInput.java
```diff
@@ -84,4 +84,8 @@ String testSubstringRemainder() {
   int testUtf8EncodedLength() {
     return "foo".getBytes(UTF_8).length;
   }
+
+  String testStringCopyValueOf() {
+    return String.copyValueOf(new char[] {'f', 'o', 'o'}, 0, 3);
+  }
```
Stephan202: We generally order the test methods the same as the rules; I'll move these test methods up :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
```diff
@@ -162,6 +162,22 @@ String after(Object object) {
     }
   }
 
+  /**
+   * Prefer direct invocation of {@link String#copyValueOf(char[])} over the indirection introduced
+   * by {@link String#copyValueOf(char[])}.
+   */
+  static final class StringCopyValueOf {
```
Stephan202: We generally name the rule after the `@AfterTemplate` code. I this case it's a bit tricky, but I'll propose something.

Stephan202 (approved):
I added a commit with some suggestions, including an extra rule.

Suggested commit message:
```
Introduce `NewStringFromCharArray{,SubSequence}` Refaster rules (#1012)
```

@rickie will also review the PR, but as he's off, that may take a week or so.

Thanks for your contribution @BLasan!

BLasan:
Hi @Stephan202 ,

Thank you for the review. Are there any resources to learn the procedure of adding new rules? Also, any other issues that I could work on :) ?

Thanks,
Benura

Stephan202:
> Are there any resources to learn the procedure of adding new rules?

That's a good question. I think that right now there's no documentation for this. What we instead try to do, is add Error Prone checks that validate and (when possible) attempt to auto-fix rules to match our guidelines. PR #1002 moves some of those rules around, but there are (vague) plans to add more such rules.

For now the best I can offer is to employ careful pattern matching by looking at existing rules :smile:

> Also, any other issues that I could work on :) ?

Certainly our work isn't done :wink:. The list of [currently open issues](https://github.com/PicnicSupermarket/error-prone-support/issues?q=is%3Aissue+is%3Aopen) could serve as inspiration, but note that most/all of those will require writing an Error Prone `BugChecker`, rather than a Refaster rule. That's quite a bit more involved, but if you're feeling adventurous: go for it :smile:.

But let me offer another idea: are there aspects of your company's code base that you'd like to see different, such as things that you (perhaps repeatedly) point out during code review? If any of those are generally applicable and solvable using Refaster, we're certainly open to considering rules for that. (This suggestion does come with a caveat: we're kind of opinionated on _some_ topics, such as preferring immutablility. But when in doubt: just ask!)

BLasan:
Hi @Stephan202 ,

Sure, will take a look at some other issues opened. I asked specifically for an issue since someone internally (from your team) might have started working on such issues :) I need to recall some of the ideas (which are solvable using Refaster)  pointed out in a code review as well as during the implementations. I'll just revisit some of the feature implementations and open issues if found any :)

Thanks,
Benura

Venorcis (approved):
LGTM 👍🏻

rickie (approved):
Thanks for the Refaster rules @BLasan! 🚀 

> Also, any other issues that I could work on :) ?

@Stephan202 and I discussed something that could be a nice thing to work on. In this PR #770 that we are about to merge, we want to make clear for type migrations which methods are not (yet) migrated. We of course want to minimize the number of methods that are not migrated, see an example of the migration from JUnit Assertions to AssertJ [here](https://github.com/PicnicSupermarket/error-prone-support/pull/770/files#diff-82a73349e4d1a89787c77dd926152d354e0b617058be0ae2f41cd6ee498ff231R46).

If working on Refaster rules increase the coverage of an JUnit Assertions -> AssertJ assertions migration sound interesting to you, you can take a few of those list. Add a few of the Refaster rules, create a PR and we can review them. If we do a few rules per PR the PR will not become too big. We can start with a few simple ones to go through this process the first time, we make sure to review it in an elaborate way to help you better understand Refaster 😄. That will make it easier for you to add new things and on our side easier to reviewing following rules. 

For example the following methods could be a nice start to migrate from `org.junit.jupiter.api.Assertions` to their AssertJ equivalent:

```
"assertFalse(BooleanSupplier)",
"assertFalse(BooleanSupplier, String)",
"assertFalse(BooleanSupplier, Supplier<String>)",
"assertTrue(BooleanSupplier)",
"assertTrue(BooleanSupplier, String)",
"assertTrue(BooleanSupplier, Supplier<String>)",
```

Something that might not be clear is that in the `JUnitToAssertJRules.java` is that the Refaster rules are sorted by the order in which the methods we migrate away from are in the `org.junit.jupiter.api.Assertions` class. 

But, don't worry we are of course here to help and guide you if you think this sounds like a nice thing! Don't hesitate to open a discussion, draft PR or an answer on this PR 😄.

Let us know what you think :).

PR #1024, opened 2024-02-10
Introduce `OptionalOrElse` check

Stephan202:
I found a branch from back in 2021; decided to dust it off :)

Suggested commit message:
```
Introduce `OptionalOrElse` check (#1024)

While there, extend the `OptionalIdentity` Refaster rule to
automatically resolve one class of `NestedOptionals` violations.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
@@ -386,9 +386,13 @@ final class OptionalRules {
    */
   static final class OptionalIdentity<T> {
     @BeforeTemplate
+    @SuppressWarnings("NestedOptionals" /* Auto-fix for the `NestedOptionals` check. */)
     Optional<T> before(Optional<T> optional, Comparator<? super T> comparator) {
       return Refaster.anyOf(
           optional.or(Refaster.anyOf(() -> Optional.empty(), Optional::empty)),
+          optional
+              .map(Optional::of)
+              .orElseGet(Refaster.anyOf(() -> Optional.empty(), Optional::empty)),
           optional.stream().findFirst(),
           optional.stream().findAny(),
           optional.stream().min(comparator),
@@ -442,9 +446,7 @@ final class OptionalRules {
   static final class OptionalStream<T> {
     @BeforeTemplate
     Stream<T> before(Optional<T> optional) {
-      return Refaster.anyOf(
-          optional.map(Stream::of).orElse(Stream.empty()),
-          optional.map(Stream::of).orElseGet(Stream::empty));
+      return optional.map(Stream::of).orElseGet(Stream::empty);
     }
 
     @AfterTemplate
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
@@ -360,7 +360,7 @@ final class OptionalRules {
   /** Prefer {@link Optional#or(Supplier)} over more verbose alternatives. */
   static final class OptionalOrOtherOptional<T> {
     @BeforeTemplate
-    @SuppressWarnings("NestedOptionals" /* Auto-fix for the `NestedOptionals` check. */)
+    @SuppressWarnings("NestedOptionals")
     Optional<T> before(Optional<T> optional1, Optional<T> optional2) {
       // XXX: Note that rewriting the first and third variant will change the code's behavior if
       // `optional2` has side-effects.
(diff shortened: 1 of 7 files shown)


Stephan202:
Rebased and added a commit to utilize JDK 17 functionality.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/OptionalOrElseTest.java
```diff
@@ -0,0 +1,135 @@
...
+            "  }",
+            "",
+            "  private <T> T foo() {",
+            "    return null;",
+            "  }",
+            "",
+            "  private <S, T> T bar() {",
+            "    return null;",
+            "  }",
+            "}")
```
mohamedsamehsalah, line 101: If I do this 
```suggestion
            "    optional.orElseGet(() -> new int[0].length);",
            "    optional.orElseGet(() -> B.NUMBER);",
            "    optional.orElseGet(B::getNumber);",
            "  }",
            "",
            "  private <T> T foo() {",
            "    return null;",
            "  }",
            "",
            "  private <S, T> T bar() {",
            "    return null;",
            "  }",
            "",
            "  class B {",
            "    static final int NUMBER = 1;",
            "",
            "    static String getNumber() {",
            "      return \"number\";",
            "    }",
            "  }",
            "}")
```

Shouldn't we expect `optional.orElseGet(() -> B.NUMBER)` to be `optional.orElseGet(() -> B.NUMBER)` ?

This is what I get instead 
Screenshot 2024-02-16 at 17 03 47

mohamedsamehsalah (commented):
Sorry for slow review 🐢 

I have some questions 🙌

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/OptionalOrElseTest.java
Stephan202: This is intended :). The idea here is that field dereferencing (`foo.bar.baz` etc.) is _very_ efficient (conceptually: just some pointer dereferencing), so there's no point in deferring that work.

Stephan202 (commented):
Tnx for reviewing @mohamedsamehsalah! :pray:

mohamedsamehsalah (approved):
Thanks for answering my questions @Stephan202 🚀

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -386,9 +386,13 @@ Optional<T> after(Optional<T> optional1, Optional<T> optional2) {
    */
   static final class OptionalIdentity<T> {
     @BeforeTemplate
+    @SuppressWarnings("NestedOptionals" /* Auto-fix for the `NestedOptionals` check. */)
```
rickie, line 389:
```suggestion
    @SuppressWarnings("NestedOptionals" /* Auto-fix for the `NestedOptionals` violation. */)
```
Not wrong, but this is an alternative.

rickie (approved):
Slightly tweaked suggested commit message.

Nice PR 🚀 !

Curious to see if it has an impact on the Checkstyle integration test, so added a comment.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
Stephan202: Hmm, I'm not feeling this suggestion, TBH. Elsewhere we use `/* This violation will be rewritten. */`, perhaps that would work? (If so, we should do the same for `OptionalOrOtherOptional`.)

Stephan202 (commented):
Tnx for the typo fix in the suggested commit message!

integration-tests/checkstyle-expected-changes.patch
```diff
@@ -53686,6 +53686,15 @@
...
      final Path path = Paths.get(XdocUtil.DIRECTORY_PATH + "/config.xml");
+@@ -1081,7 +1083,7 @@ public class XdocsPagesTest {
+         Optional.ofNullable(field)
+             .map(nonNullField -> nonNullField.getAnnotation(XdocsPropertyType.class))
+             .map(propertyType -> propertyType.value().getDescription())
+-            .orElse(fieldClass.getSimpleName());
++            .orElseGet(fieldClass::getSimpleName);
+     final String expectedValue =
+         getModulePropertyExpectedValue(sectionName, propertyName, field, fieldClass, instance);
+
```
Stephan202, line 53697: With this change the integration test [passes](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/8407845211).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
Stephan202: ^ So far `/* This violation will be rewritten. */` is used only for SonarCloud violations. In e.g `ReactorRules` and `TimeRules` I see that `NestedPublishers` and `TimeZoneUsage` violations are suppressed with out a comment. Likely we should add some text in each case, but that's for another PR, then. Will drop it here.

Stephan202 (commented):
Added a commit.

rickie:
Thanks for fixing the test and answering my questions :). Code LGTM, let's merg!

PR #1025, opened 2024-02-10
Extend `StreamIsEmpty` Refaster rule

giovannizotta:
Introduces Refaster rule(s) that avoid unnecessary collections when checking whether a stream is empty.

Quick peek at the extension introduced:
`stream.collect(collectingAndThen(collector, Collection::isEmpty));` ➡️ `stream.findAny().isEmpty();`

1. ✅ see https://github.com/PicnicSupermarket/error-prone-support/pull/1025#discussion_r1485240560 ~~Tests currently are not passing because of the errors in Snippet 3 (import statements are different in `StreamRulesTestOutput` wrt. the ones in `StreamRulesTestInput`). If I add them manually, the tests pass but then the build fails because the imports are unused. Once I format, the tests fail again for the same reason. This never happened to me previously~~ 
2. ✅ thank you @Stephan202! Looks much cleaner now ~~I could not find a way to group the rules inside of a single template/rule.~~
	1. ~~Let's say we want to have the `@BeforeTemplate` in Snippet 1: this does not work, because the result type of the collector is a `Collection`, and therefore only accepts `Collection::isEmpty`~~
	2. ~~Let's say we want to have multiple `@BeforeTemplate` inside the same rule, in order to not repeat the `@AfterTemplate`, as Snippet 2. This does not work, since the two functions have the same signature.~~
	3. ~~Very happy to hear suggestions on how to improve this!~~
3. ✅  irrelevant now ~~I'm pretty sure the name of the rules has to be changed to follow the naming algorithm: I couldn't find what the algorithm is~~

<details>
<summary>Snippet 1</summary>

```java
@BeforeTemplate
boolean before(Stream<T> stream, Collector<? super T, ?, ? extends Collection<?>> collector) {
  return stream.collect(
      collectingAndThen(
          collector,
          Refaster.anyOf(
              Collection::isEmpty,
              ImmutableList::isEmpty,
              ImmutableSet::isEmpty,
              List::isEmpty,
              Set::isEmpty)));
```

</details>

<details>
<summary>Snippet 2</summary>

```java
static final class StreamCollectingAndThenCollectionIsEmpty<T> {
  @BeforeTemplate
  boolean before(Stream<T> stream, Collector<? super T, ?, ? extends Collection<?>> collector) {
    return stream.collect(collectingAndThen(collector, Collection::isEmpty));
  }

  @BeforeTemplate
  boolean before(Stream<T> stream, Collector<? super T, ?, ? extends Set<?>> collector) {
    return stream.collect(collectingAndThen(collector, Set::isEmpty));
  }

  @AfterTemplate
  boolean after(Stream<T> stream) {
    return stream.findAny().isEmpty();
  }
}
```

</details>

<details>
<summary>Snippet 3</summary>

```
[ERROR] Failures:
[ERROR]   RefasterRulesTest.validateRuleCollection:90 value of:
    maybeFormat(...)
diff (-expected +actual):
    @@ -1,11 +1,14 @@
     package tech.picnic.errorprone.refasterrules;

    +import static com.google.common.collect.ImmutableList.toImmutableList;
    +import static com.google.common.collect.ImmutableMap.toImmutableMap;
     import static com.google.common.collect.ImmutableSet.toImmutableSet;
     import static java.util.Comparator.comparingInt;
     import static java.util.Comparator.naturalOrder;
     import static java.util.Comparator.reverseOrder;
     import static java.util.function.Function.identity;
     import static java.util.function.Predicate.not;
    +import static java.util.stream.Collectors.collectingAndThen;
     import static java.util.stream.Collectors.counting;
     import static java.util.stream.Collectors.filtering;
     import static java.util.stream.Collectors.flatMapping;
    @@ -20,16 +23,24 @@
     import static java.util.stream.Collectors.summingDouble;
     import static java.util.stream.Collectors.summingInt;
     import static java.util.stream.Collectors.summingLong;
    +import static java.util.stream.Collectors.toList;
    +import static java.util.stream.Collectors.toMap;
    +import static java.util.stream.Collectors.toSet;

     import com.google.common.collect.ImmutableList;
    +import com.google.common.collect.ImmutableMap;
     import com.google.common.collect.ImmutableSet;
     import com.google.common.collect.Streams;
     import java.util.Arrays;
    +import java.util.Collection;
     import java.util.DoubleSummaryStatistics;
     import java.util.IntSummaryStatistics;
    +import java.util.List;
     import java.util.LongSummaryStatistics;
    +import java.util.Map;
     import java.util.Objects;
     import java.util.Optional;
    +import java.util.Set;
     import java.util.function.Function;
     import java.util.function.Predicate;
     import java.util.stream.Stream;
```

</details>

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
@@ -4,6 +4,7 @@ import static com.google.errorprone.refaster.ImportPolicy.STATIC_IMPORT_ALWAYS;
 import static java.util.Comparator.naturalOrder;
 import static java.util.Comparator.reverseOrder;
 import static java.util.function.Predicate.not;
+import static java.util.stream.Collectors.collectingAndThen;
 import static java.util.stream.Collectors.counting;
 import static java.util.stream.Collectors.filtering;
 import static java.util.stream.Collectors.flatMapping;
@@ -20,6 +21,8 @@ import static java.util.stream.Collectors.summingInt;
 import static java.util.stream.Collectors.summingLong;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Streams;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.google.errorprone.refaster.Refaster;
@@ -36,9 +39,12 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.DoubleSummaryStatistics;
 import java.util.IntSummaryStatistics;
+import java.util.List;
 import java.util.LongSummaryStatistics;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.BinaryOperator;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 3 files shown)


Stephan202:
Tnx @giovannizotta! I'm taking a look :)

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StreamRulesTestInput.java
```diff
@@ -1,10 +1,13 @@
 package tech.picnic.errorprone.refasterrules;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
```
Stephan202, line 4: W.r.t. your question about how to handle those unused imports: for this we have the [`elidedTypesAndStaticImports`](https://github.com/PicnicSupermarket/error-prone-support/blob/57fa6ae2b830be9b8b67f6fb55105c59fe717244/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StreamRulesTestInput.java#L37-L58) method: by referencing the type or member there (in both input and output file) the import won't become unused.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -250,6 +256,99 @@ Optional<S> after(Stream<T> stream, Function<? super T, S> function) {
...
+    boolean before(
+        Stream<T> stream, Collector<? super T, ?, ? extends ImmutableList<?>> collector) {
+      return stream.collect(collectingAndThen(collector, ImmutableList::isEmpty));
+    }
+
+    @AfterTemplate
+    boolean after(Stream<T> stream) {
+      return stream.findAny().isEmpty();
+    }
+  }
```
Stephan202: For each of these cases: the `::isEmpty` method reference may refer to a supertype of the type to which the the values are collected. That makes the number of possible combinations much larger. This relates to your second question; I'll propose something involving additional type parameters.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -250,6 +256,99 @@ Optional<S> after(Stream<T> stream, Function<? super T, S> function) {
...
+  // XXX: This rule assumes that any matched `Collector` does not perform any filtering.
+  static final class StreamCollectingAndThenMapIsEmpty<T> {
+    @BeforeTemplate
+    boolean before(Stream<T> stream, Collector<? super T, ?, ? extends Map<?, ?>> collector) {
+      return stream.collect(collectingAndThen(collector, Map::isEmpty));
+    }
+
+    @AfterTemplate
+    boolean after(Stream<T> stream) {
+      return stream.findAny().isEmpty();
```
Stephan202: W.r.t. your third question: rules are named after their `@AfterTemplate`. For new templates we'd name this rule `StreamFindAnyIsEmpty`. However, there is already a `StreamIsEmpty` template below with exactly this same `@AfterTemplate` definition, so let's try to merge the changes into that one.

Stephan202 (approved):
I added a commit with suggestions.

Suggested commit message:
```
Extend `StreamIsEmpty` Refaster rule (#1025)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
giovannizotta: Indeed, I really didn't like that solution. It looks much better now, thanks for the suggestion! Looks like I've got to do some homework on Generics 😄

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StreamRulesTestInput.java
giovannizotta: Makes sense, thanks for the explanation!

giovannizotta:
Thank you @Stephan202 for the suggestions, looks much better now! I was wondering: would it make sense to add the `@AlsoNegation` annotation to `StreamIsEmpty`? 

The rule `CollectionIsEmpty` does have `@AlsoNegation`

Stephan202:
> I was wondering: would it make sense to add the `@AlsoNegation` annotation to `StreamIsEmpty`?

TBH, I was wondering the same yesterday, but doing that would require a bit more cleanup, as there'd then be overlap with `StreamIsNotEmpty`. But now that you asked :sweat_smile:: I filed #1028 on top of this PR.

rickie (approved):
Nice rules @giovannizotta 🚀 ! This is _just_ in time for the release 😉.

PR #1027, opened 2024-02-10
Extend `CollectionIsEmpty` Refaster rule

giovannizotta:
Add the following expression to `CollectionIsEmpty`: `collection.stream().findAny().isEmpty()`.

### Clashing issue
Currently, this rule clashes with the rule `OptionalFirstCollectionElement`, which is matching `collection.stream().findAny()`, rewriting it to `collection.stream().findFirst()`. Not sure what's the best way to solve this issue, both use cases seem valid 🤔
1. Suppressing the suggestion to rewrite `collection.stream().findAny().isEmpty()` to `collection.stream().findFirst().isEmpty()` leads to a loop since `stream().findFirst().isEmpty()` is rewritten to `stream().findAny().isEmpty()`

### Benefits 
This acts as a useful partner to `StreamIsEmpty`: expressions that match the one introduced in this PR can result as a consequence of applying the `StreamIsEmpty` rule: this allows for refactors like the following to be applied in multiple steps:

```
1. collection.stream().collect(collectingAndThen(toImmutableList(), List::isEmpty))
3. collection.stream().findAny().isEmpty()
4. collection.isEmpty()
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
@@ -35,14 +35,21 @@ final class CollectionRules {
    */
   static final class CollectionIsEmpty<T> {
     @BeforeTemplate
-    @SuppressWarnings("java:S1155" /* This violation will be rewritten. */)
+    @SuppressWarnings({
+      "java:S1155" /* This violation will be rewritten. */,
+      "LexicographicalAnnotationAttributeListing" /* `key-*` entry must remain last. */,
+      "OptionalFirstCollectionElement" /* This is a more specific template. */,
+      "StreamIsEmpty" /* This is a more specific template. */,
+      "key-to-resolve-AnnotationUseStyle-and-TrailingComment-check-conflict"
+    })
     boolean before(Collection<T> collection) {
       return Refaster.anyOf(
           collection.size() == 0,
           collection.size() <= 0,
           collection.size() < 1,
           Iterables.isEmpty(collection),
-          collection.stream().findAny().isEmpty());
+          collection.stream().findAny().isEmpty(),
+          collection.stream().findFirst().isEmpty());
     }
 
     @BeforeTemplate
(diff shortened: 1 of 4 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
```diff
@@ -35,13 +35,21 @@ private CollectionRules() {}
...
+    })
     boolean before(Collection<T> collection) {
       return Refaster.anyOf(
           collection.size() == 0,
           collection.size() <= 0,
           collection.size() < 1,
-          Iterables.isEmpty(collection));
+          Iterables.isEmpty(collection),
+          collection.stream().findAny().isEmpty(),
+          collection.stream().findFirst().isEmpty());
```
Stephan202, line 52: So `OptionalFirstCollectionElement` and `StreamIsEmpty` prefer different variants. But our Refaster integration prefers writing the longest expression matched, so this one will win.

Stephan202 (approved):
Right, so we generally prefer `findFirst()` over `findAny()` (because in the vast majority of cases the former better expresses semantics), but when followed by `isEmpty()`/`isPresent()` we prefer the latter. This is a bit of a quirk in our setup; usually we try to avoid those "contextual preferences", as it can cause a combinatorial explosion of cases to consider.

So a second option would be to change the other rules to unconditionally prefer `findFirst()`. But in this case I think a suppression may be nicer. I rebased and pushed a proposal :)

Suggested commit message:
```
Extend `CollectionIsEmpty` Refaster rule (#1027)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
giovannizotta: Understood, I was wondering indeed what happens when an expression is matched by multiple rules. It makes sense that the most specific one wins 👍

giovannizotta:
Thank you @Stephan202, indeed the need for suppression comes from the fact that we're "inconsistent" over the preference of `findAny()` and `findFirst()`. The suggestions LGTM, thanks a lot!

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
Stephan202: Yeah; implemented [here](https://github.com/PicnicSupermarket/error-prone-support/blob/1cc792c615ff9a3fdb164506255dda7e78069ad8/refaster-runner/src/main/java/tech/picnic/errorprone/refaster/runner/Refaster.java#L100-L131) :)

rickie (approved):
:rocket:

PR #1028, opened 2024-02-11
Compact and replace `StreamIs{,Not}Empty` Refaster rules

Stephan202:
~:exclamation: This PR is on top of #1025. :exclamation:~

Suggested commit message:
```
Compact and replace `StreamIs{,Not}Empty` Refaster rules (#1028)

The new `StreamFindAnyIs{Empty,Present}` rules are simpler thanks to the
use of `@AlsoNegation`. In some cases an additional application of the
`OptionalIsEmpty` rule will be required.
```


giovannizotta (approved):
Nice consolidation, thanks!

rickie (approved):

PR #1031, opened 2024-02-12
javac --release instead of source/target

hboutemy:
it is safer to avoid injecting reference to non-existing APIs
but from CI, it causes some issues in the current advanced setup

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -952,8 +952,7 @@
                             <arg>10000</arg>
                         </compilerArgs>
                         <parameters>true</parameters>
-                        <source>${version.jdk}</source>
-                        <target>${version.jdk}</target>
+                        <release>${version.jdk}</release>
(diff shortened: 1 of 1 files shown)


Stephan202:
Indeed, unfortunately it is not possible to use `--release` in combination with `--add-exports`. As this project directly accesses various JDK-internal APIs, we rely on the latter. We're creating _some_ amount of safety by [building against multiple JDKs](https://github.com/PicnicSupermarket/error-prone-support/blob/363b0c22c7a1c2a7d6f30bf7837bfe36cfbe1870/.github/workflows/build.yml#L13). @hboutemy unless you know of a better solution (would be great! :pray:), are you okay with closing the PR?

hboutemy:
ok, you know what you are doing: advanced use case...
closing, as I sadly don't have any idea on a better workaround

PR #1046, opened 2024-02-19
Have Renovate file `action-download-artifact` PRs once every four weeks

rickie:
Wanted to omit `dawidd6` to fit the PR description.

As suggested here: https://github.com/PicnicSupermarket/error-prone-support/pull/1039#issuecomment-1951878345.

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -13,6 +13,7 @@
     },
     {
       "matchDepNames": [
+        "dawidd6/action-download-artifact",
         "github/codeql-action",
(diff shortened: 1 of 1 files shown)


Stephan202 (approved):
Tnx!

(We could also apply the once-every-four-weeks schedule to all `github-actions` managers, perhaps. Something we can discuss offline, perhaps; no strong opinion.)

mohamedsamehsalah (approved):
🚀

PR #1047, opened 2024-02-19
Introduce `FluxFromIterable` Refaster rule

dirkvbok:
Based on an internal discussion, we found this case for a Refaster rule;

we can omit `.stream()` on a `Collection` and directly use `Flux.fromIterable`.

I wasn't sure about the order of rules and tests, so I placed them at the end, LMK if they need to move. :)

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1207,6 +1207,21 @@ final class ReactorRules {
   }
 
   /**
+   * Don't unnecessarily convert {@link Collection} to {@link Stream} before creating {@link Flux}.
+   */
+  static final class FluxFromIterable<T> {
+    @BeforeTemplate
+    Flux<T> before(Collection<T> c) {
+      return Flux.fromStream(c.stream());
+    }
+
+    @AfterTemplate
+    Flux<T> after(Collection<T> c) {
+      return Flux.fromIterable(c);
+    }
+  }
+
+  /**
    * Prefer {@link Flux#count()} followed by a conversion from {@code long} to {@code int} over
    * collecting into a list and counting its elements.
    */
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1899,4 +1900,19 @@ Duration after(StepVerifier.LastStep step, Duration duration) {
       return step.verifyTimeout(duration);
     }
   }
+
+  /**
+   * Don't unnecessarily convert {@link Collection} to {@link Stream} before creating {@link Flux}.
+   */
+  static final class FluxFromIterable<T> {
```
rickie: I think a bit more up in the file makes sense, now they are below the `StepVerifier` rules. There is no strict pre-determined order yet, so right now I put it under the the "last" `iterable` related rule in the file.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1899,4 +1900,19 @@ Duration after(StepVerifier.LastStep step, Duration duration) {
       return step.verifyTimeout(duration);
     }
   }
+
+  /**
+   * Don't unnecessarily convert {@link Collection} to {@link Stream} before creating {@link Flux}.
```
rickie:
```suggestion
   * Don't unnecessarily convert a {@link Collection} to a {@link Stream} before creating a {@link Flux}.
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1899,4 +1900,19 @@ Duration after(StepVerifier.LastStep step, Duration duration) {
...
+   */
+  static final class FluxFromIterable<T> {
+    @BeforeTemplate
+    Flux<T> before(Collection<T> c) {
```
rickie:
```suggestion
    Flux<T> before(Collection<T> collection) {
```
In most cases, we use the full name `collection` :).

rickie (approved):
Nice addition @dirkvbok :rocket:!!! Added a commit with some minor suggestions :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1205,6 +1206,22 @@ Flux<T> after(Flux<? extends Iterable<T>> flux, int prefetch) {
     }
   }
 
+  /**
+   * Don't unnecessarily convert a {@link Collection} to a {@link Stream} before creating a {@link
+   * Flux}.
+   */
```
Stephan202: I'll propose a slightly more generic message, that would also fit any additional `@BeforeTemplate` cases.

Stephan202 (approved):
Tnx @dirkvbok! Rebased and added a commit. Will let @rickie merge.

Stephan202:
Suggested commit message:
```
Introduce `FluxFromIterable` Refaster rule (#1047)
```

PR #1056, opened 2024-02-26
Exempt `picocli.CommandLine.Option#names` annotation attribute from reordering

rickie:
As part of https://github.com/checkstyle/checkstyle/issues/14137 I'm enabling the `LexicographicalAnnotationAttributeListing` check. I encountered a problem related to this: https://github.com/PicnicSupermarket/error-prone-support/blob/master/integration-tests/checkstyle-init.patch#L78. 

Therefore I'm filing this PR.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationAttributeListing.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/LexicographicalAnnotationAttributeListing.java
@@ -67,14 +67,15 @@ public final class LexicographicalAnnotationAttributeListing extends BugChecker
   private static final long serialVersionUID = 1L;
   private static final ImmutableSet<String> BLACKLISTED_ANNOTATIONS =
       ImmutableSet.of(
-          // XXX: unless JsonPropertyOrder#alphabetic is true...
+          // XXX: Unless `JsonPropertyOrder#alphabetic` is true...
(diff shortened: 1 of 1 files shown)


Stephan202 (commented):
Tnx @rickie! I do seem to recall that this parameter cares about the order in which names are listed, but I forgot the details, and the [documentation](https://javadoc.io/doc/info.picocli/picocli/4.7.5/picocli/CommandLine.Option.html#names--) seems silent about this. What goes wrong without this change?

(Sorry for asking rather than checking myself; rather short on time atm :grimacing:.)

rickie:
As discussed offline; for some CLI's, like the one in Checkstyle, it's just that a specific order is required. The options are expected or sorted in a specific order and therefore we shouldn't want to change this.

Stephan202 (approved):
More explicitly: the order in which flags are listed influences the order in which they are rendered in the generated help text. That is, there's a clear functional impact to the order.

mohamedsamehsalah (approved):
Clean 🥇

rickie:
Suggested commit message:
```
Exempt `picocli.CommandLine.Option#names` annotation attribute from reordering (#1056)
```

rickie:
@Stephan202 we could go for this as alternative:
```
Exempt `@picocli.CommandLine.Option#names` attribute from reordering (#1056)
```
As it will be under 80 chars, WDYT?

Stephan202:
The `@` may not be obvious. Perhaps just:
```
Exempt `picocli.CommandLine.Option#names` arguments from reordering (#1056)
```

PR #1060, opened 2024-03-01
Exclude `error-prone-{experimental,guidelines}` checks from integration tests

rickie:

PR #1064, opened 2024-03-04
Sync Checkstyle integration test

rickie:
❗ These changes are on top of #1060 ❗ 

Suggested commit message:
```
Sync Checkstyle integration test (#1064)

Summary of changes:
- Minimize `checkstyle-init.patch`.
- Test against version 10.14.0 rather than 10.13.0.
```


mohamedsamehsalah (approved):
🧹

integration-tests/checkstyle-init.patch
```diff
@@ -1,7 +1,9 @@
+diff --git a/pom.xml b/pom.xml
+index b6e23f5..4a09c65 100644
```
Stephan202, line 2: Here and below: we should omit these lines.

Stephan202 (commented):
Nice! Will have a closer look later.

Stephan202 (approved):
Rebased and added two more commits. Also tweaked the suggested commit message.

rickie:
Two questions: 

- how do I create a diff without the meta data? Am I doing something wrong? 
- why wasn't that suppression required to have?

Stephan202:
> how do I create a diff without the meta data? Am I doing something wrong?

Nope, you'll have to do this manually with `grep -vP '^(diff|index)'` [like here](https://github.com/PicnicSupermarket/error-prone-support/blob/c214733517b605a9b0e802b433447909edb446e9/integration-tests/checkstyle.sh#L177). 

> why wasn't that suppression required to have?

Because we just merged #1056 :upside_down_face:

rickie:
Ah thanks, now I know it for next time 😉.

> Because we just merged https://github.com/PicnicSupermarket/error-prone-support/pull/1056 🙃

😂 Damn, how stupid of me 😅.

PR #1075, opened 2024-03-11
Introduce Cody as IntelliJ icon

rickie:
As requested by @timtebeek 😉.


.gitignore
```diff
@@ -12,3 +12,6 @@ target
 
 # The Git repositories checked out by the integration test framework.
 integration-tests/.repos
+
+# Specific file exclusions.
```
Stephan202, line 16: So it's an _inclusion_ in the repo ;). Perhaps we can just move this to line 7, so that a comment isn't necessary?

Stephan202 (commented):
Nice idea. I can't find the menu entry to change this feature (perhaps because I didn't switch to the new UI yet), but I wonder whether we:
1. Could/should use a Cody image (perhaps still with a Picnic reference).
2. Use an SVG file ([seems so](https://youtrack.jetbrains.com/issue/IDEA-325249)).

Stephan202 (commented):
Rebased and dropped the `\r` characters from the SVG file. I tried out the new IDEA UI, and indeed saw the logo :+1:

.gitignore
```diff
@@ -3,7 +3,8 @@
 .classpath
 .DS_Store
 .factorypath
-.idea
+**/.idea/*
```
Stephan202, line 6: How does this apply to this PR / under which circumstances may this happen? :)

Would expect that this suffices: 
```suggestion
**/.idea
```

rickie: I did this as I got some changes in the `integration-test` directory of Checkstyle and the `.idea` folder. I made this change and it went away.

rickie: The `/*` part was required as we have the inclusion afterwards.

Stephan202 (approved):
Nice! GitHub claims that Cody has a bunch of empty space at the top and bottom, but IDEA and my local image viewer claim that this isn't so :+1:.

Don't want to block the PR over a `.gitattributes` discussion, but do want to understand :grimacing:

mohamedsamehsalah (approved):

mohamedsamehsalah:
Super cool 🤩 
I would personally love to see Picnic icon added as sugegsted by Stephan ☝️ 

Screenshot 2024-03-25 at 15 45 40

rickie:
> I would personally love to see a Picnic reference on the icon added as sugegsted by Stephan ☝️

Yeah we of course have the :picnic-cody: but I think it's not that visible if we would add that. So I would vote to leave it as-is.

rickie:
Circled back a bit late; but verified that with the recent changes it works :).

rickie:
Suggested commit message:
```
Introduce Picnic logo as IntelliJ icon (#1075)
```

rickie:
@Stephan202 feel free to merge this one :).

PR #1076, opened 2024-03-11
Have `step-security/harden-runner` audit the OpenSSF Scorecard update workflow

Stephan202:
Suggested commit message:
```
Have `step-security/harden-runner` audit the OpenSSF Scorecard update workflow (#1076)

When executed on `master` this workflow requires additional permissions;
let's find out what they are.
```


mohamedsamehsalah (approved):
🎑

rickie (approved):

PR #1104, opened 2024-03-23
Introduce Refaster rules that resolve `EnumOrdinal` violations

Stephan202:
Suggested commit message:
```
Introduce Refaster rules that resolve `EnumOrdinal` violations (#1104)
```

These rules resolve some of the violations reported by the [EnumOrdinal](https://errorprone.info/bugpattern/EnumOrdinal) check new in Error Prone [2.26.0](https://github.com/google/error-prone/releases/tag/v2.26.0).

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ComparatorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ComparatorRules.java
@@ -16,8 +16,11 @@ import com.google.common.collect.ImmutableSet;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.google.errorprone.refaster.Refaster;
 import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.AlsoNegation;
 import com.google.errorprone.refaster.annotation.BeforeTemplate;
 import com.google.errorprone.refaster.annotation.Matches;
+import com.google.errorprone.refaster.annotation.MayOptionallyUse;
+import com.google.errorprone.refaster.annotation.Placeholder;
 import com.google.errorprone.refaster.annotation.Repeated;
 import com.google.errorprone.refaster.annotation.UseImportPolicy;
 import java.util.Arrays;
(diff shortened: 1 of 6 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ComparatorRules.java
```diff
@@ -92,6 +95,24 @@ Comparator<T> after(Comparator<T> cmp) {
...
+    Comparator<T> before() {
+      return comparingInt(v -> toEnumFunction(v).ordinal());
+    }
+
+    @AfterTemplate
+    @UseImportPolicy(STATIC_IMPORT_ALWAYS)
+    Comparator<T> after() {
+      return comparing(v -> toEnumFunction(v));
+    }
+  }
```
Stephan202, line 114: This will rewrite `comparingInt(e -> e.getKey().ordinal())` to `comparing(v -> v.getKey())`. That result will not also be rewritten to `comparingByKey()`, because [MapEntryComparingByKey](https://github.com/PicnicSupermarket/error-prone-support/blob/bb2b1e60349b546cab99bab54bc66e0dbce1a122/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MapEntryRules.java#L47-L60) matches against `Map.Entry::getKey` but not the associated lambda expression. This limitation applies to many of our rules, so fixing it is out of scope. (We should really "just" fix the `MethodReferenceUsage` check.)

I'm mentioning this because it'll prevent a cleanup opportunity in our internal code base, I see.

Stephan202 (commented):
Some context :)

rickie (approved):

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ComparatorRules.java
rickie: Or combine this with OpenRewrite's https://docs.openrewrite.org/recipes/staticanalysis/replacelambdawithmethodreference if we can...

mohamedsamehsalah (approved):
Neat ⭐

PR #1128, opened 2024-04-09
Update `FluxTake` Refaster rule for Reactor 3.5.0+

Stephan202:
Something I noticed while going over the internal code base.

Suggested commit message:
```
Update `FluxTake` Refaster rule for Reactor 3.5.0+ (#1128)
```


werli (approved):
This change assumes users use Reactor 3.5.0+ but I think that's a fair assumption. 👍 

Let's roll with it. 🚀

rickie (approved):

PR #1135, opened 2024-04-13
Introduce `Check{Argument,State}WithMessageAndArguments` Refaster rules

Stephan202:
Suggested commit message:
```
Introduce `Check{Argument,State}WithMessageAndArguments` Refaster rules
```

The tests pass, but when I run this against an internal code base relevant violations aren't replaced. Requires a deeper look.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PreconditionsRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PreconditionsRules.java
@@ -9,8 +9,11 @@ import static com.google.errorprone.refaster.ImportPolicy.STATIC_IMPORT_ALWAYS;
 import static java.util.Objects.requireNonNull;
 
 import com.google.common.base.Preconditions;
+import com.google.errorprone.annotations.FormatMethod;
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/PreconditionsRules.java
```diff
@@ -52,6 +55,26 @@ void after(boolean condition, String message) {
     }
   }
 
+  /**
+   * Prefer {@link Preconditions#checkArgument(boolean, String, Object...)} over more verbose
+   * alternatives.
+   */
+  // XXX: This check assumes that the format string only uses `%s` placeholders.
```
Stephan202, line 62: So this isn't fully correct. One option is to create an BugChecker instead, but that's more work :thinking:

Stephan202: Okay, I went down the rabbit hole: #1139.

Stephan202 (commented):
Closing this PR in favour of #1139.

PR #1138, opened 2024-04-14
Introduce `RedundantStringEscape` check

Stephan202:
Suggested commit message:
```
Introduce `RedundantStringEscape` check (#1138)

This check aims to simplify string constants by dropping redundant
single quote escape sequences. The check is optimized for performance.

While there, update existing checks such that they do not introduce
violations of the type flagged by this new check.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/RedundantStringEscape.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/RedundantStringEscape.java
@@ -1,0 +1,89 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import static com.google.errorprone.BugPattern.LinkType.CUSTOM;
+import static com.google.errorprone.BugPattern.SeverityLevel.SUGGESTION;
+import static com.google.errorprone.BugPattern.StandardTags.SIMPLIFICATION;
+import static tech.picnic.errorprone.utils.Documentation.BUG_PATTERNS_BASE_URL;
+
+import com.google.auto.service.AutoService;
+import com.google.errorprone.BugPattern;
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.LiteralTreeMatcher;
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.LiteralTree;
+import tech.picnic.errorprone.utils.SourceCode;
+
+/** A {@link BugChecker} that flags string constants with extraneous escaping. */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Inside string expressions single quotes do not need to be escaped",
+    link = BUG_PATTERNS_BASE_URL + "RedundantStringEscape",
+    linkType = CUSTOM,
+    severity = SUGGESTION,
+    tags = SIMPLIFICATION)
+public final class RedundantStringEscape extends BugChecker implements LiteralTreeMatcher {
+  private static final long serialVersionUID = 1L;
+
+  /** Instantiates a new {@link RedundantStringEscape} instance. */
+  public RedundantStringEscape() {}
+
+  @Override
+  public Description matchLiteral(LiteralTree tree, VisitorState state) {
+    String constant = ASTHelpers.constValue(tree, String.class);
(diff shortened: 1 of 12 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/RedundantStringEscape.java
```diff
@@ -0,0 +1,89 @@
...
+public final class RedundantStringEscape extends BugChecker implements LiteralTreeMatcher {
+  private static final long serialVersionUID = 1L;
+
+  /** Instantiates a new {@link RedundantStringEscape} instance. */
+  public RedundantStringEscape() {}
+
+  @Override
+  public Description matchLiteral(LiteralTree tree, VisitorState state) {
+    String constant = ASTHelpers.constValue(tree, String.class);
+    if (constant == null || constant.indexOf('\'') == -1) {
```
Stephan202, line 36: Pitest is correct that the `|| constant.indexOf('\'') == -1` condition can be dropped. It's here for performance reasons.

error-prone-utils/src/main/java/tech/picnic/errorprone/utils/SourceCode.java
```diff
@@ -42,6 +43,27 @@ public static String treeToString(Tree tree, VisitorState state) {
...
+    for (int i = 0; i < str.length(); i++) {
+      char c = str.charAt(i);
+      if (c == '\'') {
+        result.append('\'');
+      } else {
+        result.append(Convert.quote(c));
+      }
+    }
+    return result.append('"').toString();
+  }
```
Stephan202, line 65: Just noticed we could also use the upstream [`VisitorState#getConstantExpression`](https://github.com/google/error-prone/blob/04c76af83debe4c507c842fcf22bf8d0ddfd9c58/check_api/src/main/java/com/google/errorprone/VisitorState.java#L777-L801) method, though this variant is more efficient.

Stephan202: I filed google/error-prone#4586; will update this PR.

Stephan202: Rebased and added a commit.

Stephan202:
@werli are you up for being the second reviewer? Otherwise I'll ask @mohamedsamehsalah :smile:

werli:
@Stephan202 Yes, I got pinged by Rick. ETA is EOD. 👍

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/RedundantStringEscape.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.bugpatterns.BugChecker;
+import com.google.errorprone.bugpatterns.BugChecker.LiteralTreeMatcher;
+import com.google.errorprone.fixes.SuggestedFix;
+import com.google.errorprone.matchers.Description;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.LiteralTree;
+import tech.picnic.errorprone.utils.SourceCode;
+
+/** A {@link BugChecker} that flags string constants with extraneous escaping. */
```
werli, line 19: IIUC, this covers all unnecessary escaping cases in "single-line strings", but not text blocks. With text blocks, escalating double quotes may also be unnecessary but only if it's not the last character (because the text block would end with `""""`).

I'm happy not (yet) to cover it, but should we document and make a note of it?

---

Or am I missing something since text blocks are a Java 17 feature and not yet an EPS concern?

werli (approved):
One suggestion to consider, but I don't consider it blocking. Nice job 💪

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/RedundantStringEscape.java
Stephan202: Great suggestion! I'll have a look whether this isn't too much trouble to add, and add an `XXX` comment otherwise :+1:

Stephan202: I started working on this, but there are quite some edge cases and performance aspects to consider. As such I pushed only a commit with a comment to follow up on this topic. Will push a draft branch once this PR is merged.

rickie:
LGTM, let's merge 🚀 ?

PR #1162, opened 2024-04-30
Introduce `FilesCreateTempFileToFile` Refaster rule

Stephan202:
Suggested commit message:
```
Introduce `FilesCreateTempFileToFile` Refaster rule (#1162)
```

Inspired by apache/maven-install-plugin#47, found through #1159.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
@@ -2,12 +2,15 @@ package tech.picnic.errorprone.refasterrules;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
+import com.google.errorprone.refaster.Refaster;
 import com.google.errorprone.refaster.annotation.AfterTemplate;
 import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import java.io.File;
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.attribute.FileAttribute;
 import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
 
 /** Refaster rules related to expressions dealing with files. */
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
```diff
@@ -40,4 +43,22 @@ String after(Path path) throws IOException {
...
+    @BeforeTemplate
+    @SuppressWarnings("java:S5443" /* This violation will be rewritten. */)
+    File before(String prefix, String suffix) throws IOException {
+      return Refaster.anyOf(
+          File.createTempFile(prefix, suffix), File.createTempFile(prefix, suffix, null));
+    }
+
+    @AfterTemplate
+    File after(String prefix, String suffix) throws IOException {
+      return Files.createTempFile(prefix, suffix).toFile();
```
Stephan202, line 61: SonarCloud now compains about this line, while the [documentation](https://rules.sonarsource.com/java/type/Security Hotspot/RSPEC-5443) literally says:
```java
File f = Files.createTempFile("prefix", "suffix").toFile();  // Compliant
```

I suppose this is something to be reported upstream (either it's incorrect, or unclear).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
```diff
@@ -40,4 +43,22 @@ String after(Path path) throws IOException {
       return Files.readString(path);
     }
   }
+
+  /**
+   * Prefer {@link Files#createTempFile(String, String, FileAttribute[])} over alternatives that
+   * create files will more liberal permissions.
```
mohamedsamehsalah, line 49:
```suggestion
   * create files with more liberal permissions.
```

mohamedsamehsalah (approved):
Clean 🚀

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
Stephan202: Alright, I had a look, and for POSIX file systems at least this is safe: [(a)](https://github.com/openjdk/jdk/blob/367e0a65561f95aad61b40930d5f46843fee3444/src/java.base/share/classes/java/nio/file/TempFileHelper.java#L91-L117), [(b)](https://github.com/openjdk/jdk/blob/367e0a65561f95aad61b40930d5f46843fee3444/src/java.base/share/classes/java/nio/file/TempFileHelper.java#L66-L67). I tested this on my Linux laptop: the suggested code produces a file with `-rw-------` permissions, while the other two variants create a file with `-rw-rw-r--` permissions.

The question is then what happens on Windows, but regardless this change is an improvement. I've added a comment to this effect, and suggest that this PR is good to go.

rickie (approved):

timtebeek:
> Inspired by [apache/maven-install-plugin#47](https://github.com/apache/maven-install-plugin/pull/47), found through #1159.

Fun to see that link here; Leitschuh has been using Moderne to roll that out broadly indeed. I did notice that [we support both a 2 and 3 arg variant](https://github.com/openrewrite/rewrite-java-security/blob/cb86c82b2f6af0fd3020d24ad55cd438a11b5249/src/main/java/org/openrewrite/java/security/SecureTempFileCreation.java#L102-L119), whereas this only support the two arg option without additional directory argument. Could be worth adding that 3 arg variant here too, such that we can drop our recipe and just use this one. ;)

Stephan202:
Good one! I think I can open a PR for that shortly.

Stephan202:
Filed #1282 :)

PR #1214, opened 2024-06-12
Introduce MongoDB Refaster rules

werli:
Opening this PR as draft to gather feedback on:
1. Do we want MongoDB-related Refaster rules in the first place?
2. This specific rule - I'll open a thread in the code for both.

### DIFF
--- a/error-prone-contrib/pom.xml
+++ b/error-prone-contrib/pom.xml
@@ -217,7 +217,7 @@
         <dependency>
             <groupId>org.mongodb</groupId>
             <artifactId>mongodb-driver-core</artifactId>
-            <scope>test</scope>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.openrewrite</groupId>
--- a/error-prone-contrib/pom.xml
+++ b/error-prone-contrib/pom.xml
@@ -216,6 +216,11 @@
         </dependency>
         <dependency>
             <groupId>org.mongodb</groupId>
+            <artifactId>bson</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mongodb</groupId>
             <artifactId>mongodb-driver-core</artifactId>
             <scope>provided</scope>
         </dependency>
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MongoDBRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MongoDBRules.java
@@ -1,0 +1,31 @@
+package tech.picnic.errorprone.refasterrules;
+
+import static com.google.errorprone.refaster.ImportPolicy.STATIC_IMPORT_ALWAYS;
+import static com.mongodb.client.model.Filters.eq;
+
+import com.google.errorprone.refaster.annotation.AfterTemplate;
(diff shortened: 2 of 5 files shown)


error-prone-contrib/pom.xml
```diff
@@ -177,7 +177,7 @@
         <dependency>
             <groupId>org.mongodb</groupId>
             <artifactId>mongodb-driver-core</artifactId>
-            <scope>test</scope>
+            <scope>provided</scope>
```
werli, line 180: (Unrelated to this line)

Do we want MongoDB-related Refaster rules in the first place?

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MongoDBRules.java
```diff
@@ -0,0 +1,31 @@
...
+    Bson before(String field, Enum<?> value) {
+      return eq(field, value.toString());
+    }
+
+    @AfterTemplate
+    @UseImportPolicy(STATIC_IMPORT_ALWAYS)
+    Bson after(String field, Enum<?> value) {
+      return eq(field, value);
+    }
+  }
```
werli, line 30: This rule works if `Enum#toString` and the encoded value are equivalent.

This may not always be the case. Is there a way in Error Prone to identify e.g. whether `Enum#toString` is overwritten?

If we cannot guarantee it, is a `BugChecker` the better choice?

rickie: We might be able to use a Matcher, we have a few examples here: https://github.com/PicnicSupermarket/error-prone-support/tree/master/refaster-support/src/main/java/tech/picnic/errorprone/refaster/matchers. 

I think we can infer something like you are describing in a Matcher 💪🏻 !

rickie (commented):
Bit slow, but here is a response :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MongoDBRules.java
Stephan202: I wonder: doesn't the (non)equivalence of these variants also depend on how the `CodecRegistry` of the MongoDB driver is configured? For example, if serialization is delegated to Jackson, then a `@JsonValue`-annotated method may influence the result. If I got this right, then checking whether `#toString()` is overridden won't suffice. I suggest to:
1. Test the above hypothesis.
2. If validated, implement a `EnumWithoutCustomStringRepresentation` matcher that validates `#toString()` and `@JsonValue` are absent.
3. Use that matcher for this rule.

The name of the proposed matcher is rather specific, and it's TBD that the `toString()` override can be determined accurately. It looks like for `@JsonValue` we'd also have to consider meta-annotations. There's also the fact that not all users use Jackson, but if some have to disable the rule, that's likely okay.

(For completeness I'll mention that we also have a [`RedundantStringConversion`](https://github.com/PicnicSupermarket/error-prone-support/blob/bc86b7669670875b9a82ff6440b09093e6f9bbe2/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/RedundantStringConversion.java) check, but given the `CodecRegistry` involved here, I don't think extending that check is better than the proposed Refaster rule.)

PR #1239, opened 2024-07-08
Introduce `EnumReferenceEqualityLambda` Refaster rule

mohamedsamehsalah:
Suggested commit message:
```
Introduce `EnumReferenceEqualityLambda` Refaster rule (#1239)
```

### DIFF
--- a/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/EqualityRulesTestInput.java
+++ b/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/EqualityRulesTestInput.java
@@ -1,5 +1,6 @@
 package tech.picnic.errorprone.refasterrules;
 
+import static java.util.function.Predicate.isEqual;
 import static java.util.function.Predicate.not;
 
 import com.google.common.collect.BoundType;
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/EqualityRulesTestInput.java
```diff
@@ -33,6 +34,12 @@ boolean testEqualsPredicate() {
     return Stream.of("foo").anyMatch(s -> "bar".equals(s));
   }
 
+  ImmutableSet<Boolean> testEnumsReferenceEqualityLambda() {
```
Stephan202: This one should be moved up a bit.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/EqualityRulesTestInput.java
```diff
@@ -33,6 +34,12 @@ boolean testEqualsPredicate() {
...
+    return ImmutableSet.of(
+        Stream.of(RoundingMode.UP).anyMatch(isEqual(RoundingMode.DOWN)),
+        Stream.of(RoundingMode.UP).anyMatch(RoundingMode.DOWN::equals));
```
Stephan202: We don't need the `Stream` expressions for this test.

Stephan202 (approved):
Rebased and added a commit; also updated the suggested commit message and PR title. 

(Not sure the `::equals` rewrite is always an improvement. Ideally we benchmark this; should get back to finishing the benchmark framework I started working on :grimacing:.)

rickie (approved):

PR #1244, opened 2024-07-12
Introduce `MonoFromFutureSupplier{,Boolean}` Refaster rules

Venorcis:
Futures subscribe eagerly. This means that `Mono#fromFuture(CompletableFuture)` already executes the contents of the future on assembly time (even when that `Future` is constructed from a `Mono` as we do in our reactive Caffeine caches).

I therefore suggest instead to always use `Mono#fromFuture(Supplier)`, which will only instantiate (and thus execute) the future on subscription, i.e. much closer to how we usually do everything 😄

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1906,8 +1906,13 @@ final class ReactorRules {
     }
   }
 
-  /** Prefer {@link Mono#fromFuture(Supplier)} over {@link Mono#fromFuture(CompletableFuture)}. */
-  static final class MonoFromFuture<T> {
+  /**
+   * Prefer {@link Mono#fromFuture(Supplier)} over {@link Mono#fromFuture(CompletableFuture)}, as
+   * the former may defer initiation of the asynchornous computation until subscription.
+   */
+  static final class MonoFromFutureSupplier<T> {
+    // XXX: Constrain the `future` parameter using `@NotMatches(IsIdentityOperation.class)` once
+    // `IsIdentityOperation` no longer matches nullary method invocations.
     @BeforeTemplate
     Mono<T> before(CompletableFuture<T> future) {
       return Mono.fromFuture(future);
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1904,4 +1905,33 @@ Duration after(StepVerifier.LastStep step, Duration duration) {
...
+    Mono<T> after(CompletableFuture<T> future) {
+      return Mono.fromFuture(() -> future);
+    }
+  }
+
+  /**
+   * Prefer {@link Mono#fromFuture(Supplier, boolean)} over {@link
+   * Mono#fromFuture(CompletableFuture, boolean)}.
+   */
+  static final class MonoFromFutureSuppressCancel<T> {
```
rickie, line 1926: @Stephan202 according to our undocumented naming scheme, we would now call this `MonoFromFutureBoolean` right?

rickie (approved):
Looks good! Nice one @Venorcis :).

Suggested commit message:
```
Introduce `MonoFromFuture{,SuppressCancel}` Refaster rule (#1244)
```
(Commit message pending the rename comment ofcourse)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: Strictly speaking even `MonoFromFutureSupplierBoolean` :smile:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1904,4 +1905,33 @@ Duration after(StepVerifier.LastStep step, Duration duration) {
       return step.verifyTimeout(duration);
     }
   }
+
+  /** Prefer {@link Mono#fromFuture(Supplier)} over {@link Mono#fromFuture(CompletableFuture)}. */
+  static final class MonoFromFuture<T> {
+    @BeforeTemplate
+    Mono<T> before(CompletableFuture<T> future) {
+      return Mono.fromFuture(future);
+    }
```
Stephan202, line 1919: I'll add an `XXX` to later apply `@NotMatches(IsIdentityOperation.class)`. Cause as-is this code will also rewrite `Mono.fromFuture(someVariable)`, which isn't great.

Stephan202 (approved):
Rebased and added a commit. Updated suggested commit message:
```
Introduce `MonoFromFutureSupplier{,Boolean}` Refaster rules (#1244)
```

PR #1260, opened 2024-07-21
Introduce `MonoDeferContextualFromSupplier` refaster rule

mohamedsamehsalah:


mohamedsamehsalah:
That won't work since `ContextView#get` throws an exception rather than resolve to `null`.

Stephan202:
I suppose `Mono.deferContextual(contextView -> Mono.justOrEmpty(contextView.getOrEmpty(key)))` could work, but it's not clearly better. So fine not to have a rule for this :+1:

PR #1264, opened 2024-08-01
Introduce `SetStream` Refaster rule

mohamedsamehsalah:
I think this is a candidate for "combinations of operators that don't make sense".

I tried to extend the rule with rules like 
```
  static final class SetStreamFilter<T> {
    @BeforeTemplate
    Stream<?> before(Set<T> set, Predicate<T> predicate) {
      return set.stream().filter(predicate).distinct();
    }

    @AfterTemplate
    Stream<?> after(Set<T> set, Predicate<T> predicate) {
      return set.stream().filter(predicate);
    }
  }
```

But the number of permutations of different operators before using `#distinct` is a lot.

Suggested commit message
```
Rely on Set's property to preserve distinct elements when streaming
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
@@ -184,7 +184,12 @@ final class CollectionRules {
     }
   }
 
-  /** Rely on Set's property to preserve distinct elements. */
+  /** Don't unnecessarily call {@link Stream#distinct()} on an already-unique stream of elements. */
+  // XXX: This rule assumes that the `Set` relies on `Object#equals`, rather than a custom
+  // equivalence relation.
+  // XXX: Expressions that drop or reorder elements from the stream, such as `.filter`, `.skip` and
+  // `sorted`, can similarly be simplified. Covering all cases is better done using an Error Prone
+  // check.
   static final class SetStream<T> {
     @BeforeTemplate
     Stream<?> before(Set<T> set) {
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
```diff
@@ -184,6 +184,19 @@ void after(Set<T> removeFrom, Collection<S> elementsToRemove) {
     }
   }
 
+  /** Rely on Set's property to preserve distinct elements. */
```
Stephan202: This does assume that the set doesn't use a custom equality, like a `SortedSet` might. Will add a note:
```suggestion
  /** Don't unnecessarily call {@link Stream#distinct()} on an already-unique stream of elements. */
  // XXX: This rule assumes that the `Set` relies on `Object#equals`, rather than a custom
  // equivalence relation.
  // XXX: Expressions that drop or reorder elements from the stream, such as `.filter`, `.skip` and
  // `sorted`, can similarly be simplified. Covering all cases is better done using an Error Prone
  // check.
```

(I'm not too worried about false positives: it's likely rare that there is code that wants cares about both types of deduplication in a single context.)

Stephan202 (approved):
Added a commit.

Suggested commit message:
```
Introduce `SetStream` Refaster rule (#1264)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
```diff
@@ -184,7 +184,12 @@ void after(Set<T> removeFrom, Collection<S> elementsToRemove) {
     }
   }
 
-  /** Rely on Set's property to preserve distinct elements. */
+  /** Don't unnecessarily call {@link Stream#distinct()} on an already-unique stream of elements. */
+  // XXX: This rule assumes that the `Set` relies on `Object#equals`, rather than a custom
```
mohamedsamehsalah, line 188: 🧠

rickie (approved):

mohamedsamehsalah:
> @hpnog Any additions from your side :)?

Discussed offline: We can proceed without @hpnog 's input.

PR #1269, opened 2024-08-04
Introduce `Class{Literal,Reference}Cast` Refaster rules

mohamedsamehsalah:
* Perhaps opinionated, but personally, I find the `Class.class:cast` alternative more readable, especially when working with _Streams_.
* One downside to this rule, is that we lose the compile-time cast checks in exchange for runtime checks.

What are your guys thoughts?

Suggested commit message
```
Introduce `Class(Reference|Literal)Cast` class rules
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ClassRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ClassRules.java
@@ -38,7 +38,12 @@ final class ClassRules {
     }
   }
 
-  /** Prefer {@link Class#isInstance(Object)} method references over more verbose alternatives. */
+  /**
+   * Prefer {@link Class#isInstance(Object)} method references over lambda expressions that require
+   * naming a variable.
+   */
+  // XXX: Once the `ClassReferenceIsInstancePredicate` rule is dropped, rename this rule to just
+  // `ClassIsInstancePredicate`.
   static final class ClassLiteralIsInstancePredicate<T, S> {
     @BeforeTemplate
     Predicate<S> before() {
@@ -51,7 +56,11 @@ final class ClassRules {
     }
   }
 
-  /** Prefer {@link Class#isInstance(Object)} method references over more verbose alternatives. */
+  /**
+   * Prefer {@link Class#isInstance(Object)} method references over lambda expressions that require
+   * naming a variable.
+   */
+  // XXX: Drop this rule once the `MethodReferenceUsage` rule is enabled by default.
   static final class ClassReferenceIsInstancePredicate<T, S> {
     @BeforeTemplate
     Predicate<S> before(Class<T> clazz) {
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ClassRulesTestInput.java
```diff
@@ -23,4 +24,13 @@ Predicate<String> testClassReferenceIsInstancePredicate() throws IOException {
     Class<?> clazz = CharSequence.class;
     return s -> clazz.isInstance(s);
   }
+
+  Function<Number, Integer> testClassLiteralCast() {
```
mohamedsamehsalah, line 28: Curious why the pevious three tests threw exceptions as part of their signature 🤔

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ClassRules.java
```diff
@@ -62,4 +63,31 @@ Predicate<S> after(Class<T> clazz) {
       return clazz::isInstance;
     }
   }
+
+  /** Prefer {@link Class#cast(Object)} method references over more verbose alternatives. */
```
Stephan202: "Verbose" is perhaps a bit strong/unclear. I see that this mirrors the rules above, though. Will propose something.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ClassRulesTestInput.java
Stephan202: Looks like a copy-paste error; will drop!

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ClassRulesTestInput.java
```diff
@@ -23,4 +24,13 @@ Predicate<String> testClassReferenceIsInstancePredicate() throws IOException {
...
+    return i -> (Integer) i;
+  }
+
+  Function<Number, Integer> testClassReferenceCast() {
+    Class<? extends Integer> clazz = Integer.class;
+    return i -> clazz.cast(i);
```
Stephan202: We don't need the separate variable here. (It's used above to test for a _non_-matching scenario.)

Stephan202 (approved):
I don't think that this impacts type safety, and it matches the Error Prone Support code style.

```
Introduce `Class{Literal,Reference}Cast` Refaster rules (#1269)
```

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ClassRulesTestInput.java
mohamedsamehsalah: Ah! Makes sense 👍

rickie (approved):

PR #1270, opened 2024-08-04
Package OpenRewrite recipes in separate JAR, targeting Java 8

Stephan202:
See openrewrite/rewrite-third-party#11 for context.

~Opening this PR as draft, as it introduces an unacceptable self-check regression: it appears that the [Refaster](https://github.com/PicnicSupermarket/error-prone-support/blob/290ddf19722fef7281da0437abc4fa2728ec7536/refaster-runner/src/main/java/tech/picnic/errorprone/refaster/runner/Refaster.java) bug pattern no longer flags violations during our self check. I validated that it _is_ executed, so the issue appears to be somewhere inside Refaster's (upstream, out-of-the-box) matching logic. Requires further investigation. We rely quite a lot on this feature, so IMHO the PR can't be merged without resolving this limitation.~ Went for a different solution: instead of attempting to produce Java 8 class files in the same location where previously we had Java 17 files, the recipes are now packaged into a separate JAR, with classifier `recipes`.

Suggested commit message:
```
Package OpenRewrite recipes in separate JAR, targeting Java 8 (#1270)

This is a prerequisite for openrewrite/rewrite-third-party#11. The new JAR has
classifier `recipes`.
```

CC @timtebeek.


pom.xml
```diff
@@ -965,6 +958,79 @@
...
+                                    <arg>--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED</arg>
+                                    <arg>--add-exports=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED</arg>
+                                    <arg>--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED</arg>
+                                    <arg>--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED</arg>
+                                    <arg>--add-exports=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED</arg>
+                                    <arg>--add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED</arg>
+                                    <arg>--add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED</arg>
+                                </compilerArgs>
+                                <excludes>
+                                    <exclude>**/*Rules.java</exclude>
```
Stephan202, line 999: One might think "why not customize the build only for `*Recipes.java` files?" This appears not to work, because at the start of the compilation those files don't yet exist; this a `maven-compiler-plugin` feature; not a `javac` feature. Open to suggestions, of course!

Stephan202 (commented):
Before:
```
$ find error-prone-contrib -name "*.class" | xargs file | grep -oP 'version .*?$' | sort | uniq -c
   1857 version 61.0
```

After:
```
$ find error-prone-contrib -name "*.class" | xargs file | grep -oP 'version .*?$' | sort | uniq -c
   1766 version 52.0 (Java 1.8)
     91 version 61.0
```

pom.xml
timtebeek: Great to see this start! In my earlier exploration I had wondered if we could use `<proc>only</proc>` to only generate the Java 8 compatible sources first (using Java 17/21), optionally to a different output, and then in a second pass compile those generated sources using Java 8 as target.

Stephan202: Interesting idea, to use `<proc>only</proc>`! I just tried some variations of this, but unfortunately that seems to interact badly (different kinds of errors) with subsequent steps (using either `<proc>node</proc>` or `<proc>full</proc>`).

I then tried the "different output" route, where a JAR with a custom classifier (`recipes`) is created: `error-prone-contrib-0.17.1-SNAPSHOT-recipes.jar`. I push that change; would this work for OpenRewrite?

timtebeek: A separate jar would work perfectly for us; Might help make for a lighter distribution for EPS itself as well. Thanks a lot!

Stephan202:
It's bed time, but this is surprising:
```
$ find error-prone-contrib -name "*.class" | xargs file | grep -oP 'version .*?$' | sort | uniq -c
    787 version 52.0 (Java 1.8)
   1857 version 61.0
$ unzip -qd /tmp/eps-recipes error-prone-contrib/target/error-prone-contrib-0.17.1-SNAPSHOT-recipes.jar 
$ unzip -qd /tmp/eps-non-recipes error-prone-contrib/target/error-prone-contrib-0.17.1-SNAPSHOT.jar
$ find /tmp/eps-*recipes -name "*.class" | xargs file | grep -oP 'version .*?$' | sort | uniq -c
   1816 version 61.0
$ find /tmp/eps-recipes -name "*Recipe*.class" | wc -l
787
$ find /tmp/eps-non-recipes -name "*Recipe*.class" | wc -l
0
```

Need to understand how after packaging the `*Recipe*` classes suddenly contain Java 17 bytecode again :exploding_head:

Stephan202:
Indeed, running `diff -r error-prone-contrib/target/openrewrite-recipes/tech/picnic/errorprone/refasterrules/ /tmp/eps-recipes/tech/picnic/errorprone/refasterrules/` reports that all files differ. This would point to the `maven-jar-plugin` doing "something", but that seems just... weird. (Completely unsurprisingly,) setting `<detectMultiReleaseJar>false</detectMultiReleaseJar>` doesn't fix it. I don't see any other remotely-related setting to try :thinking:. This may require proper debugging.

Stephan202:
Ah, the `*Recipe*` files are compiled twice of course, and I'm not instructing the `maven-jar-plugin` to get them from `<outputDirectory>${project.build.directory}/openrewrite-recipes</outputDirectory>`. :facepalm: Pushed a fix :smile:

```
 find error-prone-contrib -name "*.class" | xargs file | grep -oP 'version .*?$' | sort | uniq -c
    787 version 52.0 (Java 1.8)
   1857 version 61.0
$ unzip -qd /tmp/eps-recipes error-prone-contrib/target/error-prone-contrib-0.17.1-SNAPSHOT-recipes.jar 
$ unzip -qd /tmp/eps-non-recipes error-prone-contrib/target/error-prone-contrib-0.17.1-SNAPSHOT.jar
$ find /tmp/eps-*recipes -name "*.class" | xargs file | grep -oP 'version .*?$' | sort | uniq -c
    787 version 52.0 (Java 1.8)
   1029 version 61.0
```

Works :muscle:

timtebeek (approved):
🎉

pom.xml
```diff
@@ -965,6 +965,38 @@
                         https://issues.apache.org/jira/browse/MCOMPILER-209. -->
                         <useIncrementalCompilation>false</useIncrementalCompilation>
                     </configuration>
+                    <executions>
+                        <!-- OpenRewrite recipe sources, generated by the
+                        `rewrite-templating` annotation processor and
+                        identified as classes whose name ends in `Recipes`, are
+                        compiled to target Java 8 for compatiblity with the
```
rickie, line 972:
```suggestion
                        compiled to target Java 8 for compatibility with the
```

pom.xml
```diff
@@ -1112,6 +1144,39 @@
...
+                                <excludes>
+                                    <excludes>**/*Recipe$*.class</excludes>
+                                    <excludes>**/*Recipe.class</excludes>
+                                    <excludes>**/*Recipes.class</excludes>
+                                </excludes>
+                            </configuration>
+                        </execution>
+                        <execution>
+                            <!-- Creates a custom JAR with Java 8-compatible
+                            OpenRewrite recipe classess. -->
```
rickie, line 1165:
```suggestion
                            OpenRewrite recipe classes. -->
```

rickie (approved):
Very nice improvements @Stephan202 🚀 !

Found two small typos :).

mohamedsamehsalah (approved):

PR #1271, opened 2024-08-05
Update `step-security/harden-runner` configuration

Stephan202:
Suggested commit message:
```
Update `step-security/harden-runner` configuration (#1271)

While apparently the build doesn't fail without this, it is reasonable
for SonarCloud analysis to access the two additional domains.

While there, introduce subdomain wildcards for `sigstore.dev` and
`sonarcloud.io`.
```

Example runs that attempted to access these domains:
- [To `analysis-sensorcache-eu-central-1-prod.s3.amazonaws.com:443`](https://app.stepsecurity.io/github/PicnicSupermarket/error-prone-support/actions/runs/10242419439?jobid=28332113753&tab=network-events)
- [To `api.nuget.org:443`](https://app.stepsecurity.io/github/PicnicSupermarket/error-prone-support/actions/runs/10193074444?jobid=28197045945&tab=network-events)

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -24,7 +24,9 @@ jobs:
           disable-sudo: true
           egress-policy: block
           allowed-endpoints: >
+            analysis-sensorcache-eu-central-1-prod.s3.amazonaws.com:443
             api.adoptium.net:443
+            api.nuget.org:443
             api.sonarcloud.io:443
             ea6ne4j2sb.execute-api.eu-central-1.amazonaws.com:443
             github.com:443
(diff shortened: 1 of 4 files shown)


Stephan202:
Rebased and added two commits: one that introduces wildcards in all "relevant" places, and a second that reverts half of the changes. For some domains one could argue that using a wildcard is too liberal. I think that the current proposal strikes a nice balance between security and maintainability.

.github/workflows/sonarcloud.yml
```diff
@@ -24,14 +24,15 @@ jobs:
...
             api.adoptium.net:443
-            api.sonarcloud.io:443
+            api.nuget.org:443
             ea6ne4j2sb.execute-api.eu-central-1.amazonaws.com:443
             github.com:443
             objects.githubusercontent.com:443
             repo.maven.apache.org:443
             sc-cleancode-sensorcache-eu-central-1-prod.s3.amazonaws.com:443
-            scanner.sonarcloud.io:443
+            *.sonarcloud.io:443
```
rickie, line 35: `*sonarcloud.io:443` would also work if we want to 🤔

PR #1275, opened 2024-08-10
Introduce `ArraysAsList` Refaster rule

mohamedsamehsalah:
Suggested commit message
```
Introduce `ArraysAsList` Refaster rule (#1275)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
@@ -38,7 +38,6 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.DoubleSummaryStatistics;
 import java.util.IntSummaryStatistics;
-import java.util.List;
 import java.util.LongSummaryStatistics;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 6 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -122,6 +123,21 @@ Stream<T> after(T[] array) {
     }
   }
 
+  /**
+   * Prefer {@link Arrays#asList(Object[])} over {@link Stream#toList()} as the former is clearer.
+   */
+  static final class ArraysAsList<T> {
+    @BeforeTemplate
+    List<T> before(@NotMatches(IsRefasterAsVarargs.class) T[] array) {
+      return Arrays.stream(array).toList();
```
Stephan202: Let's add an `XXX` to track that we're losing unmodifiability.

(We could also target `ImmutableList#copyOf`. For now I'll just make a note about this.)

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StreamRulesTestInput.java
```diff
@@ -86,6 +87,10 @@ Stream<String> testStreamOfArray() {
     return Stream.of(new String[] {"foo", "bar"});
   }
 
+  List<String> testArraysAsList() {
+    return Arrays.stream(new String[] {"foo", "bar"}).toList();
```
Stephan202: Smaller :smile:
```suggestion
    return Arrays.stream(new String[0]).toList();
```

Stephan202 (approved):
Added a commit and tweaked the suggested commit message. Tnx @mohamedsamehsalah!

rickie (approved):

PR #1317, opened 2024-08-31
Introduce `RefasterRuleTestExtractor` for documentation generation

Stephan202:
Suggested commit message:
```
Introduce `RefasterRuleTestExtractor` for documentation generation (#1317)

This new `Extractor` implementation collects Refaster example input and 
output code from rule collection tests.

This change also introduces explicit compilation steps for the test
code. As a side-effect this produces faster feedback in case of invalid
input or output code.
```

After this PR, approximately one more PR remains to port the remainder of the logic on the `website` branch to `master`. After that we can configure the build such that any change on `master` is nearly immediately reflected on https://error-prone.picnic.tech.

(The build will fail because of [this issue](https://github.com/PicnicSupermarket/error-prone-support/pull/1069#issuecomment-2322956932); will try to address that separately.)

### DIFF
--- a/documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
+++ b/documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
@@ -27,7 +27,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import org.jspecify.annotations.Nullable;
-import tech.picnic.errorprone.documentation.BugPatternTestExtractor.TestCases;
+import tech.picnic.errorprone.documentation.BugPatternTestExtractor.BugPatternTestCases;
 
 /**
  * An {@link Extractor} that describes how to extract data from classes that test a {@code
@@ -40,7 +40,7 @@ import tech.picnic.errorprone.documentation.BugPatternTestExtractor.TestCases;
 @Immutable
 @AutoService(Extractor.class)
 @SuppressWarnings("rawtypes" /* See https://github.com/google/auto/issues/870. */)
-public final class BugPatternTestExtractor implements Extractor<TestCases> {
+public final class BugPatternTestExtractor implements Extractor<BugPatternTestCases> {
   /** Instantiates a new {@link BugPatternTestExtractor} instance. */
   public BugPatternTestExtractor() {}
 
@@ -50,7 +50,7 @@ public final class BugPatternTestExtractor implements Extractor<TestCases> {
   }
 
   @Override
-  public Optional<TestCases> tryExtract(ClassTree tree, VisitorState state) {
+  public Optional<BugPatternTestCases> tryExtract(ClassTree tree, VisitorState state) {
     BugPatternTestCollector collector = new BugPatternTestCollector();
 
     collector.scan(tree, state);
@@ -59,7 +59,7 @@ public final class BugPatternTestExtractor implements Extractor<TestCases> {
         .filter(not(ImmutableList::isEmpty))
         .map(
             tests ->
-                new AutoValue_BugPatternTestExtractor_TestCases(
+                new AutoValue_BugPatternTestExtractor_BugPatternTestCases(
                     state.getPath().getCompilationUnit().getSourceFile().toUri(),
                     ASTHelpers.getSymbol(tree).className(),
                     tests));
@@ -95,10 +95,10 @@ public final class BugPatternTestExtractor implements Extractor<TestCases> {
             .onDescendantOf("com.google.errorprone.BugCheckerRefactoringTestHelper.ExpectOutput")
             .namedAnyOf("addOutputLines", "expectUnchanged");
 
-    private final List<TestCase> collectedTestCases = new ArrayList<>();
+    private final List<BugPatternTestCase> collectedBugPatternTestCases = new ArrayList<>();
 
-    private ImmutableList<TestCase> getCollectedTests() {
-      return ImmutableList.copyOf(collectedTestCases);
+    private ImmutableList<BugPatternTestCase> getCollectedTests() {
+      return ImmutableList.copyOf(collectedBugPatternTestCases);
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


documentation-support/src/main/java/tech/picnic/errorprone/documentation/BugPatternTestExtractor.java
```diff
@@ -40,7 +40,7 @@
 @Immutable
 @AutoService(Extractor.class)
 @SuppressWarnings("rawtypes" /* See https://github.com/google/auto/issues/870. */)
-public final class BugPatternTestExtractor implements Extractor<TestCases> {
+public final class BugPatternTestExtractor implements Extractor<BugPatternTestCases> {
```
Stephan202, line 43: Renaming for disambiguation.

documentation-support/src/main/java/tech/picnic/errorprone/documentation/RefasterRuleCollectionTestExtractor.java
```diff
@@ -0,0 +1,176 @@
...
+   *
+   * @implNote This operation attempts to trim leading whitespace, such that the start and end of
+   *     the method declaration are aligned. The implemented heuristic assumes that the code is
+   *     formatted using Google Java Format.
+   */
+  // XXX: Leading Javadoc and other comments are currently not extracted. Consider fixing this.
+  private static String getFormattedSource(MethodTree method, VisitorState state) {
+    String source = SourceCode.treeToString(method, state);
+    int finalNewline = source.lastIndexOf(LINE_SEPARATOR);
+    if (finalNewline < 0) {
```
Stephan202, line 121: Pitest flags that this case isn't covered. There is actually an empty method test, but I suppose that source code does have a trailing newline (didn't check).

documentation-support/src/main/java/tech/picnic/errorprone/documentation/RefasterRuleCollectionTestExtractor.java
```diff
@@ -0,0 +1,176 @@
...
+      return source;
+    }
+
+    int indentation = source.substring(finalNewline).lastIndexOf(' ');
+    String prefixToStrip = " ".repeat(indentation);
```
Stephan202, line 126: Pitest flags that the `.repeat(indentation)` logic isn't covered. This is because the test code is properly formatted; not an issue I suppose. (I did manually check that unformatted code doesn't break the extractor.)

documentation-support/src/main/java/tech/picnic/errorprone/documentation/RefasterRuleCollectionTestExtractor.java
```diff
@@ -0,0 +1,176 @@
...
+    abstract boolean isInput();
+
+    abstract ImmutableList<RefasterTestCase> testCases();
+  }
+
+  @AutoValue
+  @JsonDeserialize(as = AutoValue_RefasterRuleCollectionTestExtractor_RefasterTestCase.class)
+  abstract static class RefasterTestCase {
+    static RefasterTestCase create(String name, String content) {
+      return new AutoValue_RefasterRuleCollectionTestExtractor_RefasterTestCase(name, content);
```
Stephan202, line 169: Pitest flags that these parameters can be swapped without failing a test. That's because we only validate that the type can be (de)serialized, without asserting the exact serialization format. That's a conscious choice. The final PR will improve coverage in this respect.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ImmutableMultisetRulesTestInput.java
```diff
@@ -26,6 +26,7 @@ ImmutableMultiset<ImmutableMultiset<Integer>> testEmptyImmutableMultiset() {
         Stream.<Integer>empty().collect(toImmutableMultiset()));
   }
 
+  @SuppressWarnings("unchecked")
```
Stephan202, line 29: Here and below: these annotations are now required, as the files are now compiled.

Stephan202 (commented):
Added some context. I also ported these changes to the [`website`](https://github.com/PicnicSupermarket/error-prone-support/compare/website) branch, where one can see how this all ties together.

Stephan202:
Rebased and added a commit to resolve a SonarCloud violation. Filed WIP PR #1320 related to this.

rickie (approved):
Only tiny suggestions.

Code looks ✨ !

documentation-support/src/main/java/tech/picnic/errorprone/documentation/RefasterRuleCollectionTestExtractor.java
Stephan202: Hmm, I generally prefer `< 0`, because any negative value isn't a valid index. We do this in a bunch of other places too:
```
$ git grep -i 'index.*< 0' | wc -l
8
```

documentation-support/src/main/java/tech/picnic/errorprone/documentation/RefasterRuleCollectionTestExtractor.java
```diff
@@ -0,0 +1,176 @@
...
+            getRuleCollectionName(tree),
+            isInputFile(sourceFile),
+            getRefasterTestCases(tree, state)));
+  }
+
+  private static String getRuleCollectionName(ClassTree tree) {
+    String className = tree.getSimpleName().toString();
+
+    // XXX: Instead of throwing an error here, it'd be nicer to have a bug checker validate key
+    // aspects of `RefasterRuleCollectionRefasterTestCase` subtypes.
```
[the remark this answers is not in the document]
Stephan202: I renamed things a few times, and used "global search and replace". Likely a result of that :sweat_smile:

Stephan202:
Tnx for the review @rickie! :rocket:

documentation-support/src/main/java/tech/picnic/errorprone/documentation/RefasterRuleCollectionTestExtractor.java
```diff
@@ -0,0 +1,176 @@
...
+            sourceFile,
+            getRuleCollectionName(tree),
+            isInputFile(sourceFile),
+            getRefasterTestCases(tree, state)));
+  }
+
+  private static String getRuleCollectionName(ClassTree tree) {
+    String className = tree.getSimpleName().toString();
+
+    // XXX: Instead of throwing an error here, it'd be nicer to have a bug checker validate key
```
mohamedsamehsalah, line 72: Unless I am missing something, you can also argue that throwing an error provides faster feedback when developing / running tests locally.

mohamedsamehsalah (approved):
🧠 🧠 🧠

documentation-support/src/main/java/tech/picnic/errorprone/documentation/RefasterRuleCollectionTestExtractor.java
Stephan202: The bug checker would also run at compile time, but would provide a (much) better error message.

Stephan202:
Synced the `website` branch. Not _too_ much logic now remains:
```
$ git diff origin/master --stat | tail -n 1
 19 files changed, 835 insertions(+), 77 deletions(-)
```
Largest remaining effort will be to write proper tests.

PR #1320, opened 2024-09-03
Introduce Refaster rules for Sonar's java:S4635 rule

Stephan202:
Suggested commit message:
```
Introduce Refaster rules for Sonar's java:S4635 rule (#1320)

As well as two related expressions that can be optimized.

See:
- https://sonarcloud.io/organizations/picnic-technologies/rules?open=java%3AS4635&rule_key=java%3AS4635
- https://github.com/SonarSource/sonar-java/blob/26157927314eed0105449e8c7bc7e86c34148a83/java-checks/src/main/java/org/sonar/java/checks/StringOffsetMethodsCheck.java
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
@@ -244,4 +244,72 @@ final class StringRules {
       return Utf8.encodedLength(str);
     }
   }
+
+  /** Prefer {@link String#indexOf(int, int)} over less efficient alternatives. */
+  static final class StringIndexOfChar {
+    @BeforeTemplate
+    int before(String string, int ch, int fromIndex) {
+      return string.substring(fromIndex).indexOf(ch);
+    }
+
+    @AfterTemplate
+    int after(String string, int ch, int fromIndex) {
+      return string.indexOf(ch, fromIndex);
+    }
+  }
+
+  /** Prefer {@link String#indexOf(String, int)} over less efficient alternatives. */
+  static final class StringIndexOfString {
+    @BeforeTemplate
+    int before(String string, String substring, int fromIndex) {
+      return string.substring(fromIndex).indexOf(substring);
+    }
+
+    @AfterTemplate
+    int after(String string, String substring, int fromIndex) {
+      return string.indexOf(substring, fromIndex);
+    }
+  }
+
+  // XXX: Once we compile Refaster templates with JDK 21 also suggest `String#indexOf(int, int,
+  // int)` and `String#indexOf(String, int, int)`.
+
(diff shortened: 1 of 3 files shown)


Stephan202:
Moved PR back to draft mode: Sonar claims that this rule applies to `lastIndexOf` in the same way that it does to `indexOf`, but as-is I don't think it does: the former method treats the specified index as an upper bound, rather than a lowerbound. (This is a nice example where automated test generation based on before- and after-templates would have been helpful.)

Stephan202:
K, added two commits. IIUC Sonar's [`.substring`](https://github.com/SonarSource/sonar-java/blob/26157927314eed0105449e8c7bc7e86c34148a83/java-checks/src/main/java/org/sonar/java/checks/StringOffsetMethodsCheck.java#L36-L40) matcher shouldn't unconditionally match the 1-arg method variant. I might look into filing a ticket for this.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
```diff
@@ -244,4 +244,105 @@ int after(String str) {
       return Utf8.encodedLength(str);
     }
   }
+
+  /** Prefer {@link String#indexOf(int, int)} over less efficient alternatives. */
+  static final class StringIndexOfChar {
```
Stephan202, line 249: Here and below: these rules will pose a nice challenge once we start to generate rule naming suggestions. For now I kept it simple.

Stephan202 (commented):
Alright, I had another look with a fresh head, and I think I now understand what the rule's documentation means to convey. Added two more rules.

mohamedsamehsalah (approved):
LGTM 🚀

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
```diff
@@ -244,4 +244,105 @@ int after(String str) {
...
+      return string.substring(fromIndex).indexOf(substring);
+    }
+
+    @AfterTemplate
+    int after(String string, String substring, int fromIndex) {
+      return Math.max(-1, string.indexOf(substring, fromIndex) - fromIndex);
+    }
+  }
+
+  // XXX: Once we compile Refaster templates with JDK 21 also suggest `String#indexOf(int, int,
```
rickie, line 276: Should we move this to the top of the class? Now it's a bit hidden 🤔.

rickie (approved):

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
Stephan202: Ah, I placed it here because those rules would be located ~here. But either works for me. :+1:

PR #1365, opened 2024-10-22
Introduce `MicrometerRules` Refaster rule collection

danylo-naumenko:
Suggested commit message:
```
Introduce `MicrometerRules` Refaster rule collection (#1365)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MicrometerRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MicrometerRules.java
@@ -1,0 +1,85 @@
+package tech.picnic.errorprone.refasterrules;
+
+import com.google.common.collect.ImmutableCollection;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.errorprone.refaster.Refaster;
+import com.google.errorprone.refaster.annotation.AfterTemplate;
+import com.google.errorprone.refaster.annotation.BeforeTemplate;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
+
+/** Refaster rules related to expressions dealing with Micrometer. */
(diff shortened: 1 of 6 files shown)


error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/MicrometerRulesTestOutput.java
```diff
@@ -0,0 +1,60 @@
...
+import tech.picnic.errorprone.refaster.test.RefasterRuleCollectionTestCase;
+
+final class MicrometerRulesTest implements RefasterRuleCollectionTestCase {
+  @Override
+  public ImmutableSet<Object> elidedTypesAndStaticImports() {
+    return ImmutableSet.of(ImmutableList.class);
+  }
+
+  ImmutableSet<Iterable<Tag>> testTagsOf1() {
+    return ImmutableSet.of(Tags.of(Tag.of("foo", "bar")), Tags.of(Tag.of("foo", "bar")));
```
rickie: We try to use distinct values in a test such that we are always sure it replaced the correct thing. Because the TagsOf4 and TagsOf5 have too many values to only use words like foo bar baz qux etc. I added the v1 v2 as values.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MicrometerRules.java
```diff
@@ -0,0 +1,85 @@
...
+      return Refaster.anyOf(ImmutableSet.of(tag), ImmutableList.of(tag));
+    }
+
+    @AfterTemplate
+    Iterable<Tag> after(Tag tag) {
+      return Tags.of(tag);
+    }
+  }
+
+  /** Prefer {@link Tags} over other immutable {@link Iterable}'s. */
```
rickie: Proposing a slight tweak to the Javadocs :).

rickie (approved):
Added a commit with some suggestions. Only style tweaks, congrats on your first PR 🚀 !

Suggested commit message (PR title):
```
Introduce `MicrometerRules` Refaster rule collection (#1365)
```

PR #1373, opened 2024-10-25
Disable `ClassLiteralCast` Refaster rule

rickie:
Temporarily disable the `ClassRules#ClassLiteralCast` Refaster rule. We want to do a release and fixing this won't make it in time for the release probably.


mohamedsamehsalah (commented):
Since we agreed offline that we can move this to a Bugchecker, does it make sens eto simply remove the rule?

mohamedsamehsalah:
We can close in favor of #1381

Stephan202:
We'll go with #1381 instead.

PR #1379, opened 2024-10-27
Introduce `NameContentEquals` Refaster rule

Stephan202:
As also flagged [here](https://github.com/PicnicSupermarket/error-prone-support/pull/783#discussion_r1502274717) and [here](https://github.com/PicnicSupermarket/error-prone-support/pull/794#discussion_r1818105009).

Suggested commit message:
```
Introduce `NameContentEquals` Refaster rule (#1379)
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BugCheckerRules.java
```diff
@@ -67,4 +68,22 @@ String after(String value) {
       return Constants.format(value);
     }
   }
+
+  /** Prefer {@link Name#contentEquals(CharSequence)} over more verbose alternatives. */
+  static final class NameContentEquals {
```
Stephan202, line 73: We can debate whether `BugCheckerRules` is the best place for this rule, but it's in line with other rule definitions in this class.

mohamedsamehsalah (approved):
🚀 🚀

rickie (approved):

PR #1381, opened 2024-10-28
Introduce `ClassCastLambdaUsage` check

mohamedsamehsalah:
Suggested commit message
```
Introduce `ClassCastLambdaUsage` check (#1381)

This new check replaces the `ClassLiteralCast` Refaster rule, as the 
latter produced invalid code by suggesting expressions of the form
`T.class::cast` for generic `T`.
```


error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ClassCastTest.java
```diff
@@ -0,0 +1,62 @@
...
+            "    // BUG: Diagnostic contains:",
+            "    Optional.of(foo).map(i -> (Integer) i);",
+            "",
+            "    Optional.of(foo).map(i -> 2);",
+            "    Optional.of(foo).map(i -> (Integer) 2);",
+            "    Optional.of(foo).map(i -> bar);",
+            "    Optional.of(foo).map(i -> (Integer) bar);",
+            "",
+            "    ImmutableList.of(Set.of(foo)).stream().map(l -> (ImmutableSet<Number>) l);",
+            "    ImmutableList.of(Set.of(foo)).stream().map(l -> (ImmutableSet<?>) l);",
```
Stephan202: Would have been nice to stick a bit closer to the test cases in `IsInstanceLambdaUsageTest`. W.r.t. the `ImmutableSet<Number>` case: let's more explicitly generate a stream of raw type elements.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ClassCast.java
```diff
@@ -0,0 +1,64 @@
...
+    if (!ASTHelpers.getSymbol(param).equals(ASTHelpers.getSymbol(typeCast.getExpression()))) {
+      return Description.NO_MATCH;
+    }
+
+    return describeMatch(tree, SuggestedFix.replace(tree, classCast + ".class::cast"));
+  }
+
+  private static boolean isGenericCastExpression(String expression) {
+    return expression.contains("<") && expression.contains(">");
+  }
```
Stephan202: We shouldn't be checking the source code like this, but instead ask the type system :grimacing:.

Stephan202 (approved):
This PR is an alternative to #1373.

Added a commit and updated the suggested commit message. Tnx @mohamedsamehsalah!

rickie (approved):
Nice work @mohamedsamehsalah @Stephan202 🚀 !

PR #1283, opened 2024-08-11
Improve `Optional#orElse{,Get}` support

Stephan202:
Suggested commit message:
```
Improve `Optional#orElse{,Get}` support

Summary of changes:
- Consolidate the `OptionalOrElseGet` Refaster rule and the 
  `OptionalOrElse` bug checker into the latter, keeping the best of
  both.
- Rename the `OptionalOrElse` bug checker to `OptionalOrElseGet` to
  avoid confusion.
- Replace the `IsLikelyTrivialComputation` matcher with the inverse
  `RequiresComputation` variant.
- Introduce an `OptionalOrElse` Refaster rule that simplifies
  expressions in the "opposite direction".
```

PR #1387, opened 2024-11-01
Fix `error-prone-contrib` runtime classpath

Stephan202:
This change resolves an issue caused by #1283, and flagged by @pzygielo [here](https://github.com/PicnicSupermarket/error-prone-support/pull/1283#issuecomment-2451777445). Thanks Piotrek!

(The bug didn't manifest while testing the release, as in all relevant contexts we also have `refaster-runner` on the annotation processor classpath, and that module has a _runtime_ dependency on `refaster-support`.)

Suggested commit message:
```
Fix `error-prone-contrib` runtime classpath (#1387)
```

We should cut a new release with this fix.

### DIFF
--- a/error-prone-contrib/pom.xml
+++ b/error-prone-contrib/pom.xml
@@ -55,7 +55,12 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>refaster-support</artifactId>
-            <scope>provided</scope>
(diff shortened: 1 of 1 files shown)


mohamedsamehsalah (approved):
📖

rickie (approved):

PR #1391, opened 2024-11-02
Run GitHub Actions workflows on `ubuntu-24.04`

Stephan202:
Suggested commit message:
```
Run GitHub Actions workflows on `ubuntu-24.04` (#1391)

See https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ ubuntu-22.04 ]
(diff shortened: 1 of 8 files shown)


mohamedsamehsalah (approved):

Stephan202:
Right, so we should first upgrade `ruby/setup-ruby`. I filed #1392.

rickie (approved):
Already approving :).

PR #1393, opened 2024-11-03
Introduce `{Mono|Flux}#using{When}` reactor refaster rules

mohamedsamehsalah:
Suggested commit message
```
Introduce `{Mono|Flux}#using{When}` reactor refaster rules
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -541,6 +541,284 @@ Mono<T> after(Mono<T> mono) {
     }
   }
 
+  /**
+   * Don't unnecessarily transform a {@link Flux#using(Callable, Function)} to a mono, instead use
+   * the equivalent API provided by {@link Mono}.
```
mohamedsamehsalah, line 546: Here and below, technically it is not strictly equivalent, unless we return in the supplier's `Function` a `Mono` in both the before and after templates.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -541,6 +541,284 @@ Mono<T> after(Mono<T> mono) {
...
+
+  /**
+   * Don't unnecessarily transform a {@link Mono#using(Callable, Function)} to a flux, instead use
+   * the equivalent API provided by {@link Flux}.
+   */
+  static final class FluxUsing<
+      D extends AutoCloseable, T, P extends Publisher<? extends T>, M extends Mono<? extends T>> {
+    @BeforeTemplate
+    Flux<T> before(Callable<D> resourceSupplier, Function<D, M> sourceSupplier) {
+      return Mono.using(resourceSupplier, sourceSupplier).flux();
```
Venorcis, line 691: not sure I agree with this one; using `Mono#flux` at least makes explicit that it can emit at most 1 element (and sometimes a `Flux` is needed to adhere to method signatures)

Venorcis (commented):
one could think of more very weird (but very straightforward) ones like:

- `Flux.error(...).single()` / `Mono.error(...).flux`
- `Flux.empty().single()` / `Mono.empty().flux`
- `Flux.just(...).single()` / `Mono.just(...).flux` _(not the Flux array signature ofc.)_

and even most `zip` \ `zipWith` variants I guess

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
mohamedsamehsalah: > not sure I agree with this one; using Mono#flux at least makes explicit that it can emit at most 1 element (and sometimes a Flux is needed to adhere to method signatures)

If you check the method signature (Also what I was trying to explain [here](https://github.com/PicnicSupermarket/error-prone-support/pull/1393#discussion_r1827310197)), is that the `Function` of the _sourceSupplier_ explicitly returns a `Mono`, so by definition, we know the whole chain returns one element. So the `Mono` > `Flux` conversion seems redundant to me.

WDYT?

mohamedsamehsalah (commented):
Thanks for review @Venorcis 🙏 
> one could think of more very weird (but very straightforward) ones like:
> 
> * `Flux.error(...).single()` / `Mono.error(...).flux`
> * `Flux.empty().single()` / `Mono.empty().flux`
> * `Flux.just(...).single()` / `Mono.just(...).flux` _(not the Flux array signature ofc.)_
> 
> and even most `zip` \ `zipWith` variants I guess

I agree. I wanted to limit the scope of this refaster rule to the `#using` API. Which started when I unknowingly had the same pattern in one of my [PRs](https://github.com/PicnicSupermarket/picnic-store/pull/5939#discussion_r1810574286).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Venorcis: `Flux.using` with a `Mono` sourceSupplier looks worse to me as IMO that just obfuscates what's going on, and thus I'd favour the explicit conversion with `flux()`

mohamedsamehsalah: Yeah, that makes sense to me 👌 

Just to make sure I understand your stance, is this a "choosing a lesser evil" type of design choice? Or do you think that what we're attempting to re-write is actually a valid case?

Venorcis: > Or do you think that what we're attempting to re-write is actually a valid case

Yeah, I think `Mono#flux` should generally not be rewritten (bar some very specific exceptions where it's ultra-clear the `Flux` variant only supplies 1 element, i.e. with `just` and `error`)

mohamedsamehsalah: Fair argument, I honestly have no strong opinion here.

Let's see what others think, then I will happily close the PR 😄 Thanks @Venorcis

Venorcis: I think the other way around is very valid though 😄 i.e. don't use `Flux#single` when you can directly use a `Mono` instead

mohamedsamehsalah: (Though it is hard to know if the pattern was applied because it was intentionally specifying a flux that emits one element; or simply an unawareness of the equivalent APIs in `Flux`; Let's assume the former)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -541,6 +541,145 @@ Mono<T> after(Mono<T> mono) {
...
 
+  /**
+   * Don't unnecessarily transform a {@link Flux#using(Callable, Function)} to a mono, instead use
+   * the equivalent API provided by {@link Mono}.
+   */
+  static final class MonoUsing<
+      D extends AutoCloseable, T, P extends Publisher<? extends T>, M extends Mono<? extends T>> {
+    @BeforeTemplate
+    Mono<T> before(Callable<D> resourceSupplier, Function<D, P> sourceSupplier) {
+      return Flux.using(resourceSupplier, sourceSupplier).single();
```
mohamedsamehsalah, line 552: _(Here & below)_
This refaster rule will cause a compiler error if the publisher in the supplier is not of type Mono. 😬 

(Need to think how I can improve the matching here)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
mohamedsamehsalah: > ..using Mono#flux at least makes explicit that it can emit at most 1 element ...

Though it is hard to know if the pattern was applied because it was intentionally specifying a flux that emits one element; or simply an unawareness of the equivalent APIs in `Flux`; Let's assume the former 👍 

Applied ✅

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
mohamedsamehsalah: Introduced a before template matcher [here](https://github.com/PicnicSupermarket/error-prone-support/pull/1393/commits/881cc12e9a8fded6fa34da8de6b7e615eb5ccdb5).

rickie:
As discussed offline; moving to the next milestone :).

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -556,6 +557,152 @@ Mono<T> after(Mono<T> mono) {
...
+  /**
+   * Don't unnecessarily transform a {@link Flux#using(Callable, Function)} to a mono, instead use
+   * the equivalent API provided by {@link Mono}.
+   */
+  static final class MonoUsing<
+      D extends AutoCloseable, T, P extends Publisher<? extends T>, M extends Mono<? extends T>> {
+    @BeforeTemplate
+    Mono<T> before(
+        Callable<D> resourceSupplier,
+        @Matches(IsFunctionReturningMono.class) Function<D, P> sourceSupplier) {
```
rickie, line 594: @Stephan202 had the idea to see if we can simply define P here by saying `P extends Mono<? extends T>` but I couldn't get that to work. I think as then the `Publisher` overload is not being used and Refaster cannot match it anymore.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -556,6 +557,152 @@ Mono<T> after(Mono<T> mono) {
...
+    }
+
+    @AfterTemplate
+    Mono<T> after(Callable<D> resourceSupplier, Function<D, M> sourceSupplier) {
+      return Mono.using(resourceSupplier, sourceSupplier);
+    }
+  }
+
+  /**
+   * Don't unnecessarily transform a {@link Flux#using(Callable, Function, boolean)} to a mono,
```
rickie: Let's link mono.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -556,6 +557,152 @@ Mono<T> after(Mono<T> mono) {
...
+        Function<D, P2> asyncCleanup) {
+      return Mono.usingWhen(resourceSupplier, resourceClosure, asyncCleanup);
+    }
+  }
+
+  /**
+   * Don't unnecessarily transform a {@link Flux#usingWhen(Publisher, Function, Function,
+   * BiFunction, Function)} to a mono, instead use the equivalent API provided by {@link Mono}.
+   */
+  static final class MonoUsingWhen2<
```
rickie: Hmm, the naming is a bit harder for this one because of the different overlaps... I did a proposal but I'm not entirely happy with it yet 🤔.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -210,6 +211,57 @@ Mono<Integer> testMonoSingle() {
...
+        Flux.usingWhen(Mono.just("foo"), foo -> Mono.just("bar"), foo -> Mono.just("baz")).single(),
+        Flux.usingWhen(Mono.just("foo"), foo -> Flux.just("bar"), foo -> Mono.just("baz"))
+            .single());
+  }
+
+  ImmutableSet<Mono<String>> testMonoUsingWhen2() {
+    return ImmutableSet.of(
+        Flux.usingWhen(
+                Mono.just("foo"),
+                foo -> Mono.just("bar"),
```
rickie: We can simplify these a little bit :)

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -210,6 +211,57 @@ Mono<Integer> testMonoSingle() {
...
+                foo -> Mono.just("baz"),
+                (foo, e) -> Mono.just("qux"),
+                foo -> Mono.just("thud"))
```
rickie:
```suggestion
                foo -> Mono.just("quux"))
```
😉

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestOutput.java
```diff
@@ -209,6 +210,55 @@ Mono<Integer> testMonoSingle() {
...
+  ImmutableSet<Mono<String>> testMonoUsing2Eager() {
+    return ImmutableSet.of(
+        Mono.using(() -> "foo", foo -> Mono.just("bar"), foo -> {}, false),
+        Flux.using(() -> "foo", foo -> Flux.just("bar"), foo -> {}, false).single());
+  }
+
+  ImmutableSet<Mono<String>> testMonoUsingWhen() {
+    return ImmutableSet.of(
+        Mono.usingWhen(Mono.just("foo"), foo -> Mono.just("bar"), foo -> Mono.just("baz")),
+        Flux.usingWhen(Mono.just("foo"), foo -> Flux.just("bar"), foo -> Mono.just("baz"))
```
rickie: Let's continue with the strings :)

rickie (approved):
LGTM! 🐌

rickie:
Rebased and resolved conflicts. @Stephan202  it's ready for review :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
mohamedsamehsalah: New names LGTM 👍

PR #1395, opened 2024-11-04
Introduce `Instant{Identity,TruncatedTo{Milliseconds,Seconds}}` Refaster rules

werli:
Some trivial rules I noticed when drafting up other rules I needed to omit.


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
```diff
@@ -142,6 +142,24 @@ OffsetDateTime after(Instant instant, ZoneId zoneId) {
...
+  /** Don't unnecessarily transform an {@link Instant} to an equivalent instance. */
+  static final class InstantIdentity {
+    @BeforeTemplate
+    Instant before(Instant instant) {
+      return Refaster.anyOf(
+          instant.plus(Duration.ZERO),
+          instant.minus(Duration.ZERO),
+          Instant.parse(instant.toString()),
+          Instant.ofEpochMilli(instant.toEpochMilli()),
+          Instant.ofEpochSecond(instant.getEpochSecond()));
```
Stephan202: These two aren't actually identity transformations, as they drop the nano and millisecond components, respectively. Let's introduce rules to suggest using `Instant#truncatedTo` instead.

Stephan202 (approved):
```
Introduce `Instant{Identity,TruncatedTo{Milliseconds,Seconds}}` Refaster rules (#1395)

While there, have `IdentityConversion` support `Instant#from`.

rickie (approved):
Nice cleanup rules :rocket:!

rickie:
@Venorcis anything from your side? :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TimeRules.java
werli: Makes sense, thanks 💪

Venorcis (approved):
LGTM 👍🏻

PR #1399, opened 2024-11-06
Enable colored Maven output

rickie:


mohamedsamehsalah (approved):
🎨

timtebeek (approved):

rickie:
Looked into this with @timtebeek. @Stephan202 You asked to investigate a little bit what it means to have the batchmode + colored output and if we need it. I think we discussed that the batch mode is responsible for removing the color, so we might not need both. According to [Sonatype docs](https://books.sonatype.com/mvnref-book/reference/running-sect-options.html):

> Batch mode is essential if you need to run Maven in a non-interactive, continuous integration environment. When running in non-interactive mode, Maven will never stop to accept input from the user. Instead, it will use sensible default values when it requires input.

This means that it's suggested to keep the batch mode enabled. To make the GHA output more readable it would help to add this flag. And should give any harm besides simply getting some nice colors :sparkles:.

---

Thanks @timtebeek for your determination to get this issue across the finish line.

Stephan202:
I pushed a change after configuring `MAVEN_ARGS='--batch-mode -Dstyle.color=never'`. Looks like this works: no color, and no download progress output (that's the primary reason for setting `--batch-mode`: reducing log output and speeding up the build).

If this all passes I'll set `MAVEN_ARGS='--batch-mode -Dstyle.color=always;` and retrigger one of the builds. If that yields colors, then I'm satisfied that this GitHub Actions configuration is properly picked up.

image

Stephan202:
Somehow this doesn't work; still no colored output in GitHub Actions. It's getting late, so will defer this to after the 0.24.0 release.

Stephan202:
The suggested commit message I optimistically already drafted (still hopeful I can get this to work later):
```
Stop forcing Maven to run in batch mode (#1399)

This way local console output will be colored again. Builds on GitHub
actions now run with `MAVEN_ARGS='--batch-mode -Dstyle.color=always'`,
such that they also have colored logs, without spamming the logs with
download progress updates.
```

timtebeek:
Finally seeing some colors in the output here! 🌈
image

.github/workflows/pitest-analyze-pr.yml
```diff
@@ -7,6 +7,8 @@ on:
   pull_request:
 permissions:
   contents: read
+env:
+  MAVEN_ARGS: --batch-mode -Dstyle.color=always -T1C
```
Stephan202, line 11: Previously we didn't parallelize the Pitest analysis; let's see whether we get away with this, resource utilization-wise.

.github/workflows/reviewdog.yml
```diff
@@ -41,15 +43,19 @@ jobs:
...
         # XXX: Cover additional output formats, besides Checkstyle and javac.
         # XXX: Run with `-Dmaven.test.failure.ignore` instead of `-DskipTests`,
         # and also report Surefire test failures.
+        # XXX: Distribute these commands across three steps?
         run: |
-          mvn -T1C clean verify -Pself-check -DskipTests -Dverification.warn \
+          mvn clean verify -Pself-check -DskipTests -Dverification.warn | tee /tmp/build.log
+          # The ANSI escape sequences that color the build output are dropped
+          # before passing the build log to Reviewdog.
+          sed 's/\x1b\[[0-9;]*m//g' < /tmp/build.log \
```
Stephan202, line 58: Tested this setup by temporarily introducing a self-check violation.

PR #1425, opened 2024-11-18
Upgrade Checkstyle integration test target 10.14.0 -> 10.21.1

rickie:
Suggested commit message (similar to the previous one):
```
Sync Checkstyle integration test (#1425)
    
Summary of changes:
- Update `checkstyle-init.patch`. 
- Update `checkstyle-expected-changes.patch`.
- Update `checkstyle-expected-warnings.txt`
- Test against version 10.21.1 rather than 10.14.0.
```

Or my slightly preferred alternative:
```
Sync Checkstyle integration test (#1425)
    
By updating and testing against 10.21.1. 
```


integration-tests/checkstyle-init.patch
```diff
@@ -1,7 +1,7 @@
 --- a/pom.xml
 +++ b/pom.xml
-@@ -366,6 +366,12 @@
-       <version>1.4.1</version>
+@@ -368,6 +368,12 @@
+       <version>1.4.4</version>
```
rickie, line 4: Hmm, this one is unexpected, if you don't recognize this one, I'll try to remove it @Stephan202 😄

Stephan202: This is a change in the patch context, not a change in the changes made by the patch. So this is not unexpected :+1:

Stephan202 (approved):
Suggested commit message:
```
Upgrade Checkstyle integration test target 10.14.0 -> 10.21.1 (#1425)
```

(This phrasing is a bit different from what we did before, but if we start to file this kind of PR more often, then the summary messages will be more informative.)

rickie:
Instead of `target` we could say `tag`? 🤔

Stephan202:
Well, the integration test _targets_ a specific Checkstyle version. The specified tag is the test target.

mohamedsamehsalah (approved):
I meant to approve ✅

Stephan202:
Re 1: not opposed, but as a non-maintainer reading the release notes, I'm not sure I'd care much about the details of our testing setup. We could of course introduce a separate label and then map it to the same release notes section. If you think that helps with filtering: works for me. (

Re 2: we'd have to check whether this is compatible with GitHub actions, but perhaps it'd be even nicer if stuff "just worked", like [this](https://github.com/PicnicSupermarket/error-prone-support/blob/27e9fe79a586104ed6d0c4c3799fb5d2a44a5903/website/generate-version-compatibility-overview.sh#L13-L17). The alternative is to parse `mvn -version` output and compare version numbers (possibly also for Maven): possible but not trivial. (But perhaps `sort -h` works well enough, didn't test.)

PR #1467, opened 2024-12-18
Introduce `StreamMapFilter` Refaster rule

zphilipp:
Prefer
```java
stream.map(map::get).filter(Objects::nonNull)
``` 
over
```java
stream.filter(map::containsKey).map(map::get)
```

### Suggested commit message
```
Introduce `StreamMapFilter` Refaster rule (#1467)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
@@ -298,18 +298,17 @@ final class StreamRules {
   }
 
   /**
-   * When using {@link Map#get(Object)} in a {@link Stream#map(Function)} operation, prefer
-   * filtering null values after the map operation with {@link Objects#nonNull(Object)} over
-   * filtering beforehand with {@link Map#containsKey(Object)} to avoid a double lookup.
+   * Prefer an unconditional {@link Map#get(Object)} call followed by a {@code null} check over a
+   * call to {@link Map#containsKey(Object)}, as the former avoids a second lookup operation.
    */
-  static final class StreamMapFilter<K, V> {
+  static final class StreamMapFilter<T, K, V> {
     @BeforeTemplate
-    Stream<V> before(Stream<K> stream, Map<K, V> map) {
+    Stream<V> before(Stream<T> stream, Map<K, V> map) {
       return stream.filter(map::containsKey).map(map::get);
     }
 
     @AfterTemplate
-    Stream<V> after(Stream<K> stream, Map<K, V> map) {
+    Stream<V> after(Stream<T> stream, Map<K, V> map) {
       return stream.map(map::get).filter(Objects::nonNull);
     }
   }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
@@ -297,7 +297,12 @@ final class StreamRules {
     }
   }
 
-  static final class StreamMapToValuesFromMap<K, V> {
+  /**
+   * When using {@link Map#get(Object)} in a {@link Stream#map(Function)} operation, prefer
+   * filtering null values after the map operation with {@link Objects#nonNull(Object)} over
+   * filtering beforehand with {@link Map#containsKey(Object)} to avoid a double lookup.
+   */
+  static final class StreamMapFilter<K, V> {
     @BeforeTemplate
     Stream<V> before(Stream<K> stream, Map<K, V> map) {
       return stream.filter(map::containsKey).map(map::get);
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
@@ -297,6 +297,18 @@ final class StreamRules {
     }
   }
 
+  static final class StreamMapToValuesFromMap<K, V> {
+    @BeforeTemplate
+    Stream<V> before(Stream<K> stream, Map<K, V> map) {
+      return stream.filter(map::containsKey).map(map::get);
+    }
+
+    @AfterTemplate
+    Stream<V> after(Stream<K> stream, Map<K, V> map) {
+      return stream.map(map::get).filter(Objects::nonNull);
+    }
+  }
+
   static final class StreamMin<T> {
     @BeforeTemplate
     @SuppressWarnings("java:S4266" /* This violation will be rewritten. */)
--- a/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StreamRulesTestInput.java
+++ b/error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StreamRulesTestInput.java
@@ -141,6 +141,11 @@ final class StreamRulesTest implements RefasterRuleCollectionTestCase {
     return Stream.of(1).findFirst().isPresent();
   }
 
+  Stream<String> testStreamMapToValuesFromMap() {
+    ImmutableMap<String, String> map = ImmutableMap.of("foo", "bar");
+    return Stream.of("foo").filter(map::containsKey).map(map::get);
+  }
+
   ImmutableSet<Optional<String>> testStreamMin() {
     return ImmutableSet.of(
         Stream.of("foo").max(comparingInt(String::length).reversed()),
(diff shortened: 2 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -297,6 +297,18 @@ boolean after(Stream<T> stream) {
     }
   }
 
+  static final class StreamMapToValuesFromMap<K, V> {
```
werli, line 300: (a) Let's add some light-weight JavaDoc on this class? E.g. similar to `StreamNoneMatch`.
(b) The Refaster rules are most often called after the resulting code, see other examples in this class. In this scenario, this would be `StreamMapFilter`. Can we rename it? :eyes:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
werli, line 1: (Unrelated to this code)
The same pattern may also be applicable for `Flux` where we should call `Flux#mapNotNull`. Maybe good follow up? :eyes:

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
zphilipp: I added documentation and renamed the rule. The name feels a bit too generic to me though, as it could also apply to other patterns. WDYT?

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
zphilipp: Sure, I can create another PR. As @mohamedsamehsalah tells me, this will get me closer to my sticker. 😁

rickie: You will definitely receive one for this contribution already!

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -297,6 +297,23 @@ boolean after(Stream<T> stream) {
...
 
+  /**
+   * When using {@link Map#get(Object)} in a {@link Stream#map(Function)} operation, prefer
+   * filtering null values after the map operation with {@link Objects#nonNull(Object)} over
+   * filtering beforehand with {@link Map#containsKey(Object)} to avoid a double lookup.
+   */
+  static final class StreamMapFilter<K, V> {
+    @BeforeTemplate
+    Stream<V> before(Stream<K> stream, Map<K, V> map) {
+      return stream.filter(map::containsKey).map(map::get);
```
rickie, line 308: I'm curious about the performance. As usually we try to moving filtering to the "earliest" spot, right? I guess preventing the double lookup makes up for that 🤔 ? 

Again a reason why having a "benchmarking-framework" in place in EPS would be useful.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/StreamRulesTestInput.java
```diff
@@ -141,6 +141,11 @@ boolean testStreamFindAnyIsPresent() {
     return Stream.of(1).findFirst().isPresent();
   }
 
+  Stream<String> testStreamMapFilter() {
+    ImmutableMap<String, String> map = ImmutableMap.of("foo", "bar");
```
rickie, line 145: We can inline these values :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
```diff
@@ -297,6 +297,23 @@ boolean after(Stream<T> stream) {
     }
   }
 
+  /**
+   * When using {@link Map#get(Object)} in a {@link Stream#map(Function)} operation, prefer
```
rickie, line 301: I would rephrase this to `Prefer {@link ..} over {@link ..} to avoid double look-up` or something :).

rickie (commented):
Nice one! Some small points (didn't apply them yet 😬)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
Stephan202: Indeed we should find time to progress on such a framework. For now I wrote a custom benchmark:
```java
import static com.google.common.collect.ImmutableMap.toImmutableMap;
import static java.util.function.Function.identity;

import com.google.common.collect.ImmutableList;
import com.google.common.collect.ImmutableMap;
import java.util.Objects;
import java.util.concurrent.TimeUnit;
import java.util.regex.Pattern;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;
import org.openjdk.jmh.profile.GCProfiler;
import org.openjdk.jmh.runner.Runner;
import org.openjdk.jmh.runner.RunnerException;
import org.openjdk.jmh.runner.options.OptionsBuilder;

@State(Scope.Benchmark)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.MICROSECONDS)
@Fork(
    jvmArgs = {"-Xms256M", "-Xmx256M"},
    value = 1)
@Warmup(iterations = 5)
@Measurement(iterations = 10) 
public class Benchmark {
  private ImmutableList<String> elements;
  private ImmutableMap<String, String> map;

  public static void main(String[] args) throws RunnerException {
    String testRegex = Pattern.quote(Benchmark.class.getCanonicalName());
    new Runner(new OptionsBuilder().include(testRegex).addProfiler(GCProfiler.class).build()).run();
  }

  @Setup
  public void setUp() {
    elements = ImmutableList.of("a", "b", "c", "d", "e", "f", "g", "h", "i", "j");
    map =
        elements.stream()
            .filter(e -> e.charAt(0) % 2 == 0)
            .collect(toImmutableMap(identity(), e -> e + e));
  }

  @Benchmark
  public long filterFirst() {
    return elements.stream().filter(map::containsKey).map(map::get).count();
  }

  @Benchmark
  public long mapFirst() {
    return elements.stream().map(map::get).filter(Objects::nonNull).count();
  }
}
```

This yielded the following results:
```
Benchmark                                             Mode  Cnt     Score    Error   Units
Benchmark.filterFirst                     avgt   10     0.116 ±  0.001   us/op
Benchmark.filterFirst:gc.alloc.rate       avgt   10  2559.065 ± 21.986  MB/sec
Benchmark.filterFirst:gc.alloc.rate.norm  avgt   10   312.000 ±  0.001    B/op
Benchmark.filterFirst:gc.count            avgt   10  1687.000           counts
Benchmark.filterFirst:gc.time             avgt   10  1060.000               ms
Benchmark.mapFirst                        avgt   10     0.095 ±  0.001   us/op
Benchmark.mapFirst:gc.alloc.rate          avgt   10  2978.998 ± 14.750  MB/sec
Benchmark.mapFirst:gc.alloc.rate.norm     avgt   10   296.000 ±  0.001    B/op
Benchmark.mapFirst:gc.count               avgt   10  1964.000           counts
Benchmark.mapFirst:gc.time                avgt   10  1416.000               ms
```

So this looks like a good change, performance-wise :+1:

Stephan202 (approved):
Rebased and added a commit. Congrats on your first contribution @zphilipp! :tada:

rickie (approved):
Congrats on your first PR @zphilipp !

rickie (approved):
Congrats on your first PR @zphilipp !

rickie:
😓 idk why it's doing it twice now.

Anyway, anything from your side @werli :)?

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java
werli: Thanks you two for challenging this and showing that it actually results in an improvement. 💪

PR #1471, opened 2024-12-19
Introduce `AssertThatThrownBy*ExceptionRootCauseHasMessage` Refaster rules

rickie:
This PR aims to get rid of the following part of the `init-patch` of the new IT: https://github.com/PicnicSupermarket/error-prone-support/pull/1468/files#diff-6b99a1fb957cc4c8081a8a95423c965335a67c526f4b5b634aafa0d6912dc4dcR61. 

Not sure about the * part in the title, let me know what options are :D.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRules.java
@@ -61,6 +61,27 @@ final class AssertJThrowingCallableRules {
     }
   }
 
+  static final class AssertThatThrownByIllegalArgumentExceptionRootCauseHasMessage {
+    @BeforeTemplate
+    @SuppressWarnings(
+        "AssertThatThrownByIllegalArgumentException" /* This is a more specific template. */)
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRules.java
```diff
@@ -489,6 +572,32 @@ static final class AssertThatThrownByHasMessage {
...
+    AbstractObjectAssert<?, ?> after(
+        ThrowingCallable throwingCallable,
+        Class<? extends Throwable> exceptionType,
+        String message) {
+      return assertThatThrownBy(throwingCallable)
+          .isInstanceOf(exceptionType)
+          .rootCause()
+          .hasMessage(message);
+    }
+  }
```
Stephan202, line 599: We should make the return types more explicit, and replace `? extends Throwable` with a shared type variable `T`, but that applies to many rules in this class, so: for another time. (Ideally both these steps are automated.)

Stephan202 (approved):
Suggested commit message:
```
Introduce `AssertThatThrownBy*ExceptionRootCauseHasMessage` Refaster rules (#1471)
```

Not great, but works :)

mohamedsamehsalah (approved):
LGMT 🚀

PR #1484, opened 2024-12-28
Extend `JUnitToAssertJRules` Refaster rule collection

Stephan202:
Suggested commit message:
```
Extend `JUnitToAssertJRules` Refaster rule collection (#1484)

By migrating all remaining `assertArrayEquals` methods.
```

Analogous to #1483, this is me putting up a PR for a existing branch, this time after extending the work with the help of GitHub Copilot. (Also here, perhaps an Error Prone check is in order, but not today.)

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
@@ -5,6 +5,8 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.Assertions.fail;
+import static org.assertj.core.api.Assertions.offset;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
```diff
@@ -302,6 +272,436 @@
...
+  static final class AssertThatBooleanArrayWithFailMessageContainsExactly {
+    @BeforeTemplate
+    void before(boolean[] actual, String message, boolean[] expected) {
+      assertArrayEquals(expected, actual, message);
+    }
+
+    @AfterTemplate
+    @UseImportPolicy(STATIC_IMPORT_ALWAYS)
+    void after(boolean[] actual, String message, boolean[] expected) {
+      assertThat(actual).withFailMessage(message).containsExactly(expected);
```
mohamedsamehsalah, line 297: I wanted to suggest this to make it consistent with `AssertJThrowingCallableRules` where we do the message assertion after the checking the expected value.
```suggestion
      assertThat(actual).containsExactly(expected).withFailMessage(message);
```
But Intellij warns `Result of 'AbstractAssert. withFailMessage()' is ignored `; which I think is because of `@CheckReturnValue` annotation.

mohamedsamehsalah (approved):
LGTM 👍

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JUnitToAssertJRules.java
Stephan202: Yep, see also the Javadoc of `AbstractAssert#withFailMessage`:
```
* You must set the message <b>before</b> calling the assertion otherwise it is ignored as the failing assertion breaks
* the call chain by throwing an {@link AssertionError}.
```

mohamedsamehsalah: That makes more sense 😄

Stephan202:
Validated that these changes do not impact the integration tests.

rickie (approved):

PR #1488, opened 2024-12-29
Extend `ImmutableMapRules` Refaster rule collection

### LINKED ISSUE #1223: Rewrite `ImmutableMap.Builder#build` to `#buildOrThrow`
### Problem

An `ImmutableMap.Builder` has two terminal methods which accomplish the same thing: `build` or `buildOrThrow`.

The `buildOrThrow` method should be preferred, because the documentation of `build` specifies:

> Prefer the equivalent method {@link #buildOrThrow()} to make it explicit that the method will throw an exception if there are duplicate keys. The {@code build()} method will soon be deprecated.

    - Make the behavior evident: an exception may be thrown (in case of duplicate keys).
- [ ] Improve performance.

```java
ImmutableMap.Builder<K, V> builder = ImmutableMap.builder();
builder.build();
```

to:
```java
ImmutableMap.Builder<K, V> builder = ImmutableMap.builder();
builder.buildOrThrow();
```

Stephan202:
I set out to resolve #1223, then noticed a few other ways in which internal code can be simplified.

Suggested commit message:
```
Extend `ImmutableMapRules` Refaster rule collection (#1488)

Resolves #1223.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ImmutableMapRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ImmutableMapRules.java
@@ -373,14 +373,14 @@ final class ImmutableMapRules {
     abstract boolean keyFilter(@MayOptionallyUse K key);
 
     @BeforeTemplate
-    ImmutableMap<K, V> before(ImmutableMap<K, V> map) {
+    ImmutableMap<K, V> before(Map<K, V> map) {
       return map.entrySet().stream()
           .filter(e -> keyFilter(e.getKey()))
           .collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
     }
 
     @AfterTemplate
-    ImmutableMap<K, V> after(ImmutableMap<K, V> map) {
+    ImmutableMap<K, V> after(Map<K, V> map) {
       return ImmutableMap.copyOf(Maps.filterKeys(map, k -> keyFilter(k)));
     }
   }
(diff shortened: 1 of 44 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ImmutableMapRules.java
```diff
@@ -44,12 +45,28 @@ ImmutableMap.Builder<K, V> after() {
...
+    @BeforeTemplate
+    ImmutableMap<K, V> before(ImmutableMap.Builder<K, V> builder) {
+      return builder.build();
+    }
+
+    @AfterTemplate
+    ImmutableMap<K, V> after(ImmutableMap.Builder<K, V> builder) {
+      return builder.buildOrThrow();
+    }
+  }
```
Stephan202, line 62: The builder subtypes for `ImmutableBiMap` and `ImmutableSortedMap` extend from `ImmutableMap.Builder`, so those `build()` calls are also covered.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ImmutableMapRules.java
```diff
@@ -139,8 +157,6 @@ abstract static class StreamOfMapEntriesToImmutableMap<E, K, V> {
     @Placeholder(allowsIdentity = true)
     abstract V valueFunction(@MayOptionallyUse E element);
 
-    // XXX: We could add variants in which the entry is created some other way, but we have another
-    // rule that covers canonicalization to `Map.entry`.
```
Stephan202, line 143: Dropped, because this describes our approach more generally these days.

Stephan202 (commented):
Some context.

mohamedsamehsalah (approved):
LGTM 🚀 

Nice suggestion @EnricSala 🙌

integration-tests/prometheus-java-client-expected-changes.patch
```diff
@@ -4565,15 +4603,24 @@
...
 -class PrometheusMetricDataTest {
 +final class PrometheusMetricDataTest {
  
    Map<Object, Object> translations =
-       Map.ofEntries(
+-      Map.ofEntries(
++      ImmutableMap.ofEntries(
+           entry("days", "d"),
+           entry("hours", "h"),
+           entry("minutes", "min"),
```
Stephan202, line 4623: This isn't simplified further, because it's a _long_ list. We could add more rules, but: meh.

Stephan202 (commented):
Rebased and updated the integration test expectations.

rickie (approved):
✨

PR #1493, opened 2024-12-31
Introduce `FluxMapNotNull{,Transformation}OrElse` Refaster rules

werli:
I want to advocate for using `Flux#mapNotNull` more, even if that includes defaulting to an explicit `null` value.

I'm happy to learn how to fuse those into one rule. :eyes:

Suggested commit message:
```
Introduce `FluxMapNotNull{,Transformation}OrElse` Refaster rules (#1493)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -889,14 +889,17 @@ final class ReactorRules {
     }
   }
 
-  /** Prefer {@link Flux#mapNotNull(Function)} over more contrived alternatives. */
-  abstract static class FluxMapNotNullMapOptional<T, S> {
+  /**
+   * Prefer immediately unwrapping {@link Optional} transformation results inside {@link
+   * Flux#mapNotNull(Function)} over more contrived alternatives.
+   */
+  abstract static class FluxMapNotNullTransformationOrElse<T, S> {
     @Placeholder(allowsIdentity = true)
     abstract Optional<S> transformation(@MayOptionallyUse T value);
 
     @BeforeTemplate
     Flux<S> before(Flux<T> flux) {
-      return flux.map(x -> transformation(x)).mapNotNull(x -> x.orElse(null));
+      return flux.map(v -> transformation(v)).mapNotNull(o -> o.orElse(null));
     }
 
     @AfterTemplate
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -889,6 +889,35 @@ Flux<S> after(Flux<T> flux) {
...
   }
 
+  /** Prefer {@link Flux#mapNotNull(Function)} over more contrived alternatives. */
+  abstract static class FluxMapNotNullMapOptional<T, S> {
+    @Placeholder(allowsIdentity = true)
+    abstract Optional<S> transformation(@MayOptionallyUse T value);
+
+    @BeforeTemplate
+    Flux<S> before(Flux<T> flux) {
+      return flux.map(x -> transformation(x)).mapNotNull(x -> x.orElse(null));
```
Stephan202: It turns out that the test for this rule currently fails because `x` is reused. Official Java scoping rules have these variables be independent, but I suspect that Refaster takes a shortcut somewhere. Good that we have tests :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -889,6 +889,35 @@ Flux<S> after(Flux<T> flux) {
     }
   }
 
+  /** Prefer {@link Flux#mapNotNull(Function)} over more contrived alternatives. */
```
Stephan202: Both variants use `mapNotNull`; will propose something.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -889,6 +889,35 @@ Flux<S> after(Flux<T> flux) {
...
+  abstract static class FluxMapNotNullMapOptional<T, S> {
```
Stephan202: We're still figuring out our preferred naming strategy, but tentatively it'd be something like `FluxMapNotNullOrElse`. Since the rule below would yield the same name, perhaps we'd go for `FluxMapNotNullTransformationOrElse` here :thinking:. Will use that for now.

Stephan202 (approved):
Nice one @werli! I rebased and added a commit. Suggested commit message:
```
Introduce `FluxMapNotNull{,Transformation}OrElse` Refaster rules (#1493)
```

rickie (approved):

werli:
Thanks @Stephan202 & @rickie - LGTM 👍

PR #1547, opened 2025-02-14
Introduce `AssertJOptionalAssertion` check

JulianBroudy:
Refaster:
```
assertThat(optional.orElseThrow()).isEqualTo(value);
```
to:
```
assertThat(optional).hasValue(value);
```

## Suggested Commit Message :tada:

```
Introduce `AssertJOptionalAssertion` check (#1547)
```


rickie (commented):
Ahh nice that you opened the PR 🚀 ! 

We can probably add one case to the Refaster rule above [here](https://github.com/PicnicSupermarket/error-prone-support/pull/1547/files#diff-e596e760da2c2f3f9746053e72bcccc126159ee81e89f455cd71e47d44ee915fR95) `AbstractOptionalAssertHasValue`. 

The build will probably fail as we also need to add tests 😄. Try running `mvn clean install` to see whether the build passes 😉.

JulianBroudy:
> We can probably add one case to the Refaster rule above [here](https://github.com/PicnicSupermarket/error-prone-support/pull/1547/files#diff-e596e760da2c2f3f9746053e72bcccc126159ee81e89f455cd71e47d44ee915fR95) AbstractOptionalAssertHasValue.

Maybe we can go over it together? Because IIUC we cannot add the following:
image

If we stay with the new function, should it maybe also return `AbstractAssert<?, ?>` instead of `ObjectAssert<T>`?


> The build will probably fail as we also need to add tests 😄. Try running mvn clean install to see whether the build passes 😉.

On it!

Stephan202 (commented):
Congrats on your first contribution, @JulianBroudy! :tada: Too bad this is a tricky one. :grimacing:

I rebased and added a commit to show the issue. Normally for this kind of change we would use this commit message:
```
Introduce `AssertThatHasValue` Refaster rule (#1547)
```

But as this approach won't work, we'll have to consider an alternative: introducing an `AssertThatHasValue` Error Prone checker, just like we did for [`AssertJIsNull`](https://github.com/PicnicSupermarket/error-prone-support/blob/fed281dc2b819b2699a7739f2594ed08b6d638ed/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AssertJIsNull.java). This variant would be _slightly_ more complex, but not by much. Are you up for that? I'm sure @rickie is up for providing some pointers :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJOptionalRules.java
```diff
@@ -103,6 +104,19 @@ AbstractOptionalAssert<?, T> after(AbstractOptionalAssert<?, T> optionalAssert,
     }
   }
 
+  static final class AssertThatHasValue<T> {
+    @BeforeTemplate
+    AbstractStringAssert<?> before(Optional<String> optional, String value) {
```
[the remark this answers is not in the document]
JulianBroudy: Thanks @Stephan202, always up for anything!

We started with having it as generic and then went on to making it specific for each of String, Boolean, etc. Finally decided to stay only with String for now. I will take another look with @rickie.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AssertThatHasValue.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * Returns the {@code optional.orElseThrow()} invocation if the receiver of the given {@code
+   * isEqualTo} call is an {@code assertThat} call with exactly one argument that is a call to
+   * {@link Optional#orElseThrow()}.
+   */
+  private static Optional<MethodInvocationTree> getOrElseThrowInvocation(
+      MethodInvocationTree isEqualToCall, VisitorState state) {
+    ExpressionTree receiver = ASTHelpers.getReceiver(isEqualToCall);
```
rickie: These two checks are done as part of the Matcher :)

rickie (commented):
Already looked quite sharp sir! 

I was playing a bit with the matchers, but didn't arrive at a nicer setup for them 🤔. 
Also played a bit with the naming, made some proposals and slightly re-arranged some things. Added a commit with suggestions, feel free to tweak :)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/AssertThatHasValueTest.java
```diff
@@ -0,0 +1,85 @@
...
+    CompilationTestHelper.newInstance(AssertThatHasValue.class, getClass())
+        .addSourceLines(
+            "A.java",
+            "import static org.assertj.core.api.Assertions.assertThat;",
+            "",
+            "import java.util.Optional;",
+            "",
+            "class A {",
+            "  void m() {",
+            "    assertThat(Optional.of(1).get()).isEqualTo(1);",
```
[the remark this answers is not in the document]
rickie: On this line and the rest of the tests, we make sure to use different numbers on every line to ensure we are testing it correctly. And instead of strings like `test` we usually go for `foo` and `bar`, see other tests :)

rickie (commented):
some more context :)

rickie (commented):
Haha, I did some validation incorrectly and now had to revert something I thought was flagged by Pitest :thinking:. For now will revert and we can continue next week :). 
My bad 😬.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AssertThatHasValue.java
```diff
@@ -0,0 +1,108 @@
...
+    ExpressionTree optionalTree = methodSelect.getExpression();
+    return SuggestedFixes.renameMethodInvocation(
+            assertMethodTree, getReplacementMethod(assertMethodTree), state)
+        .toBuilder()
+        .replace(orElseThrowTree, SourceCode.treeToString(optionalTree, state))
+        .build();
+  }
+
+  private static String getReplacementMethod(MethodInvocationTree tree) {
+    String methodName = ASTHelpers.getSymbol(tree).getSimpleName().toString();
```
rickie: We don't need to cast to a string, let me suggest something :)

rickie (commented):
Very nice work @JulianBroudy ! Only pending is naming of the check :) 

Added a commit with some of the suggestions.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AssertThatHasValue.java
```diff
@@ -0,0 +1,108 @@
...
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.util.ASTHelpers;
+import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MemberSelectTree;
+import com.sun.source.tree.MethodInvocationTree;
+import java.util.Optional;
+import tech.picnic.errorprone.utils.SourceCode;
+
+/**
+ * A {@link BugChecker} that flags AssertJ {@code
```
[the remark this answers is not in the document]
JulianBroudy: No problem, was following this class `AssertJNullnessAssertion`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AssertJOptionalAssertion.java
```diff
@@ -0,0 +1,107 @@
...
+  private static final Matcher<MethodInvocationTree> ASSERT_METHOD =
+      allOf(
+          instanceMethod()
+              .onDescendantOf("org.assertj.core.api.Assert")
+              .namedAnyOf("isEqualTo", "isSameAs"),
+          argumentCount(1));
+  private static final Matcher<ExpressionTree> OPTIONAL_OR_ELSE_THROW =
+      instanceMethod()
+          .onExactClass(Optional.class.getCanonicalName())
+          .named("orElseThrow")
```
Stephan202: For Error Prone checks we're generally less strict in assuming that all Refaster rules have been applied, as it's likely that OSS users run some of these checks without also apply all Refaster rules. So let's also match `#get()` here.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AssertJOptionalAssertion.java
```diff
@@ -0,0 +1,107 @@
...
+ * A {@link BugChecker} that flags AssertJ usages of {@code OptionalAssert} for simplification.
+ *
+ * <p>This bug checker cannot be replaced with a simple Refaster rule, as the Refaster approach
+ * would require that all overloads of the mentioned methods (such as {@link
+ * org.assertj.core.api.AbstractStringAssert#isEqualTo(String)}) are explicitly enumerated. This bug
+ * checker generically matches all such current and future overloads.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Prefer `assertThat(optional).hasValue(value)` over more verbose alternatives",
```
Stephan202: I suppose we should also mention `containsSame`; I'll use the same pattern as is used by `AssertJNullnessAssertion`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AssertJOptionalAssertion.java
```diff
@@ -0,0 +1,107 @@
...
+      allOf(
+          instanceMethod()
+              .onDescendantOf("org.assertj.core.api.Assert")
+              .namedAnyOf("isEqualTo", "isSameAs"),
+          argumentCount(1));
+  private static final Matcher<ExpressionTree> OPTIONAL_OR_ELSE_THROW =
+      instanceMethod()
+          .onExactClass(Optional.class.getCanonicalName())
+          .named("orElseThrow")
+          .withNoParameters();
```
Stephan202: As we'll only be modifying test code, I think it's okay to also suggest simplifying `.orElseThrow` with a custom exception message. I can't think of a good use case for that, but if it exists, users can suppress the warning.

PR #1553, opened 2025-02-18
Extend `MonoSingleOptional` Refaster rule

werli:
Another bite-sized Refaster rule extension.

Suggested commit message:
```
Extend `MonoSingleOptional` Refaster rule (#1553)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -1082,11 +1082,12 @@ final class ReactorRules {
   // rule. Consider introducing an Error Prone check for this.
   static final class MonoSingleOptional<T> {
     @BeforeTemplate
-    Mono<Optional<T>> before(Mono<T> mono) {
+    Mono<Optional<T>> before(Mono<T> mono, Optional<T> optional, Mono<Optional<T>> alternate) {
       return Refaster.anyOf(
           mono.flux().collect(toOptional()),
           mono.map(Optional::of),
-          mono.singleOptional().defaultIfEmpty(Optional.empty()),
+          mono.singleOptional().defaultIfEmpty(optional),
+          mono.singleOptional().switchIfEmpty(alternate),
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -1081,7 +1081,8 @@ static final class MonoSingleOptional<T> {
     Mono<Optional<T>> before(Mono<T> mono) {
       return Refaster.anyOf(
           mono.flux().collect(toOptional()),
-          mono.map(Optional::of).defaultIfEmpty(Optional.empty()),
+          mono.map(Optional::of),
+          mono.singleOptional().defaultIfEmpty(Optional.empty()),
```
werli, line 1085: `Mono#singleOptional` will default to `Optional.empty()` anyways, so keeping that in mind we can break this rule into two:
1. Re-write `mono.map(Optional::of)`
2. Omit the unnecessary `.defaultIfEmpty(Optional.empty()` when possible.

The existing case would still be re-written by applying to (1) and then (2).

Stephan202: Looks like we can even generalize this to any `defaultIfEmpty` and `switchIfEmpty` argument :)

Stephan202 (approved):
Rebased and added a commit. Tnx!

Stephan202:
Build failure is due to an interesting case of non-determinism. Will tweak test case.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
werli: :bulb: Since the `Mono` will never be empty. Nice, I didn't even consider this.

Stephan202: This is why we have code review :smile: `:team-work:`

rickie (approved):

PR #1556, opened 2025-02-20
Extend `MonoThen` Refaster rule

mohamedsamehsalah:
* `Mono#when` aggregates multiple publishers into one `Mono<Void>`, passing only one publisher to the argument is redundant and can be re-fastered into a `Mono#then`. 

Suggested commit message
```
Extend `MonoThen` Refaster rule (#1556)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -939,7 +939,11 @@ final class ReactorRules {
   static final class MonoThen<T> {
     @BeforeTemplate
     Mono<@Nullable Void> before(Mono<T> mono) {
-      return Refaster.anyOf(mono.ignoreElement().then(), mono.flux().then(), Mono.when(mono));
+      return Refaster.anyOf(
+          mono.ignoreElement().then(),
+          mono.flux().then(),
+          Mono.when(mono),
+          Mono.whenDelayError(mono));
     }
 
     @AfterTemplate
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -939,7 +939,7 @@ Flux<T> after(Mono<T> mono) {
   static final class MonoThen<T> {
     @BeforeTemplate
     Mono<@Nullable Void> before(Mono<T> mono) {
-      return Refaster.anyOf(mono.ignoreElement().then(), mono.flux().then());
+      return Refaster.anyOf(mono.ignoreElement().then(), mono.flux().then(), Mono.when(mono));
```
mohamedsamehsalah, line 942: `Mono#when` aggregates multiple publishers into one `Mono<Void>`, passing only one publisher to the argument is redundant.

rickie (approved):
Suggested commit message:
```
Extend `MonoThen` Refaster rule (#1556)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: We can also add `Mono.whenDelayError(mono)`.

Stephan202 (approved):
Rebased and added a commit. Nice one; tnx @mohamedsamehsalah!

@rickie feel free to merge if you agree with the latest change.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -345,7 +345,8 @@ ImmutableSet<Mono<Void>> testMonoThen() {
     return ImmutableSet.of(
         Mono.just("foo").ignoreElement().then(),
         Mono.just("bar").flux().then(),
-        Mono.when(Mono.just("baz")));
+        Mono.when(Mono.just("baz")),
+        Mono.when(Mono.just("qux")));
```
mohamedsamehsalah, line 349:
```suggestion
        Mono.whenDelayError(Mono.just("qux")));
```

Stephan202: Doh; good catch!

PR #1581, opened 2025-03-10
Extend `FluxJust` Refaster rule

JulianBroudy:
COM-1201

```java
Extend `FluxJust` Refaster rule (#1581)
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -489,9 +489,13 @@ final class ReactorRules {
       return Flux.range(value, 1);
     }
 
+    // XXX: Consider generalizing part of this template using an Error Prone check that covers any
+    // sequence of explicitly enumerated values passed to an iteration order-preserving collection
+    // factory method.
     @BeforeTemplate
     Flux<T> before(T value) {
       return Refaster.anyOf(
+          Mono.just(value).flux(),
           Mono.just(value).repeat().take(1),
           Flux.fromIterable(ImmutableList.of(value)),
           Flux.fromIterable(ImmutableSet.of(value)));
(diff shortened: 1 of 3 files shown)


rickie (approved):
Nice contribution @JulianBroudy 🚀

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -491,7 +491,10 @@ Flux<Integer> before(int value) {
 
     @BeforeTemplate
     Flux<T> before(T value) {
-      return Mono.just(value).repeat().take(1);
+      return Refaster.anyOf(
+          Mono.just(value).repeat().take(1),
+          Flux.fromIterable(ImmutableList.of(value)),
```
werli, line 496: Good for thought: Couldn't we also include varargs for the `ImmutableList` case? Essentially avoiding every unnecessary list creation.

rickie: Yes; good one! @Stephan202 mentioned this could extend to a `Matcher` that checks for singleton lists, and potentially even more as you suggest :D

Stephan202 (approved):
I rebased and added a commit, covering one more case.

Due to the many `Immutable{List,Set}#of` overloads, it's not easy/useful to support the varargs variant using Refaster. While offline indeed I suggested introducing a `Matcher`, [this code](https://github.com/PicnicSupermarket/error-prone-support/blob/40b3c87b72d82c99dbee185da1cb7e7a723c6690/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StreamRules.java#L707-L775) and [this class](https://github.com/PicnicSupermarket/error-prone-support/blob/40b3c87b72d82c99dbee185da1cb7e7a723c6690/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/MicrometerRules.java) shows that it'd be better to introduce an Error Prone rule. But in the interest of merging this PR, I decided to add only a comment describing that possible future.

@rickie if you agree with my changes; feel free to merge.

PR #1596, opened 2025-03-18
Introduce two `FileRules` Refaster rules

timtebeek:
Proposing a quick rule for something that's come up repeatedly on our end. Figured it'd be a good match for a Refaster rule.

The goal here is to avoid concurreny issues when two processes are simultaneously checking if a directory exists, before attempting to create that directory, as part of a single check. By flipping around these two commands you avoid conflicts.

I didn't spot any existing tests for any `FileRule`; let me know if I missed any. If so I'd propose to also verify the negation of `!Files.exists(path) && !path.toFile().mkdirs()`.

- As discovered on https://github.com/openrewrite/rewrite/pull/5189

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
@@ -156,4 +156,18 @@ final class FileRules {
       return path.toFile().mkdirs() || Files.exists(path);
     }
   }
+
+  /** Prefer {@link File#mkdirs} before {@link File#exists} to avoid concurrency issues. */
+  static final class FileMkDirsFileExists {
+    @BeforeTemplate
+    boolean before(File file) {
+      return file.exists() || file.mkdirs();
+    }
+
+    @AfterTemplate
+    @AlsoNegation
+    boolean after(File file) {
+      return file.mkdirs() || file.exists();
+    }
+  }
 }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
@@ -143,9 +143,7 @@ final class FileRules {
     }
   }
 
-  /**
-   * Prefer `File#mkdirs` before `Files#exists` to avoid concurrency issues.
-   */
+  /** Prefer `File#mkdirs` before `Files#exists` to avoid concurrency issues. */
   static final class MkdirsBeforeFilesExists {
     @BeforeTemplate
     boolean before(Path path) {
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
@@ -143,8 +143,8 @@ final class FileRules {
     }
   }
 
-  /** Prefer `File#mkdirs` before `Files#exists` to avoid concurrency issues. */
-  static final class MkdirsBeforeFilesExists {
+  /** Prefer {@link File#mkdirs} before {@link Files#exists} to avoid concurrency issues. */
+  static final class PathToFileMkDirsFilesExists {
     @BeforeTemplate
     boolean before(Path path) {
       return Files.exists(path) || path.toFile().mkdirs();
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
```diff
@@ -141,4 +142,18 @@ File after(File directory, String prefix, String suffix) throws IOException {
       return Files.createTempFile(directory.toPath(), prefix, suffix).toFile();
     }
   }
+
+  /** Prefer `File#mkdirs` before `Files#exists` to avoid concurrency issues. */
+  static final class MkdirsBeforeFilesExists {
+    @BeforeTemplate
+    boolean before(Path path) {
```
timtebeek, line 149: I specifically cover `Path` here, but the same could apply for `File` as well. Let me know if that should be added.

kisssandoradam (approved):

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
```diff
@@ -141,4 +142,18 @@ File after(File directory, String prefix, String suffix) throws IOException {
       return Files.createTempFile(directory.toPath(), prefix, suffix).toFile();
     }
   }
+
+  /** Prefer `File#mkdirs` before `Files#exists` to avoid concurrency issues. */
```
rickie: Let's use `@link` :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
```diff
@@ -141,4 +142,18 @@ File after(File directory, String prefix, String suffix) throws IOException {
...
+  static final class MkdirsBeforeFilesExists {
```
rickie: It's not very obvious, but we use a "naming algorithm" for after templates that we at some point want to automate. It's based on the content of the after template :), will push a suggestion.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/FileRulesTestInput.java
```diff
@@ -39,4 +39,9 @@ ImmutableSet<File> testFilesCreateTempFileToFile() throws IOException {
   File testFilesCreateTempFileInCustomDirectoryToFile() throws IOException {
     return File.createTempFile("foo", "bar", new File("baz"));
   }
+
+  boolean testMkdirsBeforeFilesExists() {
+    Path foo = Path.of("foo");
```
rickie: We generally don't create local variables in tests. I can see how in this case it _would_ make sense. However, I see that it still works if we inline it and create a `Path` twice. Now I'm not 100% what we prefer, as officially it should be the same variable. 

One of the other similar case is where we have e.g. `CheckElementIndexWithMessage`. Will go with the approach we took there and therefore inline the variable.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/FileRulesTestOutput.java
```diff
@@ -39,4 +39,9 @@ ImmutableSet<File> testFilesCreateTempFileToFile() throws IOException {
   File testFilesCreateTempFileInCustomDirectoryToFile() throws IOException {
     return Files.createTempFile(new File("baz").toPath(), "foo", "bar").toFile();
   }
+
+  boolean testMkdirsBeforeFilesExists() {
+    Path foo = Path.of("foo");
+    return !foo.toFile().mkdirs() && !Files.exists(foo);
```
rickie: Will push a test for the negation as well (as you pointed out 😄 )

rickie (approved):
Only consideration now is the Javadocs that only differ by 1 letter 😋.

Thanks for the contribution @timtebeek 🚀.

We're going to create a release in the coming days 😄. 

I pushed a few suggestions PTAL.

Stephan202 (approved):
Nice suggested @timtebeek! I rebased and added a commit.

Suggested commit message:
```
Introduce two `FileRules` Refaster rules (#1596)

Invoking `File#mkdirs` before testing for existence of the specified
path, rather than the other way around, avoids a subtle concurrency
issue.

See also openrewrite/rewrite#5189.
```

@rickie if you're okay with the changes, feel free to merge :)

rickie:
> Nice suggested @timtebeek! I rebased and added a commit.

I don't see the commit, right? @Stephan202

Stephan202:
:see_no_evil:

PR #1609, opened 2025-03-22
Upgrade JDKs used by GitHub Actions builds

Stephan202:
:exclamation: This PR is on top of #1686. :exclamation:

Suggested commit message:
```
Upgrade JDKs used by GitHub Actions builds (#1609)

Summary of changes:
- Use JDK 17.0.15 instead of 17.0.13.
- Use JDK 21.0.7 instead of 21.0.5.
- Use JDK 24.0.1 instead of 23.0.1.
- Have GitHub issue template reference more recent version numbers.

See:
- https://adoptium.net/temurin/release-notes/?version=jdk-17.0.14+7
- https://adoptium.net/temurin/release-notes/?version=jdk-17.0.15+6
- https://adoptium.net/temurin/release-notes/?version=jdk-21.0.6+7
- https://adoptium.net/temurin/release-notes/?version=jdk-21.0.7+6
- https://adoptium.net/temurin/release-notes/?version=jdk-24+36
- https://adoptium.net/temurin/release-notes/?version=jdk-24.0.1+9
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -42,9 +42,9 @@ Please replace this sentence with log output, if applicable.
 <!-- Please complete the following information: -->
 
 - Operating system (e.g. MacOS Monterey).
-- Java version (i.e. `java --version`, e.g. `17.0.13`).
-- Error Prone version (e.g. `2.35.1`).
-- Error Prone Support version (e.g. `0.19.0`).
+- Java version (i.e. `java --version`, e.g. `17.0.15`).
+- Error Prone version (e.g. `2.38.0`).
+- Error Prone Support version (e.g. `0.23.0`).
 
 ### Additional context
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,12 +10,12 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-24.04, macos-15, windows-2025 ]
-        build-jdk: [ 21.0.5 ]
-        target-jdk: [ 17.0.13 ]
+        build-jdk: [ 21.0.7 ]
+        target-jdk: [ 17.0.15 ]
         include:
           - os: ubuntu-24.04
-            build-jdk: 23.0.1
-            target-jdk: 17.0.13
+            build-jdk: 24.0.1
+            target-jdk: 17.0.15
     runs-on: ${{ matrix.os }}
     steps:
       - name: Install Harden-Runner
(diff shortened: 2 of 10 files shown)


Stephan202:
The build isn't yet JDK 24-compatible, IIUC because [OpenRewrite uses Lombok](https://github.com/openrewrite/rewrite/blob/fca7f30864043417a6d0811be886d34b4d33b2a6/rewrite-java-17/src/main/java/org/openrewrite/java/isolated/ReloadableJava17Parser.java#L119-L129), while the used version of Lombok isn't compatible with JDK 24: projectlombok/lombok#3772.

CC @timtebeek.

NB: The referenced OpenRewrite code seems to treat Lombok as an optional dependency, but as we don't otherwise depend on Lombok, should this code path be taken at all? :thinking:

timtebeek:
It looks like the folks at Lombok are working on a release right now, based the comments here:
- https://github.com/projectlombok/lombok/issues/3847#issuecomment-2763341555
- https://github.com/projectlombok/lombok/commit/f158f2c24151181eaf5cb1a1ff4d76c822ebb2ac
- https://projectlombok.org/download-edge

Once that's available from Maven Central I think it should be easy for us to put out a new patch release using that particular version, if needed.

Stephan202:
The build now fails due to openrewrite/rewrite-templating#118 / openrewrite/rewrite#4768; the top-level issue to track is openrewrite/rewrite#5211.

Stephan202:
Bumped JDK versions; new release notes not yet available.

timtebeek:
Java 24 support has been merged on our side and will go out early next week in a new release.

timtebeek:
Seeing:
```
org.openrewrite.java.isolated.ReloadableJava17ParserVisitor.convert(ReloadableJava17ParserVisitor.java:1765)
	... 76 more
Caused by: java.lang.ClassNotFoundException: com.sun.tools.javac.code.Type$UnknownType
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:349)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:557)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:490)
```
We've fixed that in rewrite-java-21(!) a while ago. Should be a drop in replacement here, or addition there and then we ought to pick the right one as needed. 
https://github.com/PicnicSupermarket/error-prone-support/blob/7b888f039c823efcfd26fc506a3ee9c598a3aa85/error-prone-contrib/pom.xml#L234

Hope that helps! Only recalled that we had pinned this just now. We had back ported another fix to 17 earlier, but for this one it only makes sense in the 21 parser, or the 25 parser to come still.

Stephan202:
The situation is unchanged from [this comment](https://github.com/PicnicSupermarket/error-prone-support/pull/1557#issuecomment-2677775806). That said, I sat down to look into the Maven Toolchain route, by reusing some code that was written and dropped again in the context of #603. The result is #1686.

Stephan202:
I rebased on top of #1686. Build now passes. :muscle:

mohamedsamehsalah (approved):
📈

rickie (approved):

PR #1621, opened 2025-03-29
Introduce `SpringTestRules` Refaster rule collection

Stephan202:
Suggested commit message:
```
Introduce `SpringTestRules` Refaster rule collection
```

Opening this as a draft PR, because I can't get the build to pass: despite the `@SuppressWarnings("deprecation")` annotations, warnings are emitted, in duplicate. I _suspect_ that there's an interesting interaction with the annotation processors/compiler plugins, but that's TBD.


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/SpringTestRules.java
```diff
@@ -0,0 +1,47 @@
...
+    @BeforeTemplate
+    @SuppressWarnings("deprecation" /* This deprecated method invocation will be rewritten. */)
+    BodyContentSpec before(BodyContentSpec spec, String expectedJson) {
+      return Refaster.anyOf(spec.json(expectedJson), spec.json(expectedJson, /* strict= */ false));
+    }
+
+    @AfterTemplate
+    BodyContentSpec after(BodyContentSpec spec, String expectedJson) {
+      return spec.json(expectedJson, JsonCompareMode.LENIENT);
+    }
```
Badbond, line 29: Alternative to write it to the overload that does this by default:

```suggestion
    BodyContentSpec after(BodyContentSpec spec, String expectedJson) {
      return spec.json(expectedJson);
    }
```

Stephan202:
I rebased, and the build now passes. @Badbond if you agree with the argument above, feel free to approve ;)

rickie (approved):

PR #1645, opened 2025-04-18
Share same generic parameters with the templates

fmodesto:
**Summary**
Refactor a few Refaster recipes by moving type variable declarations from the method level to the class level. This ensures consistency between before() and after() templates.

**Motivation**
OpenRewrite doesn’t support method-level generic type variables, which can result in incorrect or uncompilable replacements.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TestNGToAssertJRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TestNGToAssertJRules.java
@@ -904,30 +904,30 @@ final class TestNGToAssertJRules {
     }
   }
 
-  static final class AssertEqualIteratorIterationOrder {
+  static final class AssertEqualIteratorIterationOrder<S, T extends S> {
     @BeforeTemplate
-    void before(Iterator<?> actual, Iterator<?> expected) {
+    void before(Iterator<S> actual, Iterator<T> expected) {
       assertEquals(actual, expected);
     }
 
     @AfterTemplate
     @UseImportPolicy(STATIC_IMPORT_ALWAYS)
-    <S, T extends S> void after(Iterator<S> actual, Iterator<T> expected) {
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/TestNGToAssertJRulesTestOutput.java
```diff
@@ -245,6 +245,9 @@ void testAssertEqualIteratorIterationOrder() {
         .toIterable()
         .containsExactlyElementsOf(
             copyOf(Iterators.unmodifiableIterator(new ArrayList<>().iterator())));
+    assertEquals(
```
mohamedsamehsalah, line 248: I'd say we can omit the negative match.

mohamedsamehsalah (commented):
Thanks 💪

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TestNGToAssertJRules.java
```diff
@@ -939,64 +939,64 @@ <S, T extends S> void after(Iterator<S> actual, String message, Iterator<T> expe
 
   // XXX This rule fails for `java.nio.file.Path` as it is `Iterable`, but AssertJ's
   // `assertThat(Path)` does not support `.containsExactlyElementsOf`.
```
Stephan202, line 945: This comment also applies to the rule below. Will copy it over.

Stephan202 (approved):
Rebased and added a commit. Thanks for this contribution @fmodesto!

Suggested commit message:
```
Improve `TestNGToAssertJRules` generic type constraints (#1645)
```

mohamedsamehsalah (approved):
Thanks for the changes @fmodesto ! 🚀 

Suggestions LGTM @Stephan202 🙏

Stephan202:
Ah, I should have pressed the button to approve running of the workflows :facepalm:

PR #1648, opened 2025-04-19
Drop IDEA-342187 workaround

Stephan202:
Suggested commit message:
```
Drop IDEA-342187 workaround (#1648)

As of IntelliJ IDEA 2024.3.2.2, the IDE gracefully handles the absence
of explicitly specified annotation processor path dependency versions,
consistent with `maven-compiler-plugin`'s dependency resolution.
```


mohamedsamehsalah (approved):
🧹

rickie (approved):
Very nice that we can drop this 🚀 !

PR #1650, opened 2025-04-19
Drop unmaintained `reactor-error-prone` dependency

Stephan202:
Suggested commit message:
```
Drop unmaintained `reactor-error-prone` dependency (#1650)

This dependency is incompatible with Error Prone 2.38.0. By dropping it, 
the `non-maven-central` Maven profile and Jitpack mirror can also be
dropped.
```


rickie (approved):

mohamedsamehsalah (approved):
✅

Issue #1646, opened 2025-04-19
Add Error Prone 2.38.0 support


ben-manes:
> java.lang.NoSuchMethodError: 'com.google.errorprone.util.ASTHelpers$TargetType com.google.errorprone.util.ASTHelpers.targetType(com.google.errorprone.VisitorState)'
at tech.picnic.errorprone.bugpatterns.IdentityConversion.matchMethodInvocation(IdentityConversion.java:109)

The temporary workaround is to disable `IdentityConversion`. You might consider having a compatibility build using the errorprone repository, [similar](https://github.com/uber/NullAway/issues/1110) to how NullAway tests its latest against Caffeine's latest.

<details>

```console
* What went wrong:
Execution failed for task ':caffeine:compileJavaPoetJava'.
> Compilation failed; see the compiler output below.
  error: An unhandled exception was thrown by the Error Prone static analysis plugin.
       Please report this at https://github.com/google/error-prone/issues/new and include the following:

       error-prone version: 2.38.0
       BugPattern: (see stack trace)
       Stack Trace:
       java.lang.NoSuchMethodError: 'com.google.errorprone.util.ASTHelpers$TargetType com.google.errorprone.util.ASTHelpers.targetType(com.google.errorprone.VisitorState)'
        at tech.picnic.errorprone.bugpatterns.IdentityConversion.matchMethodInvocation(IdentityConversion.java:109)
        at com.google.errorprone.scanner.ErrorProneScanner.processMatchers(ErrorProneScanner.java:509)
        at com.google.errorprone.scanner.ErrorProneScanner.visitMethodInvocation(ErrorProneScanner.java:824)
        at com.google.errorprone.scanner.ErrorProneScanner.visitMethodInvocation(ErrorProneScanner.java:170)
        at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCMethodInvocation.accept(JCTree.java:1877)
        at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:92)
        at com.google.errorprone.scanner.Scanner.scan(Scanner.java:74)
        at com.google.errorprone.scanner.Scanner.scan(Scanner.java:48)
        at jdk.compiler/com.sun.source.util.TreeScanner.visitReturn(TreeScanner.java:546)
        at com.google.errorprone.scanner.ErrorProneScanner.visitReturn(ErrorProneScanner.java:931)
[...]

PR #1649, opened 2025-04-19
Upgrade Error Prone 2.37.0 -> 2.38.0

Stephan202:
Renovate didn't open a PR, so did it manually.

Suggested commit message:
```
Upgrade Error Prone 2.37.0 -> 2.38.0 (#1649)

Resolves #1646.

See:
- https://github.com/google/error-prone/releases/tag/v2.38.0
- https://github.com/google/error-prone/compare/v2.37.0...v2.38.0
- https://github.com/PicnicSupermarket/error-prone/compare/v2.37.0-picnic-1...v2.38.0-picnic-1
```

PR #1651, opened 2025-04-19
Validate compatibility with latest Error Prone release

Stephan202:
See discussion in #1646. I opted not to bother with testing against a "locally" built HEAD release of Error Prone, as that would at most speed up incompatibility identification by 24 hours, while potentially causing weeks/months of build failures between the time an incompatibility is introduced and a subsequent release.

Suggested commit message:
```
Validate compatibility with latest Error Prone release (#1651)

Every day and after each merged change we now run a GitHub Actions
workflow that builds the project against the latest Error Prone release,
as determined by Maven. This should pro-actively inform us of
incompatibilities that demand a new release.

While there:
- Drop a redundant permission declaration from the SonarCloud workflow.
- Update `run-full-build.sh` to parallelize the build.
```

### DIFF
--- a/run-full-build.sh
+++ b/run-full-build.sh
@@ -11,9 +11,11 @@ set -e -u -o pipefail
 settings="$(dirname "${0}")/settings.xml"
 
 mvn clean install \
+  -T1C \
   -s "${settings}" \
   $@
 mvn clean install \
+  -T1C \
   -s "${settings}" \
   -Perror-prone-fork \
   -Pself-check \
--- a/.github/workflows/error-prone-compat.yml
+++ b/.github/workflows/error-prone-compat.yml
@@ -1,0 +1,36 @@
+# Validates compatibility with the latest Error Prone release.
+#
+# XXX: This workflow compiles the code against the `LATEST` Error Prone release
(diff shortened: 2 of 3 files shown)


Stephan202:
New check fails as expected. Build should pass after #1649.

rickie (commented):
Didn't review in depth, but very nice set up @Stephan202 !

Stephan202:
Rebased and added a commit to track a [nice suggestion](https://github.com/PicnicSupermarket/error-prone-support/issues/1646#issuecomment-2816824543) by @ben-manes.

The build now passes, so dropped the test logic. That means the PR is ready for review.

run-full-build.sh
```diff
@@ -11,9 +11,11 @@ set -e -u -o pipefail
 settings="$(dirname "${0}")/settings.xml"
 
 mvn clean install \
+  -T1C \
```
mohamedsamehsalah, line 14: ⚙️

mohamedsamehsalah (approved):
Nice 🌞

rickie (approved):

PR #1664, opened 2025-05-03
Extend `AssertJMapRules` Refaster rule collection

timtebeek:
While working on https://github.com/openrewrite/rewrite-testing-frameworks/pull/713 I discovered some patterns that weren't included as rules here, and figured you all would appreciate having these as well.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
@@ -12,7 +12,6 @@ import com.google.errorprone.refaster.annotation.UseImportPolicy;
 import java.util.Collection;
 import java.util.Map;
 import org.assertj.core.api.AbstractAssert;
-import org.assertj.core.api.AbstractBooleanAssert;
 import org.assertj.core.api.AbstractCollectionAssert;
 import org.assertj.core.api.AbstractMapAssert;
 import org.assertj.core.api.MapAssert;
@@ -154,13 +153,9 @@ final class AssertJMapRules {
 
   static final class AssertThatMapContainsKey<K, V> {
     @BeforeTemplate
-    AbstractBooleanAssert<?> before(Map<K, V> map, K key) {
-      return assertThat(map.containsKey(key)).isTrue();
-    }
-
-    @BeforeTemplate
-    AbstractCollectionAssert<?, Collection<? extends K>, K, ?> before2(Map<K, V> map, K key) {
-      return assertThat(map.keySet()).contains(key);
+    AbstractAssert<?, ?> before(Map<K, V> map, K key) {
+      return Refaster.anyOf(
+          assertThat(map.containsKey(key)).isTrue(), assertThat(map.keySet()).contains(key));
     }
 
     @AfterTemplate
@@ -172,13 +167,9 @@ final class AssertJMapRules {
 
   static final class AssertThatMapDoesNotContainKey<K, V> {
     @BeforeTemplate
-    AbstractBooleanAssert<?> before(Map<K, V> map, K key) {
-      return assertThat(map.containsKey(key)).isFalse();
-    }
-
-    @BeforeTemplate
-    AbstractCollectionAssert<?, Collection<? extends K>, K, ?> before2(Map<K, V> map, K key) {
-      return assertThat(map.keySet()).doesNotContain(key);
+    AbstractAssert<?, ?> before(Map<K, V> map, K key) {
+      return Refaster.anyOf(
+          assertThat(map.containsKey(key)).isFalse(), assertThat(map.keySet()).doesNotContain(key));
     }
 
     @AfterTemplate
@@ -188,10 +179,13 @@ final class AssertJMapRules {
     }
   }
 
+  // XXX: Strictly speaking this rule could be merged into `AssertThatMapContainsOnlyKeys` below,
+  // but that rule targets another `containsOnlyKeys` overload. Review how cases like this should
+  // impact the preferred naming scheme.
   static final class AssertThatMapContainsOnlyKey<K, V> {
     @BeforeTemplate
-    AbstractCollectionAssert<?, Collection<? extends K>, K, ?> before2(Map<K, V> map, K key) {
-      return assertThat(map.keySet()).containsOnly(key);
+    AbstractCollectionAssert<?, Collection<? extends K>, K, ?> before(Map<K, V> map, K key) {
+      return assertThat(map.keySet()).containsExactly(key);
     }
 
     @AfterTemplate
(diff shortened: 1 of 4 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJRules.java
```diff
@@ -311,7 +311,7 @@ static final class AssertThatMultisetsAreEqual<S, T extends S> {
   // `assertThat` overload. Consider defining a `BugChecker` instead.
   static final class AssertThatMapContainsEntry<K, V> {
     @BeforeTemplate
-    ObjectAssert<?> before(Map<K, V> map, K key, V value) {
+    ObjectAssert<V> before(Map<K, V> map, K key, V value) {
```
timtebeek, line 314: I'd noticed this one didn't match on our end; wasn't sure if this change factored in, but didn't see a reason to have a wildcard here. Also wasn't sure why this test isn't in `AssertJMapRules`, as it seemed to better fit in there.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
```diff
@@ -171,13 +176,31 @@ AbstractBooleanAssert<?> before(Map<K, V> map, K key) {
...
+    }
+
     @AfterTemplate
     @UseImportPolicy(STATIC_IMPORT_ALWAYS)
     MapAssert<K, V> after(Map<K, V> map, K key) {
       return assertThat(map).doesNotContainKey(key);
     }
   }
 
+  static final class AssertThatMapContainsOnlyKey<K, V> {
```
timtebeek, line 191: As compared to the `+ "s"` variant just below this matches a single element, as opposed to an `Iterable`. The class name had to be unique, even though they have the same target method in `after`, which dictates the class name. Due to differing arguments to the templates this had to be a separate class.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
```diff
@@ -171,13 +176,31 @@ AbstractBooleanAssert<?> before(Map<K, V> map, K key) {
...
+    @BeforeTemplate
+    AbstractCollectionAssert<?, Collection<? extends K>, K, ?> before2(Map<K, V> map, K key) {
+      return assertThat(map.keySet()).containsOnly(key);
```
timtebeek, line 194: Interestingly this has overlap with the rule we have for sets, but the AfterTemplate here is more specific for Maps. 🤔 
https://github.com/PicnicSupermarket/error-prone-support/blob/bce570f2d0626bade3b09ca764889512c543a3f8/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJRules.java#L240-L251

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJRules.java
Stephan202: The change is fine :+1:. This rule hasn't been moved to `AssertJMapRules` yet because in practice it will almost never match, for the reason outlined in the `XXX` comment.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
Stephan202: Refaster does support incompatible parameter types in a way that would allow us to merge these rules, but it's awkward at best. I'll just add a comment to track this topic for another time.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
```diff
@@ -171,13 +176,31 @@ AbstractBooleanAssert<?> before(Map<K, V> map, K key) {
...
     @AfterTemplate
     @UseImportPolicy(STATIC_IMPORT_ALWAYS)
     MapAssert<K, V> after(Map<K, V> map, K key) {
       return assertThat(map).doesNotContainKey(key);
     }
   }
 
+  static final class AssertThatMapContainsOnlyKey<K, V> {
+    @BeforeTemplate
+    AbstractCollectionAssert<?, Collection<? extends K>, K, ?> before2(Map<K, V> map, K key) {
```
Stephan202:
```suggestion
    AbstractCollectionAssert<?, Collection<? extends K>, K, ?> before(Map<K, V> map, K key) {
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
```diff
@@ -158,6 +158,11 @@ AbstractBooleanAssert<?> before(Map<K, V> map, K key) {
       return assertThat(map.containsKey(key)).isTrue();
     }
 
+    @BeforeTemplate
+    AbstractCollectionAssert<?, Collection<? extends K>, K, ?> before2(Map<K, V> map, K key) {
+      return assertThat(map.keySet()).contains(key);
+    }
```
Stephan202: We generally prefer `Refaster.anyOf` in cases like these. (That does hide the different return types, so might be something we change our mind on later, but likely only if/when we have an Error Prone check to suggest one or the other format as appropriate.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJMapRules.java
Stephan202: Indeed: our self-check flags that in this case the new check should match against `containsExactly`, as existing code is expected to have already been rewritten by the `AssertThatSetContainsExactlyOneElement` rule.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/AssertJMapRulesTestInput.java
```diff
@@ -111,6 +111,10 @@ MapAssert<Integer, Integer> testAbstractMapAssertHasSameSizeAs() {
     return assertThat(ImmutableMap.of(1, 2).containsKey(3)).isFalse();
   }
 
+  AbstractAssert<?, ?> testAssertThatMapContainsOnlyKey() {
+    return assertThat(ImmutableMap.of(1, 2).keySet()).containsOnly(3);
+  }
```
Stephan202: I'll also add test cases for the other new templates.

Stephan202 (approved):
I rebased and added a commit. Thanks @timtebeek!

Suggested commit message:
```
Extend `AssertJMapRules` Refaster rule collection (#1664)

By adding new templates to existing rules, and introducing the
`AssertThatMapContainsOnlyKey` rule.
```

rickie (approved):
Awesome! Thanks for the PR @timtebeek 🚀

timtebeek:
Thanks both for the quick review & fixes; look forward to pulling these in on our end! :)

PR #1669, opened 2025-05-07
Fix integration test setup

Stephan202:
:exclamation: Draft PR. Would like to check whether we can also fix the Checkstyle tests. :exclamation:

Suggested commit message:
```
Fix integration test setup (#1669)

Summary of changes:
- Fix the `checkstyle` build by suppressing Java 11-incompatible patch
  suggestions.
- Fix `prometheus-java-client` integration tests by allowing Docker
  containers to start on GitHub actions.
```

### DIFF
--- a/.github/workflows/run-integration-tests.yml
+++ b/.github/workflows/run-integration-tests.yml
@@ -24,7 +24,10 @@ jobs:
       - name: Install Harden-Runner
         uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
         with:
-          disable-sudo-and-containers: true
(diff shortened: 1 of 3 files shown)


Stephan202:
This PR is ready for review. [Build output here](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/14943653105?pr=1669). The `prometheus-java-client` build still fails there, because it takes the GHA config from `master`. I just added a temporary commit to validate that the change on this branch works.

Stephan202:
Works! Commit reverted.

rickie (approved):
What about always running this when we merge to master? 🤔 

The earlier we detect it fails the better.

PR #1673, opened 2025-05-10
Extend `AssertJEnumerableRules` Refaster rule collection

Stephan202:
Suggested commit message:
```
Extend `AssertJEnumerableRules` Refaster rule collection

By porting and generalizing the `AssertThatObjectEnumerableIsEmpty` rule
from the `AssertJRules` rule collection.

While there, update a partially-outdated comment in `AssertJRules`.
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJRules.java
@@ -7,7 +7,6 @@ import com.google.common.collect.Multiset;
 import com.google.errorprone.refaster.Refaster;
 import com.google.errorprone.refaster.annotation.AfterTemplate;
 import com.google.errorprone.refaster.annotation.BeforeTemplate;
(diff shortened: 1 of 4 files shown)


mohamedsamehsalah (approved):
LGTM 👍

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJRules.java
```diff
@@ -162,35 +159,6 @@ OptionalLongAssert after(OptionalLong optional, long expected) {
     }
   }
 
-  //
```
rickie, line 165: Don't we want to keep this comment?

rickie (approved):
LGTM! Sorry for delay in reviewing.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJRules.java
Stephan202: Ah, I thought it was the only rule in this section; reinstated it. :+1:

PR #1710, opened 2025-06-03
Introduce `AbstractThrowableAssertHasCauseReference` Refaster rule

timtebeek:
- Following https://github.com/assertj/assertj/issues/3715

@rickie would this be an accepted contribution if tests are added?

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRules.java
@@ -759,4 +759,18 @@ final class AssertJThrowingCallableRules {
       return abstractThrowableAssert.withFailMessage(message, args);
     }
   }
+
+  static final class AbstractThrowableAssertHasCauseReference {
+    @BeforeTemplate
+    AbstractThrowableAssert<?, ? extends Throwable> before(
+        AbstractThrowableAssert<?, ? extends Throwable> throwableAssert, Throwable expected) {
+      return throwableAssert.hasCauseReference(expected);
+    }
+
+    @AfterTemplate
+    AbstractThrowableAssert<?, ? extends Throwable> after(
+        AbstractThrowableAssert<?, ? extends Throwable> throwableAssert, Throwable expected) {
+      return throwableAssert.cause().isSameAs(expected);
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


rickie:
Yeah this looks like a nice one! Ideal to make a simple migration of this deprecation easier 😄.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRules.java
```diff
@@ -759,4 +759,19 @@ static final class AbstractThrowableAssertWithFailMessage {
...
+        AbstractThrowableAssert<?, ? extends Throwable> throwableAssert, Throwable expected) {
+      return throwableAssert.hasCauseReference(expected);
+    }
+
+    @AfterTemplate
+    AbstractThrowableAssert<?, ? extends Throwable> after(
+        AbstractThrowableAssert<?, ? extends Throwable> throwableAssert, Throwable expected) {
+      return throwableAssert.cause().isSameAs(expected);
+    }
+  }
```
Stephan202: Other rules in this class suffer the same issue, so not a deal-breaker, but note that this rewrite changes semantics: subsequent chained assertions will now be applied to the cause, rather than the top-level exception. All add an `XXX` comment calling this out.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/AssertJThrowingCallableRulesTestInput.java
```diff
@@ -218,4 +219,12 @@ void testAssertThatThrownByIsInstanceOf() {
             .isInstanceOf(IllegalArgumentException.class)
             .withFailMessage(String.format("foo %s %s", "bar", 1)));
   }
+
+  @SuppressWarnings("deprecation" /* Migrates away from deprecated API. */)
+  AbstractThrowableAssert<?, ? extends Throwable> testAbstractThrowableAssertCauseIsSameAs() {
+    IllegalArgumentException illegalArgumentException = new IllegalArgumentException();
+    IllegalStateException illegalStateException =
+        new IllegalStateException(illegalArgumentException);
+    return assertThat(illegalStateException).hasCauseReference(illegalArgumentException);
```
Stephan202: Usually we keep these tests minimal, even if nonsensical; will propose something.

Stephan202 (approved):
Generally we drop (or at least don't keep around) rules that reference deprecated code, based on the assumption that after an initial cleanup, they are unlikely to match again. But there is value in this, of course, so let's go :)

Suggested commit message:
```
Introduce `AbstractThrowableAssertCauseIsSameAs` Refaster rule (#1710)
```

PR #1755, opened 2025-07-11
Introduce `EmptyString` and `StringIdentity` Refaster rules

timtebeek:
As discussed on 
- fixes https://github.com/openrewrite/rewrite-static-analysis/issues/620#issuecomment-3052939671

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
@@ -222,6 +222,19 @@ final class StringRules {
     }
   }
 
+  /** Prefer empty string literals over creating new empty String objects. */
+  static final class EmptyStringLiteral {
+    @BeforeTemplate
+    String before() {
+      return new String("");
+    }
+
+    @AfterTemplate
+    String after() {
+      return "";
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


timtebeek:
We could; that would pull it into potentially breaking. You all are a little more opinionated, and I'd be fine with that as well, but figured start small and safe. Can adjust if desired.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/StringRules.java
```diff
@@ -222,6 +222,19 @@ Function<Object, String> after() {
     }
   }
 
+  /** Prefer empty string literals over creating new empty String objects. */
+  static final class EmptyStringLiteral {
```
Stephan202: We don't really have a good heuristic for the ordering of rules inside a rules collection, but roughly speaking right now the simplest / "most fundamental" go at the top. Let me move it there.

(Perhaps at some point we'll just automatically order them lexicographically 😅.)

Stephan202 (approved):
I rebased and added a commit with the suggested generalization, but also introduced an alternative variant to the originally suggested rule. 😄

Suggested commit message:
```
Introduce `EmptyString` and `StringIdentity` Refaster rules (#1755)
```

PR #1756, opened 2025-07-11
Introduce `RandomGeneratorRules` Refaster rule collection

timtebeek:
- Fixes https://github.com/openrewrite/rewrite-static-analysis/issues/621

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/RandomGeneratorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/RandomGeneratorRules.java
@@ -7,6 +7,10 @@ import java.util.random.RandomGenerator;
 import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
 
 /** Refaster rules related to expressions dealing with {@link RandomGenerator} instances. */
+// XXX: Consider introducing an Error Prone bug pattern that identifies expressions of the form `c +
+// r.next{Int,Long,Double}(b)` for literals `c` and `b`, and replaces them with
+// `r.next{Int,Long,Double}(c, s)` where `s = c + b`. Similarly with the operands commuted, and
+// for the latter variant also for operator `-`.
 @OnlineDocumentation
 final class RandomGeneratorRules {
   private RandomGeneratorRules() {}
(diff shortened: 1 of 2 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/RandomRules.java
```diff
@@ -0,0 +1,30 @@
...
+
+/** Refaster rules related to expressions dealing with {@link Random}. */
+@OnlineDocumentation
+final class RandomRules {
+  private RandomRules() {}
+
+  /** Prefer {@link Random#nextInt(int)} over casting a scaled random double to int. */
+  static final class RandomNextInt {
+    @BeforeTemplate
+    int nextDouble(Random random, int bound) {
```
rickie, line 16: We usually call the before template methods `before` and `before2` :).

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/RandomRulesTestInput.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.Random;
+import tech.picnic.errorprone.refaster.test.RefasterRuleCollectionTestCase;
+
+final class RandomRulesTest implements RefasterRuleCollectionTestCase {
+  @Override
+  public ImmutableSet<Object> elidedTypesAndStaticImports() {
+    return ImmutableSet.of();
+  }
+
+  int testRandomNextInt() {
```
rickie, line 13: We generally have 1 test method per Refaster rule :).

There is a `RandomRules.class` missing in the `RefasterRulesTest#RULE_COLLECTIONS` 😄. We'll see that the test will fail as it expects these tests to be combined in one method :)

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/RandomRulesTestInput.java
```diff
@@ -0,0 +1,20 @@
+package tech.picnic.errorprone.refasterrules;
+
+import com.google.common.collect.ImmutableSet;
+import java.util.Random;
+import tech.picnic.errorprone.refaster.test.RefasterRuleCollectionTestCase;
+
+final class RandomRulesTest implements RefasterRuleCollectionTestCase {
+  @Override
```
rickie, line 8: I think we can simply omit this one :)

rickie (commented):
Thanks for the PR @timtebeek. We'll include this one with your other PR soon. There were some holidays that resulted in some slower reviews 😬.

Stephan202:
I see there's also `RandomGenerator#nextDouble(double bound)`; added another commit.

Stephan202:
Rebased and added one more commit :)

PR #1763, opened 2025-07-16
Introduce BugChecker for implicit var usage

dskop:
This PR introduces an `error-prone` pattern to check for implicit `var` keyword usage.
Examples:
```java
var a = getList(); // should be flagged

var a = "string literal"; // should NOT be flagged because the expression includes literal
var a = (List) getList(); // should NOT be flagged because the expression includes type cast
var a = new Object(); // should NOT be flagged because the expression includes `new` operation
var a = getForType(List.class); // should NOT be flagged because the expression includes .class literal
var a = getList().map(String::new).findAny(); // should NOT be flagged because the expression includes constructor reference
var a = ImmutableList.of(42) // should NOT be flagged because the expression includes static constructor method
```


rickie:
Nice work @dskop, by running `./apply-error-prone-suggestions.sh` we can resolve most of the errors that are pointed out! 

Make sure to read the README on how to make sure you can run that :). If you're stuck with it, feel free to DM me on Slack.

PR #1773, opened 2025-07-20
Introduce `AssertJDurationRules` Refaster rule collection

timtebeek:
Figured this would fit in nicely with the other AssertJ rules for more expressive assertions.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJDurationRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJDurationRules.java
@@ -13,15 +13,14 @@ import tech.picnic.errorprone.refaster.annotation.OnlineDocumentation;
 /**
  * Refaster rules related to AssertJ assertions over {@link Duration}s.
  *
- * <p>These rules simplify and improve the readability of Duration assertions by using the more
- * specific AssertJ Duration assertion methods instead of generic assertions on the Duration's
- * numeric values.
+ * <p>These rules simplify and improve the readability of tests by using {@link Duration}-specific
+ * AssertJ assertion methods instead of generic assertions.
  */
 @OnlineDocumentation
 final class AssertJDurationRules {
   private AssertJDurationRules() {}
 
-  static final class AbstractDurationAssertHasNanos {
+  static final class AssertThatHasNanos {
     @BeforeTemplate
     AbstractLongAssert<?> before(Duration duration, long nanos) {
       return assertThat(duration.toNanos()).isEqualTo(nanos);
@@ -33,7 +32,7 @@ final class AssertJDurationRules {
     }
   }
 
-  static final class AbstractDurationAssertHasMillis {
+  static final class AssertThatHasMillis {
     @BeforeTemplate
     AbstractLongAssert<?> before(Duration duration, long millis) {
       return assertThat(duration.toMillis()).isEqualTo(millis);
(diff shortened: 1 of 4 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJDurationRules.java
```diff
@@ -0,0 +1,119 @@
...
+    @AfterTemplate
+    AbstractDurationAssert<?> after(Duration duration, long millis) {
+      return assertThat(duration).hasMillis(millis);
+    }
+  }
+
+  static final class AbstractDurationAssertHasSeconds {
+    @BeforeTemplate
+    AbstractLongAssert<?> before(Duration duration, long seconds) {
+      return assertThat(duration.toSeconds()).isEqualTo(seconds);
```
Stephan202, line 50: We can also rewrite `getSeconds()` calls. (Further up, that's not true for `getNanos()`.)

_Edit:_ no need; [`JavaDurationGetSecondsToToSeconds`](https://errorprone.info/bugpattern/JavaDurationGetSecondsToToSeconds) rewrites `getSeconds()` to `toSeconds()` already.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJDurationRules.java
```diff
@@ -0,0 +1,119 @@
...
+    @BeforeTemplate
+    AbstractBooleanAssert<?> before(Duration duration) {
+      return assertThat(duration.isZero()).isTrue();
+    }
+
+    @AfterTemplate
+    AbstractDurationAssert<?> after(Duration duration) {
+      return assertThat(duration).isZero();
+    }
+  }
```
Stephan202, line 113: For numeric types we migrate away from `isZero()` in favour of `isEqualTo`([example](https://github.com/PicnicSupermarket/error-prone-support/blob/f5c2943d61250655f00874469263402b5f5cb738/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJIntegerRules.java#L42-L52)), but for this non-numeric type I suppose the opposite direction does make more sense. Let's then also add a match for `isEqualTo(Duration.ZERO)`.

(I _could_ see us change our mind on this, but then at least we'll remember to update both cases.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJDurationRules.java
```diff
@@ -0,0 +1,119 @@
...
+
+  static final class AbstractDurationAssertIsNegative {
+    @BeforeTemplate
+    AbstractBooleanAssert<?> before(Duration duration) {
+      return assertThat(duration.isNegative()).isTrue();
+    }
+
+    @AfterTemplate
+    AbstractDurationAssert<?> after(Duration duration) {
+      return assertThat(duration).isNegative();
```
Stephan202, line 145: If we similarly add a match against `isLessThan(Duration.ZERO)`, then we can also introduce an `isPositive` rule; will add. (Though `Duration#isPositive()` is JDK 18+, so with the current build I can't add a rule for that. Raising the build baseline is out of scope for this PR :).)

Stephan202 (approved):
Another nice one! Added a commit.

Suggested commit message:
```
Introduce `AssertJDurationRules` Refaster rule collection (#1773)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJDurationRules.java
rickie: Yeah I could also see us go the other way for using the `isZero` specific operators. I know that on the OpenRewrite side, they are preferring this for numeric types.

rickie (approved):
LGTM!

PR #1775, opened 2025-07-20
Add `AssertJPathRules`

timtebeek:
- Similar to https://github.com/PicnicSupermarket/error-prone-support/pull/1773


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJPathRules.java
```diff
@@ -0,0 +1,216 @@
...
+    @SuppressWarnings("AssertThatHasToString")
+    AbstractStringAssert<?> before(Path path, String fileName) {
+      return assertThat(path.getFileName().toString()).isEqualTo(fileName);
+    }
+
+    @AfterTemplate
+    AbstractPathAssert<?> after(Path path, String fileName) {
+      return assertThat(path).hasFileName(fileName);
+    }
+  }
```
timtebeek, line 155: Calling out the `@SuppressWarnings` here for review scrutiny: We're converting to `hasFileName` here instead of `hasToString`.

Alternatively we could (instead or also) convert
`assertThat(path.getFileName()).hasToString(fileName)` to 
`assertThat(path).hasFileName(fileName)`.

Stephan202 (commented):
Very cool. Rebased and added a commit. Some changes rely on #1777.

Suggested commit message:
```
Introduce `AssertJPathRules` Refaster rule collection (#1775)
```

rickie (approved):
Very nice improvements 🎇

PR #1776, opened 2025-07-20
Add AssertJ rules for comparable, equality and array length

timtebeek:
With an eye towards fixing
- https://github.com/openrewrite/rewrite-testing-frameworks/issues/394
- https://rules.sonarsource.com/java/RSPEC-5838


Stephan202 (commented):
Nice! I added a commit. So far looked only at the `AssertJObjectRules` changes. Remainder is TBD.

I suspect that the `AssertJPrimitiveRules` additions should be moved elsewhere, as they're not specific to primitives? Might also be nicer to move this to a separate PR. The current PR could be merged with a message such as
```
Extend `AssertJObjectRules` Refaster rule collection (#1776)
```

WDYT @timtebeek?

@rickie I suggest that we merge this PR _after_ #1779, so that the new references to `AssertJIsNull` make sense.

timtebeek:
I had indeed briefly considered adding a new rule for array assertions, but it felt a bit niche when all we cover is array length which itself is a primitive. Let me know if you'd want me to make adjustments!

Stephan202:
I suspect we'll need templates for each primitive array type as well (to be checked how Refaster handles this). Could be in the same rules if we ignore type safety (in the future we might have a check that flags this). So I suppose a new `AssertJArrayRules` class makes some sense. :thinking:.

timtebeek:
> I suspect we'll need templates for each primitive array type as well (to be checked how Refaster handles this). Could be in the same rules if we ignore type safety (in the future we might have a check that flags this). So I suppose a new `AssertJArrayRules` class makes some sense. 🤔.

Sure! For now I've split off the `T[]` array length assertions into
- https://github.com/PicnicSupermarket/error-prone-support/pull/1781

Sorry for the review spam 😇

Stephan202:
Rebase and added a commit. I'll look into the remaining build failure later; requires a bit of thinking.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJPrimitiveRules.java
```diff
@@ -19,15 +25,49 @@ static final class AssertThatIsEqualTo {
...
+    AbstractLongAssert<?> before(long actual, long expected) {
+      return Refaster.anyOf(
+          assertThat(actual).isSameAs(expected), assertThat(actual).isSameAs(expected));
+    }
+
+    @BeforeTemplate
+    AbstractFloatAssert<?> before(float actual, float expected) {
+      return Refaster.anyOf(
+          assertThat(actual).isSameAs(expected), assertThat(actual).isSameAs(expected));
+    }
```
Stephan202, line 65: Here and below: due to overload matching we now need to specify all variants explicitly.

Stephan202:
Locally ran integration tests: no changes.

PR #1778, opened 2025-07-20
Introduce `SuggestedFixToBuilder` Refaster rule

Stephan202:
Suggested commit message:
```
Introduce `SuggestedFixToBuilder` Refaster rule
```

PR #1779, opened 2025-07-20
Generalize `AssertJIsNull` check

Stephan202:
:exclamation: This PR is on top of #1778. :exclamation:

Inspired by changes in #1776.

Suggested commit message:
```
Generalize `AssertJIsNull` check

Summary of changes:
- Also map `isSameAs(null)` to `isNull()`.
- Introduce similar suggestions for `isNotNull()`.
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AssertJIsNull.java
```diff
@@ -19,28 +21,36 @@
...
  */
 @AutoService(BugChecker.class)
 @BugPattern(
-    summary = "Prefer `.isNull()` over `.isEqualTo(null)`",
+    summary = "Prefer `.isNull()` and `.isNotNull()` over more verbose alternatives",
     link = BUG_PATTERNS_BASE_URL + "AssertJIsNull",
     linkType = CUSTOM,
     severity = SUGGESTION,
     tags = SIMPLIFICATION)
 public final class AssertJIsNull extends BugChecker implements MethodInvocationTreeMatcher {
```
Stephan202, line 43: I think that the check's name is fine to keep as-is, but open to alternatives. (If we do rename, then the PR title and suggested commit message must be updated accordingly.)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/AssertJIsNullTest.java
```diff
@@ -16,15 +16,17 @@ void identification() {
...
+            "    assertThat(1.0).isNotSameAs(1.0);",
             "    // BUG: Diagnostic contains:",
             "    assertThat(1).isEqualTo(null);",
             "    // BUG: Diagnostic contains:",
-            "    assertThat(\"foo\").isEqualTo(null);",
-            "    isEqualTo(null);",
-            "  }",
-            "",
-            "  private boolean isEqualTo(Object value) {",
-            "    return value.equals(\"bar\");",
```
Stephan202, line 27: I dropped this test, as it seemed redundant.

Stephan202 (commented):
Some context.

Stephan202:
Rebased; PR now targets `master`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/AssertJIsNull.java
rickie: I think that makes sense yes! I like the name.

Stephan202: K, will change.

rickie (approved):
Nice improvements ✨ 

Approving already, the name suggestion LGTM!

PR #1781, opened 2025-07-21
Introduce `AssertJArrayRules` Refaster rule collection

timtebeek:
As suggested on
- https://github.com/PicnicSupermarket/error-prone-support/pull/1776#issuecomment-3094796173
- Fixes https://github.com/openrewrite/rewrite-testing-frameworks/issues/394


timtebeek:
As a quick aside I would have expected the reviewdog workflow to correct my formatting issues with a helpful comment here 🤷🏻

Stephan202 (approved):
Rebased and added another commit

```
Introduce `AssertJArrayRules` Refaster rule collection (#1781)

The new rules rely on a new `IsMultidimensionalArray` matcher. While
there, trim `IsArrayTest` and use it as a template for the new matcher's
tests.
```

PR #1784, opened 2025-07-22
Deploy SNAPSHOTs to the Maven Central snapshot repository

Stephan202:
Suggested commit message:
```
Deploy SNAPSHOTs to the Maven Central snapshot repository (#1784)

Successful builds on the default branch are now followed by a `mvn
deploy` step that signs and uploads the produced SNAPSHOT artifacts. 

While there:
- Remove some unnecessary quotes from GitHub Actions workflow files.
- Move a comment about an Error Prone flag to its proper location.
```

PR #1785, opened 2025-07-22
Migrate to the new Maven Central Publisher Portal

Stephan202:
This is a subset of #1784. Validated by creating a 0.23.1 release and deploying it. (And then dropping it from the portal, as well as deleting the created tags.)

Suggested commit message:
```
Migrate to the new Maven Central Publisher Portal (#1785)

As the OSSRH service reached end-of-life. This unblocks local deployment
of new releases.
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,13 +64,9 @@
         <url>https://github.com/PicnicSupermarket/error-prone-support/actions</url>
     </ciManagement>
     <distributionManagement>
-        <repository>
-            <id>ossrh</id>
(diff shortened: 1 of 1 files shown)


rickie (approved):

Stephan202:
To be sure: the few existing SNAPSHOTs are due to local tests by me, matching the 0.23.0 release scope. Proper support will come with #1784, but I'll only be able to finalize that after my holiday. Stay tuned!

PR #1786, opened 2025-07-23
Introduce `JUnitMethodSourceGenericParams` bug checker

mohamedsamehsalah:
JUnit's parameterized tests, annotated with `@MethodSource`, do not have generic type information at runtime. This means that we can pass a Set of Integers as an argument to a test method which accepts a Set of Strings, and JUnit would disregard the accepted type.

This bug checker flags such cases.

#### Potential Bug ⚠️ 
```java
  private static ImmutableSet<Arguments> demoTestCases() {
    return ImmutableSet.of(arguments(ImmutableSet.of(1)));
  }

  @ParameterizedTest
  @MethodSource("demoTestCases")
  void demo(ImmutableSet<String> setOfStrings) {
    ImmutableSet<Integer> setOfIntegers = ImmutableSet.of(1);

    assertThat(setOfStrings).isEqualTo(setOfIntegers); // true
  }
```

### Suggested commit message
```
Introduce `JUnitMethodSourceGenericParams` bug checker
```


rickie (commented):
Nice check idea! Mostly testing that should be improved. I didn't go through the code yet properly, will do that after my holiday :)

mohamedsamehsalah (commented):
Thanks for review @rickie 🙌 

I tried to address the mutation test reports, but I will need help to cover all of them. I tried to put the reasoning in the comments.

rickie (commented):
Only some small comments so far. Takes a while to review, but flashing already.

rickie (commented):
Lol I confused myself with flushing what I had and not 😬.
Pushing a commit with some changes. Let's discuss some of the mutants to get that more sorted out :). 

The check has a nice shape. Let's do these two things:
- Run it on a repo of ours internally (I recall you might have done this a while back as you asked me about it 🤔).
- I've to continue reviewing :)

PR #1795, opened 2025-08-07
Add BufferedReader Refaster rules to FileRules

timtebeek:
## Summary
- Adds four new Refaster rules to simplify `BufferedReader` creation from `FileInputStream`
- Implements transformations described in https://github.com/openrewrite/rewrite-static-analysis/issues/694
- Closes https://github.com/PicnicSupermarket/error-prone-support/pull/1795

## Changes
Added the following Refaster rules to `FileRules.java`:

1. **FilesNewBufferedReaderFromStringPath**: Transforms `new BufferedReader(new InputStreamReader(new FileInputStream(String)))` to `Files.newBufferedReader(Path.of(String))`
2. **FilesNewBufferedReaderFromFilePath**: Transforms File-based variant to `Files.newBufferedReader(File.toPath())`  
3. **FilesNewBufferedReaderFromStringPathWithCharset**: Adds charset support for String paths
4. **FilesNewBufferedReaderFromFilePathWithCharset**: Adds charset support for File paths

## Test plan
- [x] Added test cases in `FileRulesTestInput.java` and `FileRulesTestOutput.java`
- [ ] Tests pass with `mvn clean verify` (pending -Werror resolution)

🤖 Generated with [Claude Code](https://claude.ai/code)


timtebeek:
Ok `mvn com.spotify.fmt:fmt-maven-plugin:format` does not agree with the testing framework: remove the imports and the tests fail, retain the imports and the formatter fails. Any suggestion @mohamedsamehsalah ? I think you're the only one not on holiday there. ;)

mohamedsamehsalah:
> Any suggestion @mohamedsamehsalah ? I think you're the only one not on holiday there. ;)

Hey @timtebeek , thanks for reaching out 🙌 

@Stephan202 should be back, but I can't speak about his (un)availability. I will take a look.

Stephan202:
The fix is to add an `elidedTypesAndStaticImports` method to both the input and output file. I rebased and added a commit. (A future version of the test framework should do away with this arcane solution, but at the current pace that's months out.)

I have yet to review the PR, but have to leave soon. Will do later/tomorrow :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
```diff
@@ -174,4 +177,70 @@ boolean after(File file) {
...
+  static final class FilesNewBufferedReaderFromStringPath {
+    @BeforeTemplate
+    @SuppressWarnings("DefaultCharset" /* This violation will be rewritten. */)
+    BufferedReader before(String path) throws IOException {
+      return new BufferedReader(new InputStreamReader(new FileInputStream(path)));
+    }
+
+    @AfterTemplate
+    BufferedReader after(String path) throws IOException {
+      return Files.newBufferedReader(Path.of(path));
```
Stephan202, line 196: We can actually add one more case here: dropping `UTF-8` when specified explicitly. (This one may be slightly controversial, but in this repo we _generally_ prefer dropping defaults when possible.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/FileRules.java
```diff
@@ -174,4 +177,70 @@ boolean after(File file) {
...
+    }
+  }
```
Stephan202: Here and below: strictly speaking this isn't behavior-preserving, as we change the charset used. I'll add a comment calling this out.

(I also considered adding an extra template to also rewrite the UTF-8 case, but i.c.w. the suggestion above that case is taken care of when chained with the one below.)

Stephan202 (approved):
```
Introduce `FilesNewBufferedReader{PathOf,ToPath}{,WithCharset}` Refaster rules (#1795)
```

Stephan202:
I can imagine that the skipped SonarCloud analysis will cause something to be flagged when we merge this. We should once again [review](https://community.sonarsource.com/t/no-sonarcloud-analysis-available-for-prs-from-fork/132569/2) whether to enable automatic analysis, or do something fancier.

PR #1801, opened 2025-08-11
Stabilize `metrics` integration test setup

Stephan202:
Suggested commit message:
```
Stabilize `metrics` integration test setup (#1801)

By disabling two identically-named tests that attempt to connect to
`example.com:80`. This connection operation frequently times out.

While there, disable GitHub Action's default fail-fast behavior, so that
a failure in one integration test does not cause the others to be
cancelled.
```

Will run the integration tests a few times to assess whether this indeed improves stability.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -15,8 +15,7 @@ on:
   issue_comment:
     types: [ created ]
   push:
-    # XXX: Revert
-    branches: [ 'master', 'sschroevers/stabilize-metrics-integration-test' ]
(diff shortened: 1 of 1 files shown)


Stephan202:
Looking at the stability of the integration tests on `master`, and based on some local testing, it looks like `example.com:80` is now stable again. I'll consider backing that part of the change out, for now.

rickie:
Having the `fail-fast` would be amazing, makes it easier to see which ones are actually failing.

Stephan202:
Reverted the test changes. PR is ready for review.

PR #1808, opened 2025-08-13
First rule: MonoFromFluxJustToMonoJust

natanshimeles:


Stephan202:
(I dropped the `WIP` label because it's redundant with this PR being in draft status :).)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -512,9 +513,13 @@ Flux<T> after(T value) {
   /** Don't unnecessarily transform a {@link Mono} to an equivalent instance. */
   static final class MonoIdentity<T> {
     @BeforeTemplate
+    @SuppressWarnings("IdentityConversion" /* This is a more specific template. */)
```
natanshimeles, line 516: I was not really sure about this, as this is covered by IdentityConversion

Stephan202: Good catch. So in fact we can revert this change, because indeed `IdentityConversion` covers this case already. 👍

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -97,8 +97,9 @@ Mono<T> after() {
   /** Prefer {@link Mono#just(Object)} over more contrived alternatives. */
   static final class MonoJust<T> {
     @BeforeTemplate
+    @SuppressWarnings("FluxNext" /* This is a more specific template. */)
     Mono<T> before(T value) {
-      return Mono.justOrEmpty(Optional.of(value));
+      return Refaster.anyOf(Mono.justOrEmpty(Optional.of(value)), Mono.from(Flux.just(value)));
```
Stephan202, line 102: Thanks for using the correct suppression text ❤️. I think this is right, but looking at it, if we _anyway_ want to add a special rule for `Flux.just(value)`, then instead of adding a case for `Mono.from(Flux.just(value))`, we can add a case for `Flux.just(value).next()`. That way:
1. Implicitly we still cover the former (just with two rewrite steps).
2. We cover one more case.
3. We won't need the warning suppression.

Stephan202 (commented):
Might have some OCD feedback later (which I'm happy to directly push to the branch), but for now a quick GitHub UI-based review. Nice stuff!

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
natanshimeles: Perfect 👨‍💻 !

mohamedsamehsalah (approved):
Congratulations on your first contribution @natanshimeles 👏 

Thanks @Stephan202 for the feedback 🙇

PR #1814, opened 2025-08-15
Suggest simplification of various `Flux` creation expressions

Stephan202:
Suggested commit message:
```
Suggest simplification of various `Flux` creation expressions (#1814)

Summary of changes:
- Update `ExplicitArgumentEnumeration` to replace `Flux#fromIterable`
  invocations with equivalent `Flux#just` invocations where possible.
- Introduce Refaster templates for `Flux#fromStream` invocations that
  can be replaced with simpler `Flux#just` invocations.
- Introduce a new Refaster rule to suggest `Flux#fromArray` over more
  ambiguous or verbose alternatives, the former of which may be
  introduced by the other two changes.
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/ExplicitArgumentEnumeration.java
```diff
@@ -90,11 +90,10 @@ public final class ExplicitArgumentEnumeration extends BugChecker
                   .named("copyOf"),
               symbolMatcher(
                   (symbol, state) ->
-                      state
-                          .getSymtab()
-                          .arrayClass
-                          .equals(((MethodSymbol) symbol).params().get(0).type.tsym))),
+                      state.getTypes().isArray(((MethodSymbol) symbol).params().get(0).type))),
```
Stephan202, line 93: Nonfunctional simplification.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -491,16 +493,12 @@ Flux<Integer> before(int value) {
...
-    // XXX: Consider generalizing part of this template using an Error Prone check that covers any
-    // sequence of explicitly enumerated values passed to an iteration order-preserving collection
-    // factory method.
     @BeforeTemplate
     Flux<T> before(T value) {
       return Refaster.anyOf(
           Mono.just(value).flux(),
-          Mono.just(value).repeat().take(1),
-          Flux.fromIterable(ImmutableList.of(value)),
-          Flux.fromIterable(ImmutableSet.of(value)));
```
Stephan202, line 503: These two were dropped, as the `ExplicitArgumentEnumeration` changes cover many more variants.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -491,16 +493,12 @@ Flux<Integer> before(int value) {
...
+          Flux.fromStream(() -> Stream.of(value)),
```
Stephan202, line 500: This one needs to be split off from `FluxJustArray` below: that rule doesn't match this case, and due to there being two `Flux#just` overloads, changing the other rule to use only `@Repeated T values` will cause _only_ the singleton case to be covered.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/ExplicitArgumentEnumerationTest.java
```diff
@@ -41,12 +41,15 @@ void identification() {
...
+            "    Flux.fromIterable(ImmutableList.copyOf(new String[0]));",
+            "    Flux.fromIterable(ImmutableList.copyOf(ImmutableList.of()));",
+            "",
             "    ImmutableList.builder()",
             "        // BUG: Diagnostic contains:",
-            "        .addAll(ImmutableList.of())",
-            "        // BUG: Diagnostic contains:",
-            "        .addAll(ImmutableList.copyOf(new String[0]))",
-            "        .addAll(ImmutableList.copyOf(ImmutableList.of()))",
+            "        .addAll(ImmutableList.of(1))",
```
Stephan202, line 52: Just moved these up, as the `Flux` expression is now first.

Stephan202 (commented):
Some context.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
mohamedsamehsalah: It would be cool if the `ExplicitArgumentEnumeration` could detect this, by checking if the supplied value has a unary argument. 

I tried to make this work, but this would require extending the logic of checking the return types of the supplier argument, too complex for a Saturday 🦥

mohamedsamehsalah (approved):
Smart stuff 🧠

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
Stephan202: Yep, this would be nice. I'll add a comment to track this idea.

Stephan202 (commented):
Rebased and added a commit. Besides tracking the discussed idea, also simplified one of the Refaster rules.

PR #1832, opened 2025-08-22
Suppress various SonarCloud violations

Stephan202:
Suggested commit message:
```
Suppress various SonarCloud violations
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/AssertJRules.java
```diff
@@ -99,6 +99,9 @@
 // See how we can combine these things. Do note that (at present) their Refaster rules don't
 // show up as Error Prone checks. So we'd have to build an integration for that.
 // XXX: Cover all cases listed by https://rules.sonarsource.com/java/RSPEC-5838/
+// XXX: For `E extends Comparable<? super E>`, rewrite
+// `assertThat(iterable).isEqualTo(Refaster.<Object>anyOf(ImmutableSortedSet.of(expected),
+// ImmutableSortedMultiset.of(expected)))` to `assertThat(iterable).containsExactly(expected)`.
```
Stephan202, line 104: Not sure this will even work. I noticed other rules in this class that may in fact be dodgy. Something for another time.

rickie (approved):

PR #1871, opened 2025-09-18
Replace `hasSize(0)` check with `isEmpty()` in `AssertJIterableRules`

timtebeek:
- Closes https://github.com/openrewrite/rewrite-testing-frameworks/issues/811


timtebeek:
Weird how this was reported to us as a bug of a missed conversion. Need to explore a bit more, and possibly then close this PR.

PR #1887, opened 2025-09-23
Rewrite `{BigDecimal,Integer,Long}SignumIs{Positive,Negative}` Refaster rules

henri-tremblay:
The `compareTo` in Java was made to match the mental comparison we would do with numbers. x - y < 0, x - y == 0, x - y > 0 would transfer to x.compareTo(y) < 0, x.compareTo(y) == 0, x.compareTo(y) > 0. We never directly compare to -1, 0, 1 to keep the lisibility.

When using signum, it's the same thing. You would do x.signum() < 0 making it clear it's a negative number. When doing x.signum() == -1, you need to think about it. "What do -1 mean?" "Ah! It means negative".

I would highly suggest that error prone follows these guidelines. To be sure it wasn't only me, I asked some my favorite AI what it prefers. Here is the answer.

> ## Recommendation
> **Use `bigDecimal.signum() > 0`** because:
> 1. It clearly communicates that you want to check if the number is positive
> 2. It's more maintainable and less brittle
> 3. It follows the principle of expressing intent rather than implementation details
> 4. It's the same pattern you'd use for other numeric types (`number > 0`)

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
@@ -97,7 +97,7 @@ final class BigDecimalRules {
   }
 
   /**
-   * Prefer a {@link BigDecimal#signum()} comparison to 1 over more contrived or less clear
+   * Prefer a {@link BigDecimal#signum()} comparison to 0 over more contrived or less idiomatic
    * alternatives.
    */
   static final class BigDecimalSignumIsPositive {
@@ -106,6 +106,7 @@ final class BigDecimalRules {
       return Refaster.anyOf(
           value.compareTo(BigDecimal.ZERO) > 0,
           BigDecimal.ZERO.compareTo(value) < 0,
+          value.signum() == 1,
           value.signum() >= 1);
     }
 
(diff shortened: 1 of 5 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java
```diff
@@ -93,14 +93,13 @@ boolean before(BigDecimal value) {
       return Refaster.anyOf(
           value.compareTo(BigDecimal.ZERO) > 0,
           BigDecimal.ZERO.compareTo(value) < 0,
-          value.signum() > 0,
```
Stephan202, line 109: Here and below: this line should not be dropped, but be replaced with the `value.signum() == 1` case that we no longer write to.

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/BigDecimalRulesTestOutput.java
```diff
@@ -31,25 +31,25 @@ ImmutableSet<Boolean> testBigDecimalSignumIsZero() {
...
-        BigDecimal.valueOf(7).signum() != 1,
-        BigDecimal.valueOf(8).signum() != 1);
+        BigDecimal.valueOf(1).signum() > 0,
+        BigDecimal.valueOf(2).signum() > 0,
+        BigDecimal.valueOf(3).signum() > 0,
+        BigDecimal.valueOf(4).signum() > 0,
+        BigDecimal.valueOf(5).signum() <= 0,
+        BigDecimal.valueOf(6).signum() <= 0,
+        BigDecimal.valueOf(7).signum() <= 0,
+        BigDecimal.valueOf(8).signum() <= 0);
```
Stephan202, line 45: The input test cases should be updated to reflect the updated rules.

henri-tremblay:
Great! Thank you

PR #1919, opened 2025-10-13
Introduce assorted `{Flux,Mono}#timeout` fallback Refaster rules

mohamedsamehsalah:
I feel like the 2nd option is less contrived, as it eliminates the need to know that `timeout` throws a `TimeoutException`.

### Suggested commit message 🎉 
```
Introduce `MonoTimeout(Empty|Fallback)` refaster rules
```

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
@@ -103,7 +103,7 @@ final class ReactorRules {
    * Prefer {@link Mono#timeout(Duration, Mono)} over more contrived or less performant
    * alternatives.
    */
-  static final class MonoTimeoutMonoEmpty<T> {
+  static final class MonoTimeoutDurationMonoEmpty<T> {
     @BeforeTemplate
     Mono<T> before(Mono<T> mono, Duration duration) {
       return mono.timeout(duration).onErrorComplete(TimeoutException.class);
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -98,6 +99,102 @@ Mono<T> after() {
...
+  }
+
+  /**
+   * Prefer {@link Mono#timeout(Duration, Mono)} over more contrived or less performant
+   * alternatives.
+   */
+  static final class MonoTimeoutMonoFallback<T> {
+    @BeforeTemplate
+    Mono<T> before(Mono<T> mono, Duration duration, Mono<T> fallback) {
+      return mono.timeout(duration).onErrorResume(TimeoutException.class, e -> fallback);
```
Stephan202: Let's validate that Refaster doesn't rewrite this case if `e` is used by `fallback`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -98,6 +99,102 @@ Mono<T> after() {
...
+    Mono<T> after(Mono<T> mono, Publisher<S> other) {
+      return mono.timeout(other, Mono.empty());
+    }
+  }
+
+  /**
+   * Prefer {@link Mono#timeout(Duration, Mono)} over more contrived or less performant
+   * alternatives.
+   */
+  static final class MonoTimeoutMonoJust<T> {
```
Stephan202: For consistency these rules should be have `Duration` rather than `Publisher` in their name.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -98,6 +99,102 @@ Mono<T> after() {
...
+    Mono<T> after(Mono<T> mono, Duration duration, T fallbackValue) {
+      return mono.timeout(duration, Mono.just(fallbackValue));
+    }
+  }
+
+  /**
+   * Prefer {@link Mono#timeout(Duration, Mono)} over more contrived or less performant
+   * alternatives.
+   */
+  static final class MonoTimeoutMonoFallback<T> {
```
Stephan202: I think here we can leave out `Fallback`; where possible we only reference type and method names. (For `Publisher` vs. `Duration` it's inevitable.)

error-prone-contrib/src/test/resources/tech/picnic/errorprone/refasterrules/ReactorRulesTestInput.java
```diff
@@ -72,6 +74,40 @@ ImmutableSet<Mono<String>> testMonoEmpty() {
...
+  }
+
+  ImmutableSet<Mono<Integer>> testMonoTimeoutMonoFallback() {
+    return ImmutableSet.of(
+        Mono.just(1)
+            .timeout(Duration.ofSeconds(2))
+            .onErrorResume(TimeoutException.class, e -> Mono.empty()),
+        Mono.just(3)
+            .timeout(Duration.ofSeconds(4))
+            .onErrorResume(TimeoutException.class, e -> Mono.just(5)));
```
Stephan202: Here and below `Mono.empty()` and `Mono.just()` aren't different enough to require two variants. (But I'll replace one case to test the "`e` is used" case.)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/ReactorRules.java
```diff
@@ -98,6 +99,102 @@ Mono<T> after() {
...
+    Mono<T> after(Mono<T> mono, Duration duration) {
+      return mono.timeout(duration, Mono.empty());
+    }
+  }
+
+  /**
+   * Prefer {@link Mono#timeout(Publisher, Mono)} over more contrived or less performant
+   * alternatives.
+   */
+  static final class MonoTimeoutPublisherMonoEmpty<T, S> {
```
Stephan202: Let's group the `Publisher` variants.

rickie:
Latest changes LGTM!

PR #1927, opened 2025-10-19
Target Java 21, build with Java 25

Stephan202:
:exclamation: ~~To be merged after #1942~~. :exclamation:

Suggested commit message (to be augmented later):
```
Target Java 21, build with Java 25 (#1927)

Raising the runtime baseline to Java 21 is necessary for compatibility with
Error Prone 2.43.0+. By raising the build baseline to Java 25 we reduce
maintenance overhead, without impacting downstream users.

Summary of changes:
- Generalize `generate-version-compatibility-overview.sh` to handle distinct
  JDK requirements and omit custom Git tags.
- Introduce SDKMAN! configuration to streamline environment setup for external
  contributors and to simplify future
  `generate-version-compatibility-overview.sh` maintenance.
- Update code to make use of Java 21 language features.
- Update code comments in accordance with the new Java baseline.
- Drop obsolete flags from `.mvn/jvm.config`.
```

This is a ~draft~ PR, to be merged just before we upgrade to the next version of Error Prone, which itself will [target](https://github.com/google/error-prone/commit/6a6afc9d05c1349135e16769082463a2d0121873) Java 21. Open points:
- ~Release one more version of Error Prone Support targeting Java 17 before merging this PR.~ :heavy_check_mark: 
- ~Move the Refaster rule changes to a separate PR, to be merged after this one.~ :heavy_check_mark: 
- ~Drop the Java 17 assumption made by `website/generate-version-compatibility-overview.sh`.~ :heavy_check_mark: 
- ~Understand why not all `list.get(0)` occurrences are flagged by the new `ListGetFirst` Refaster rule.~ :heavy_check_mark:
- ~Investigate other cleanup opportunities and improvements that may be unlocked by targeting Java 21.~ :heavy_check_mark:

Note that raising the _build_ baseline to Java 25 isn't strictly necessary, but it (a) matches what we do for Maven (i.e., always requires the latest version) and (b) means that we retain the need for the Maven Toolchains setup (otherwise I'd be torn to drop all that for the time being, only to have to reintroduce it again later). I propose that we officially move to a cadence where we _build_ with the latest LTS and _target_ the oldest LTS supported by Error Prone.

PR #1938, opened 2025-10-30
Introduce `ObjectMapperValueToTree` Refaster rule

mohamedsamehsalah:
### Suggested commit message 🎉 
```
Introduce `ObjectMapperValueToTree` Refaster rule (#1938)
```

### DIFF
--- a/error-prone-contrib/pom.xml
+++ b/error-prone-contrib/pom.xml
@@ -50,6 +50,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-core</artifactId>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JacksonRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JacksonRules.java
@@ -48,7 +48,10 @@ final class JacksonRules {
     }
   }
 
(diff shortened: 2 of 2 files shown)


error-prone-contrib/pom.xml
```diff
@@ -47,6 +47,10 @@
             <artifactId>jackson-annotations</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
```
Stephan202: We should omit this one from the runtime classpath:
```suggestion
            <artifactId>jackson-core</artifactId>
            <scope>provided</scope>
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/JacksonRules.java
```diff
@@ -45,4 +47,17 @@ Optional<JsonNode> after(JsonNode node, String fieldName) {
       return node.optional(fieldName);
     }
   }
+
+  /** Prefer {@link ObjectMapper#valueToTree(Object)} over more contrived alternatives. */
```
Stephan202: It's also significantly less efficient, IIUC.

Stephan202 (approved):
Added a small commit. Tnx @mohamedsamehsalah!

I'd like to wrap up and merge #1927 before this one, so that we stop seeing the Error Prone compatibility check failure on `master`.

rickie (approved):
Nice addition @mohamedsamehsalah !

PR #1946, opened 2025-11-01
Introduce `DequeRules` Refaster rule collection

Stephan202:
~:exclamation: This PR is on top of #1927. :exclamation:~

Suggested commit message:
```
Introduce `DequeRules` Refaster rule collection (#1946)
```


mohamedsamehsalah (approved):
From my days of preparing for solving coding algorithms, that set of rules change a lot of the common terminology used 😄 

But if I think about it, being more "explicit" about what each API does also makes things easier and alleviates some assumptions.

PR #1961, opened 2025-11-07
Introduce `StringFormatted` Refaster rule

timtebeek:
As discussed with @rickie


rickie (approved):
Thanks @timtebeek 🚀 !

PR #1973, opened 2025-11-12
Fix integration tests

rickie:


integration-tests/metrics-expected-changes.patch
```diff
@@ -4343,7 +4382,7 @@
      snapshot.dump(output);
  
 -    assertThat(output.toString()).isEqualTo(String.format("1%n2%n3%n4%n5%n"));
-+    assertThat(output).hasToString(String.format("1%n2%n3%n4%n5%n"));
```
rickie: This one is interesting 🤔, this doesn't do anything right?

Stephan202: It does do something: it replaces `%n` with `System.lineSeparator()`.

rickie:
It seems I was using 21 🤔.

rickie:
[Build](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/19303051827?pr=1973+%28comment%29).

Stephan202:
/integration-test -> [build](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/19389539444?pr=1973+%28comment%29).

integration-tests/metrics-init.patch
```diff
@@ -10,6 +10,19 @@
...
+--- a/metrics-jvm/src/test/java/io/dropwizard/metrics5/jvm/ThreadDumpTest.java
++++ b/metrics-jvm/src/test/java/io/dropwizard/metrics5/jvm/ThreadDumpTest.java
+@@ -37,6 +37,10 @@ class ThreadDumpTest {
+         });
+     }
+ 
++    // XXX: The Refaster `StringFormatted` rule doesn't properly wrap the string concatenation,
++    // causing only the final `%n` occurrence to be replaced with a system line separator.
++    // See https://github.com/google/error-prone/issues/4866.
++    @SuppressWarnings("StringFormatted")
```
Stephan202: This resolves the issue. Let's follow up with a PR that adds possibly-redundant parentheses, and then depend on [`UnnecessaryParentheses`](https://errorprone.info/bugpattern/UnnecessaryParentheses) to remove the parentheses if indeed redundant. This will require an extra pass, and ideally we fix google/error-prone#4866, but let's unblock a possible release first.

NB: Other modified integration test code is impacted by the same issue; it just doesn't cause a test failure.

Stephan202:
Suggested commit message:
```
Fix integration tests (#1973)

By updating the expected changes and working around an identified bug.
```

integration-tests/metrics-init.patch
Stephan202: I filed #1982.

PR #1998, opened 2025-11-22
Disallow static import of `values`

Stephan202:
Suggested commit message:
```
Disallow static import of `values` (#1998)

This matches changes in Error Prone's `BadImport` check.

See https://github.com/google/error-prone/commit/7e6e113bad29f9e53cc0bc3b6613dcf778831cde
```


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/NonStaticImport.java
```diff
@@ -144,7 +144,8 @@ public final class NonStaticImport extends BugChecker implements CompilationUnit
           "newInstance",
           "of",
           "parse",
-          "valueOf");
+          "valueOf",
+          "values");
```
Stephan202: We don't have a test for every value in this list; decided not to add one.

PR #2030, opened 2025-12-12
Omit default commit message prefix for upgrade PRs

rickie:


Stephan202 (approved):
I don't see which internal change caused the prefix to now show up, but let's do this :+1:

Suggested commit message:
```
Omit default commit message prefix for upgrade PRs (#2030)
```

PR #2031, opened 2025-12-12
Introduce `UnvalidatedEnumValueOfInvocation` check

ghokun:
## Context
`Enum.valueOf` method usage may lead to runtime exceptions when the provided argument is not checked against all possible values of method owner.

Example:
```java
class Test {
  enum A {
    ONE, TWO, THREE
  }
  
  enum B {
    ONE, TWO, THREE, FOUR
  }
  
  // Will throw a runtime exception if b is equal to FOUR.
  A convertUnsafe(B b) {
    return A.valueOf(b.name());
  }
}
```

## Scope
Of course it is not able to capture all possible checks on `b` before supplying it into `A.valueOf`. Therefore scope will be limited to following checks.

```java
// BUG: Raw string
A convertUnsafe(String raw) {
  return A.valueOf(raw);
}

// BUG: From invalid string constant
A convertUnsafe() {
  return A.valueOf("FOUR");
}

// BUG: Where B has values that A do not have
A convertUnsafe(B b) {
  return A.valueOf(b.name());
}

// BUG: Value from case is not present in A
A convertUnsafe(B b) {
  return switch(b) { case FOUR -> A.valueOf(b.name()); };
}

// NOT BUG: Value from case is present in A
A convertUnsafe(B b) {
  return switch(b) { case ONE -> A.valueOf(b.name()); };
}

// NOT BUG: From valid string constant
A convertSafe() {
  return A.valueOf("ONE");
}
```

Above checks should also consider the polymorphic variant of `valueOf`:
```
public static <T extends Enum<T>> T valueOf(Class<T> enumClass, String name)
```


ghokun:
@rickie This one is ready for review now.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/EnumValueOfUsage.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    Type enumType = ASTHelpers.getReceiverType(tree);
+    ExpressionTree nameArgument = extractNameArgument(tree, state);
+    ImmutableSet<String> valuesOfSource = findEnumValuesOfReceiver(enumType);
+
+    // Match constants
+    String constantValue = ASTHelpers.constValue(nameArgument, String.class);
+    if (constantValue != null && !valuesOfSource.contains(constantValue)) {
+      return buildDescription(tree)
```
ghokun: I don't think we can introduce a suggested fix for this. Description already highlights possible values.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UncheckedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,165 @@
...
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!ENUM_VALUE_OF.matches(tree, state)) {
+      return Description.NO_MATCH;
+    }
+
+    Type enumType = ASTHelpers.getReceiverType(tree);
+    ExpressionTree nameArgument = extractNameArgument(tree, state);
+    ImmutableSet<String> valuesOfSource = findEnumValuesOfReceiver(enumType);
+
+    // Match constants
```
rickie: We generally don't need comments :)

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/UncheckedEnumValueOfInvocationTest.java
```diff
@@ -0,0 +1,117 @@
...
+            "class Test {",
+            "  void unsafeCases(String raw, B b) {",
+            "    // BUG: Diagnostic matches: INVALID_VALUE",
+            "    A.valueOf(\"FOUR\");",
+            "    // BUG: Diagnostic matches: INVALID_VALUE",
+            "    A.valueOf(A.class, \"FOUR\");",
+            "    // BUG: Diagnostic matches: AVOID_RAW",
+            "    A.valueOf(raw);",
+            "    // BUG: Diagnostic matches: AVOID_RAW",
+            "    A.valueOf(A.class, raw);",
```
rickie: I'm slightly re-ordering some of the test methods in this class

rickie (commented):
Thanks for the cool check! I did a first pass and added a commit with some suggestions 😄.

I didn't put every change I made on as comment on the PR as I was mostly reviewing while I was flying ✈️ and didn't have internet.
If you have any questions about why I made a certain suggestion, let me know :).

The PR is already shaping up nicely!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UncheckedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,181 @@
...
+      return Description.NO_MATCH;
+    }
+
+    Type enumType = ASTHelpers.getReceiverType(tree);
+    Optional<ExpressionTree> optionalNameArgument = extractNameArgument(tree, state);
+
+    if (optionalNameArgument.isEmpty()) {
+      return buildDescription(tree)
+          .setMessage(
+              "No `String` typed `name` argument was found on %s"
```
ghokun: I am open to suggestions on the message for this one.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UncheckedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,171 @@
...
+  public UncheckedEnumValueOfInvocation() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!ENUM_VALUE_OF.matches(tree, state)) {
+      return Description.NO_MATCH;
+    }
+
+    Type enumType = ASTHelpers.getReceiverType(tree);
+    ExpressionTree nameArgument = tree.getArguments().getLast();
```
ghokun: @rickie Let me know what do you think about this approach.

```
A.valueOf(name)
A.valueOf(A.class, name)
```

`name` is basically the last argument in all possible cases. This allows capturing `null` easily, as we do not filter by string type.

ghokun: One thing that I have not verified is, if `tree.getArguments()` is ordered or not. So far it seems so.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UncheckedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,185 @@
...
+ * arguments.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Avoid passing unvalidated arguments to `Enum#valueOf`",
+    link = BUG_PATTERNS_BASE_URL + "UncheckedEnumValueOfInvocation",
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = FRAGILE_CODE)
+public final class UncheckedEnumValueOfInvocation extends BugChecker
```
ghokun:
```suggestion
public final class UnvalidatedEnumValueOfInvocation extends BugChecker
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/UncheckedEnumValueOfInvocationTest.java
```diff
@@ -0,0 +1,136 @@
+package tech.picnic.errorprone.bugpatterns;
+
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+final class UncheckedEnumValueOfInvocationTest {
```
ghokun:
```suggestion
final class UnvalidatedEnumValueOfInvocationTest {
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UncheckedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,185 @@
...
+import tech.picnic.errorprone.utils.SourceCode;
+
+/**
+ * A {@link BugChecker} that flags {@link Enum#valueOf} invocations that contain unchecked
+ * arguments.
+ */
+@AutoService(BugChecker.class)
+@BugPattern(
+    summary = "Avoid passing unvalidated arguments to `Enum#valueOf`",
+    link = BUG_PATTERNS_BASE_URL + "UncheckedEnumValueOfInvocation",
```
ghokun:
```suggestion
    link = BUG_PATTERNS_BASE_URL + "UnvalidatedEnumValueOfInvocation",
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UncheckedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,185 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.SwitchExpressionTree;
+import com.sun.source.tree.SwitchTree;
+import com.sun.tools.javac.code.Type;
+import java.util.List;
+import tech.picnic.errorprone.utils.MoreASTHelpers;
+import tech.picnic.errorprone.utils.SourceCode;
+
+/**
+ * A {@link BugChecker} that flags {@link Enum#valueOf} invocations that contain unchecked
```
ghokun:
```suggestion
 * A {@link BugChecker} that flags {@link Enum#valueOf} invocations that contain unvalidated
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,185 @@
...
+  /** Instantiates a new {@link UnvalidatedEnumValueOfInvocation} instance. */
+  public UnvalidatedEnumValueOfInvocation() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!ENUM_VALUE_OF_METHOD.matches(tree, state)) {
+      return Description.NO_MATCH;
+    }
+
+    Type enumType = ASTHelpers.getReceiverType(tree);
```
Stephan202: The comment below discusses `A.valueOf(A.class, name)`, but as that's a static method, it should be invoked as `Enum.valueOf(A.class, name)`. In that case this statement does not correctly determine the enum type.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,171 @@
...
+  public UncheckedEnumValueOfInvocation() {}
+
+  @Override
+  public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+    if (!ENUM_VALUE_OF.matches(tree, state)) {
+      return Description.NO_MATCH;
+    }
+
+    Type enumType = ASTHelpers.getReceiverType(tree);
+    ExpressionTree nameArgument = tree.getArguments().getLast();
```
Stephan202: Indeed `tree.getArguments()` is ordered. However, my [comment](https://github.com/PicnicSupermarket/error-prone-support/pull/2031/changes#r3201679055) above shows that we need to handle the instance method `Enum#valueOf(String)` and `Enum#valueOf(Class, String)` distinctly anyway. Having two `Matcher`s for those automatically causes us to skip "fake" `valueOf` implementations.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,185 @@
...
+    if (!MoreASTHelpers.isStringTyped(nameArgument, state)) {
+      return buildDescription(tree)
+          .setMessage(
+              "`%s` is not a valid type for `%s`, expected: `String`"
+                  .formatted(ASTHelpers.getType(nameArgument), enumType))
+          .build();
+    }
```
Stephan202, line 87: If we introduce the two `Matcher`s mentioned above, then we can skip this check.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,185 @@
...
+    linkType = CUSTOM,
+    severity = WARNING,
+    tags = FRAGILE_CODE)
+public final class UnvalidatedEnumValueOfInvocation extends BugChecker
+    implements MethodInvocationTreeMatcher {
+  private static final long serialVersionUID = 1L;
+  private static final Matcher<ExpressionTree> ENUM_VALUE_OF_METHOD =
+      staticMethod().onDescendantOf(Enum.class.getCanonicalName()).named("valueOf");
+  private static final Matcher<ExpressionTree> ENUM_NAME_OR_TO_STRING_METHOD =
+      instanceMethod().onDescendantOf(Enum.class.getCanonicalName()).namedAnyOf("name", "toString");
```
Stephan202: We should test that these methods don't take any argument.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,185 @@
...
+          : buildDescription(tree)
+              .setMessage(
+                  "`%s` might generate values which are missing in `%s`: %s"
+                      .formatted(
+                          SourceCode.treeToString(nameArgument, state), enumType, missingValues))
+              .build();
+    }
+
+    // Match identifiers.
+    return nameArgument instanceof IdentifierTree ? describeMatch(tree) : Description.NO_MATCH;
```
Stephan202, line 129: How confident are we that this doesn't yield significantly more false positives than that it flags bugs? Even in cases where a bug is found, the path towards resolving the warning involves more than fixing the domain of values taken on by the identified variable; avoiding suppression also requires rewriting the code in one of the shapes covered further up. As-is, users are not informed of that. TBH, I suspect we'll just not want to flag this. 🤔

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,185 @@
...
+   *     default -> A.valueOf(b.name());
+   * }
+   * }</pre>
+   */
+  // XXX: Fall-through in colon-style switch statements is not analyzed; only the enclosing case's
+  // labels are considered.
+  private static ImmutableSet<String> findSwitchCoveredValues(
+      ExpressionTree enumValueArgument, ImmutableSet<String> valuesOfReceiver, VisitorState state) {
+    CaseTree enclosingCaseTree = ASTHelpers.findEnclosingNode(state.getPath(), CaseTree.class);
+    /* Fast path: avoids unnecessary switch-tree lookups when not inside a case. */
```
Stephan202: Usually we place this "fast path" comment inside the branch.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,185 @@
...
+      SwitchTree switchStmtTree = ASTHelpers.findEnclosingNode(state.getPath(), SwitchTree.class);
+      /* Defensive: a `CaseTree` is structurally always inside a switch. */
+      if (switchStmtTree == null) {
+        return valuesOfReceiver;
+      }
+      switchExpressionType = ASTHelpers.getType(switchStmtTree.getExpression());
+      switchCases = switchStmtTree.getCases();
+    }
+
+    Type nameInvocationReceiverType = ASTHelpers.getReceiverType(enumValueArgument);
```
Stephan202: This logic assumes that `enumValueArgument` matches `ENUM_NAME_OR_TO_STRING_METHOD`, which is validated outside the scope of this method; that's rather fragile.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
```diff
@@ -0,0 +1,185 @@
...
+    if (!ASTHelpers.isSameType(switchExpressionType, nameInvocationReceiverType, state)) {
+      return valuesOfReceiver;
+    }
```
Stephan202: IIUC, here we shouldn't check whether these guys are the same type, but in fact point to the same symbol. Let's also make sure to add a counter example test case for this.

Stephan202 (commented):
Quick partial review while @rickie and I were checking the code. Cool stuff @ghokun!

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
ghokun: This still captures `null`s. Such as:


```
"    // BUG: Diagnostic matches: INVALID_TYPE",
"    A.valueOf(null);",
"    // BUG: Diagnostic matches: INVALID_TYPE",
"    A.valueOf(A.class, null);",
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
ghokun: Added `.withNoParameters();` to end.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
ghokun: Replaced type check with symbol check. Moved symbol capture outside.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
ghokun: Implemented symbol comparison.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
ghokun: These are currently matching

```
"    A.valueOf(raw);",
"    A.valueOf(A.class, raw);",
"    Enum.valueOf(A.class, raw);",
```

We can leave it up to the consumers that they have validated this raw string outside manually and remove this check completely.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
ghokun: Added explicit matchers for all cases:

```
A.valueOf(name);
A.valueOf(A.class, name);
Enum.valueOf(A.class, name);
```

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocationTest.java
```diff
@@ -0,0 +1,205 @@
...
+            "      case ONE, FOUR -> {}",
+            "      default -> {",
+            "        // BUG: Diagnostic matches: MISSING_VALUE",
+            "        consume(A.valueOf(b.name()));",
+            "        // BUG: Diagnostic matches: MISSING_VALUE",
+            "        consume(Enum.valueOf(A.class, b.name()));",
+            "      }",
+            "    }",
+            "",
+            "    if (d instanceof D(B deconstructedB)) {",
```
ghokun, line 187: There are also potential cases which I ignored on purpose such as:


```java
enum A {ONE};
enum B {ONE, TWO};
record D(B b) {}

D d = new D(B.ONE);
switch (d) {
  case D(B deconstructedB) when B == ONE -> ..
  case D(B deconstructedB) when B == TWO -> ..
};
```

I think this will complicate things even further. Should we place it inside the ignored part?

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocationTest.java
```diff
@@ -0,0 +1,205 @@
...
+            "",
+            "  void ignoredCases(B b) {",
+            "    A.valueOf(getClass().getSimpleName());",
+            "    A.valueOf(A.class, getClass().getSimpleName());",
+            "    Enum.valueOf(A.class, getClass().getSimpleName());",
+            "    A.valueOf(null, getClass().getSimpleName());",
+            "    Enum.valueOf(null, getClass().getSimpleName());",
+            "    List.of(\"\").stream().map(A::valueOf);",
+            "",
+            "    // Colon-style fall-through: `case FOUR` reaches the invocation, but only the",
```
ghokun, line 76: Placed here to be explicit about the ignored case, if you see an easy way to implement it or that it is possible, let me know.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
ghokun: Made matchers explicit.

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocationTest.java
```diff
@@ -0,0 +1,220 @@
...
+            "        // BUG: Diagnostic matches: MISSING_VALUE",
+            "        consume(Enum.valueOf(A.class, b.name()));",
+            "      }",
+            "    }",
+            "    switch (b) {",
+            "      case ONE, TWO, THREE -> {",
+            "      // BUG: Diagnostic matches: MISSING_VALUE",
+            "        consume(A.valueOf(B.FOUR.name()));",
+            "      // BUG: Diagnostic matches: MISSING_VALUE",
+            "        consume(Enum.valueOf(A.class, B.FIVE.name()));",
```
ghokun, line 182: - Switch Cases are valid
- Types match
- Symbols do not match
- Value from the provided enum is a MemberSelectTree which is not present in A.class

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/UnvalidatedEnumValueOfInvocation.java
ghokun: Covered here: https://github.com/PicnicSupermarket/error-prone-support/pull/2031/files#r3266069897

PR #2212, opened 2026-04-13
Record and validate Maven dependency checksums

Stephan202:
Suggested commit message:
```
Record and validate Maven dependency checksums (#2212)

See https://maven.apache.org/resolver/expected-checksums.html
```


.renovaterc.json
```diff
@@ -34,6 +34,16 @@
       "schedule": "every 4 weeks on Monday"
     }
   ],
+  "postUpgradeTasks": {
+    "commands": [
+      "rm .mvn/checksums/checksums-central.sha256",
+      "mvn -B release:prepare release:clean -DdryRun=true -Daether.artifactResolver.postProcessor.trustedChecksums.record=true -DpreparationGoals='clean install' -Darguments='-Daether.artifactResolver.postProcessor.trustedChecksums.record=true -Dverification.warn'"
```
Stephan202: This command aims to trigger downloading of "all" dependencies. Not sure that I in fact I succeeded; we may need two-three iterations on this.

.renovaterc.json
```diff
@@ -34,6 +34,16 @@
...
+    ],
+    "executionMode": "branch",
+    "fileFilters": [
+      ".mvn/checksums/checksums-central.sha256"
+    ]
+  },
```
Stephan202, line 47: @rickie the local Renovate test setup doesn't work for me, for some reason. If it does for you: would you mind testing this setup locally with some Maven upgrade, to validate that the command runs successfully, and that `.mvn/checksums/checksums.sha256` is updated? (If Renovate runs the commands inside a container, perhaps we need to in [install](https://docs.renovatebot.com/configuration-options/#postupgradetasksinstalltools) the right version of Maven and Java; that could be a hassle.)

Stephan202 (commented):
With these changes a corrupt local Maven repository should be easier to identify, and this should go some way towards preventing supply chain attacks.

.renovaterc.json
Stephan202: I now replaced this with a much cleaner approach.

pom.xml
```diff
@@ -571,6 +571,11 @@
                         </execution>
                     </executions>
                 </plugin>
+                <plugin>
+                    <groupId>com.github.ferstl</groupId>
+                    <artifactId>depgraph-maven-plugin</artifactId>
+                    <version>4.0.3</version>
+                </plugin>
```
Stephan202, line 578: Added for the [Automatic Dependency Submission (Maven)](https://github.com/PicnicSupermarket/error-prone-support/actions/runs/24414583486/job/71320211017?pr=2212) job, enabled in the settings:

Image

(The documentation is unclear as to the benefit, though 🤔.)

pom.xml
```diff
@@ -660,6 +665,16 @@
...
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-antrun-plugin</artifactId>
+                    <version>3.1.0</version>
+                </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-assembly-plugin</artifactId>
+                    <version>3.7.1</version>
+                </plugin>
```
Stephan202, line 677: These are defined in the [super pom](https://maven.apache.org/ref/3.9.14/maven-model-builder/super-pom.html). Generally not used, but I noticed that `mvn release:prepare` did try to download them.

pom.xml
```diff
@@ -1198,6 +1213,11 @@
                         </execution>
                     </executions>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-help-plugin</artifactId>
+                    <version>3.5.1</version>
+                </plugin>
```
Stephan202, line 1220: Just making sure that `mvn help:describe` etc. work well.

Stephan202 (commented):
Some context. This is now blocked on testing whether this is compatible with our Renovate setup.

.renovaterc.json
Stephan202: Now even cleaner: a separate `update-maven-checksums.sh` script.

.renovaterc.json
Stephan202: > @rickie the local Renovate test setup doesn't work for me, for some reason.

Turns out this was due to a Redis/Valkey storage incompatibility as a result of an upgrade; `docker compose down --volumes` fixed this.

Anyway, I now tested this setup: it requires setting [`allowedcommands`](https://docs.renovatebot.com/self-hosted-configuration/#allowedcommands) server-side, but even then it doesn't work because `mvn` isn't available. This may be fixable, but then we'd also need to keep the Maven and Java versions there updated. That sounds like too much hassle.

Stephan202:
I found [this interesting thread](https://lists.apache.org/thread/1nyxpdrwcknvgrzqw89sxbno732kcock), which confirms the issues I observed so far as well, such as the [lacking Renovate support](https://github.com/PicnicSupermarket/error-prone-support/pull/2212#discussion_r3075697793) and [issues with ad hoc plugin execution](https://github.com/PicnicSupermarket/error-prone-support/pull/2212#discussion_r3082357865). (I also read that Jitpack would throw up another hurdle, but luckily we don't publish this project via Jitpack.)

Unless we bother to update and maintain a Picnic-internal Renovate image, the only alternative I see is to have a GitHub Actions workflow that updates the hashes after version bumps. But to prevent a bunch of build failures when the PR is first opened, that would require further changes. Right now I can think of a few approaches:

###### Modify Renovate upgrade branches before Renovate opens a PR 
1. Enable the [Renovate Dependency Dashboard](https://docs.renovatebot.com/key-concepts/dashboard/).
2. Set [`"prCreation" = "approval"`](https://docs.renovatebot.com/configuration-options/#prcreation).
3. Introduce a workflow that:
    1. On pushes to `renovate/*` branches runs the update script.
    2. If that yields any change, pushes a commit to the branch.
    3. Updates the dashboard comment to approve PR opening.

###### Modify Renovate upgrade branches before Picnic-Bot opens a PR
1. Provide upstream support for [`"prCreation" = "never"`](https://docs.renovatebot.com/configuration-options/#prcreation), as such a feature [is likely to be accepted](https://github.com/renovatebot/renovate/discussions/15484#discussioncomment-2700220).
2. Once released, enable this feature.
3. Introduce a workflow that:
    1. On pushes to `renovate/*` branches runs the update script.
    2. If that yields any change, pushes a commit to the branch.
    3. If the branch is not yet associated with a PR, opens one.

###### Modify Renovate upgrade branches after Renovate opens a PR
1. [Configure](https://docs.renovatebot.com/configuration-options/#addlabels) an additional `no-validation` label for new Renovate PRs.
5. Update all "relevant" workflows to:
    1. Respect this label by skipping execution if it is present.
    2. Also trigger if the label is removed.
6. Introduce a workflow that:
    1. On pushes to `renovate/*` PRs runs the update script on the associated branch.
    5. If that yields any change, pushes a commit to the branch.
    7. Removes the `no-validation` label from the PR if present.
8. Introduce another workflow that disallows merging if the `no-validation` label is present.

A fourth option would be to extend Renovate's [`postUpdateOptions`](https://docs.renovatebot.com/configuration-options/#postupdateoptions) or similar, but without standardization of the ecosystem, I'm not sure what that implementation would exactly look like.

I think that the third solution is _way_ to complex. I may circle back to the first or second solution somewhere in the coming weeks; TBD.

NB: Thanks @cstamas for the nice write-ups on this topic!

cstamas:
:beers:

PR #2223, opened 2026-04-18
Extend `RefasterSourceCompatibility` to account for class-level type variables

Stephan202:
Suggested commit message:
```
Extend `RefasterSourceCompatibility` to account for class-level type variables (#2223)

The checker now handles before/after template method return and parameter
types that contain class-level type variables. In such cases, it
recursively infers type-variable-to-concrete-type mappings from the
corresponding before-template types, then substitutes those mappings to
determine compatibility.

As a result, the four `{Mono,Flux}Then{,WithMono}` rules in `ReactorRules`
no longer require `@PossibleSourceIncompatibility`: the type variable `S`
can be mapped to `Void`, making the after-template return type compatible
with that of each before-template.
```


Stephan202:
Sorry @rickie and @mohamedsamehsalah; ignore this one. Again Claude Code being over-eager and opening a PR without being asked.

mohamedsamehsalah (commented):
Cleaning up pending review notifications. Please re request review when needed.

PR #2264, opened 2026-05-30
Add Refaster rule to rewrite Optional.of(x).orElse(y) to requireNonNull(x)

### LINKED ISSUE #436: Rewrite `Optional.of(A).orElse(B)` to `requireNonNull(A)`
### Problem

As @EnricSala mentioned in https://github.com/PicnicSupermarket/error-prone-support/issues/364, we might want to investigate rewriting `Optional.of(A).orElse(B)` to `requireNonNull(A)`. This will probably require a `BugChecker` as we probably want to provide two fixes, have to do a more extensive analysis, and are making some non behavior preserving changes. 

We should also consider the `Optional#orElseGet` case. 

- [ ] Improve performance.

```java
Optional.of(1).orElse(2); // Holds for non-nullable things.
Optional.of(null).orElse(2); // Holds for nullable things.
```

to:
```java
requireNonNull(1);  // Holds for non-nullable things.
Optional.ofNullable(null).orElse(2); // Holds for nullable things.
```

Vinu2111:
Closes #436

Adds a new `RequireNonNull` Refaster rule to `OptionalRules.java` that rewrites `Optional.of(value).orElse(fallback)` to `requireNonNull(value)`. Since `Optional.of` throws `NullPointerException` if `value` is null, `fallback` can never be returned it is permanently dead code. This rewrite makes the non-null intent explicit.

Caveats are documented in the Javadoc and via XXX comment:
- The rewrite is not behaviour-preserving if `fallback` has side effects (documented via `Warning` paragraph in Javadoc).
- A follow-up `BugChecker` could handle the broader case where `Optional.of` is used with a possibly-null value, offering two suggested fixes: rewrite to `requireNonNull` or `Optional.ofNullable` (documented via XXX comment).

Files changed:
- `OptionalRules.java` - new `RequireNonNull` rule + import
- `OptionalRulesTestInput.java`-  before pattern
- `OptionalRulesTestOutput.java` - expected after pattern + import

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
@@ -447,10 +447,13 @@ final class OptionalRules {
   }
 
   /**
-   * Prefer {@link java.util.Objects#requireNonNull(Object)} over less explicit alternatives.
+   * Prefer {@link java.util.Objects#requireNonNull(Object)} over an {@link
+   * java.util.Optional#orElse(Object)} call on an {@link java.util.Optional} that is known to be
+   * non-empty.
    *
-   * <p><strong>Warning:</strong> this rewrite is not behaviour-preserving if the fallback
-   * expression has side effects; such side effects will no longer be executed after the rewrite.
+   * <p><strong>Warning:</strong> This rewrite is not behaviour-preserving if the {@code fallback}
+   * expression has side effects: after the rewrite, such side effects will no longer be executed.
+   * In practice such code would be extremely fragile, so this is considered acceptable.
    */
   // XXX: Consider introducing a `BugChecker` that also handles the case where `Optional.of` is
   // used with a possibly-null value, offering two suggested fixes: rewrite to `requireNonNull`
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
@@ -1,6 +1,7 @@
 package tech.picnic.errorprone.refasterrules;
 
 import static com.google.errorprone.refaster.ImportPolicy.STATIC_IMPORT_ALWAYS;
+import static java.util.Objects.requireNonNull;
 
 import com.google.common.collect.Streams;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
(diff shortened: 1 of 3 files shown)


Vinu2111:
Hi @rickie,  as discussed with @Stephan202 on issue #436, I've gone ahead and opened this PR. Happy to iterate on the implementation based on your feedback. Let me know if you'd like any changes to the rule structure, the XXX comments, or the test cases.

rickie:
Hi @Vinu2111 , due to sick leave and holidays we were a bit slow. Thanks for the contribution already!

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -445,6 +446,31 @@ Optional<V> after(Optional<S> optional, Function<W, V> mapper) {
...
 
+  /**
+   * Prefer {@link java.util.Objects#requireNonNull(Object)} over an {@link
+   * java.util.Optional#orElse(Object)} call on an {@link java.util.Optional} that is known to be
+   * non-empty.
+   *
+   * <p><strong>Warning:</strong> This rewrite is not behaviour-preserving if the {@code fallback}
+   * expression has side effects: after the rewrite, such side effects will no longer be executed.
+   * In practice such code would be extremely fragile, so this is considered acceptable.
+   */
```
rickie, line 457:
```suggestion
   /**
     * Prefer {@link java.util.Objects#requireNonNull(Object)} over more contrived alternatives.
   *
   * <p>Since {@link Optional#of(Object)} rejects {@code null}, the {@link Optional#orElse(Object)}
   * fallback is unreachable and merely obscures a plain non-{@code null} assertion.
    *
    * <p><strong>Warning:</strong> This rewrite is not behaviour-preserving if the {@code fallback}
    * expression has side effects: after the rewrite, such side effects will no longer be executed.
    * In practice such code would be extremely fragile, so this is considered acceptable.
    */

```

Not sure if it's my git setup (I recently changed something there) but I'm not able to push a commit 🤔. So I'm adding it as comment :)

rickie (commented):
This is the only suggestion, the PR looks great! 

Again, thanks for contributing @Vinu2111.

I think we'll soon release this one as part of 0.30.0.

Vinu2111:
Thanks @rickie! I've applied the suggested Javadoc update and pushed it. Let me know if anything else needs tweaking, really glad it's looking close to ready for 0.30.0!

Vinu2111:
Hi @rickie  apologies for the disruption here. #2264 ended up stuck in a broken state on GitHub's side (it shows "Merge status cannot be loaded" and the Reopen button stays disabled). I checked with GitHub Support and confirmed this is a known issue with no manual fix available, the documented recovery is to recreate the PR from the same branch.

I've opened a new PR with the exact same commits, including the Javadoc wording you suggested: #2275

Sorry for the extra noise, and thanks again for your patience and review on this one!

rickie:
No worries!

PR #2275, opened 2026-06-18
Extend `RequireNonNullExpression` Refaster rule

### LINKED ISSUE #436: Rewrite `Optional.of(A).orElse(B)` to `requireNonNull(A)`

Vinu2111:
Recreation of #2264, which got stuck in a broken state on GitHub's side ("Merge status cannot be loaded" after an accidental close confirmed via GitHub Support that there's no supported manual recovery for this state).

This PR contains the exact same work, including the Javadoc wording @rickie suggested in review on #2264.

Closes #436

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
@@ -447,9 +447,10 @@ final class OptionalRules {
   }
 
   /**
-   * Prefer {@link java.util.Objects#requireNonNull(Object)} over an {@link
-   * java.util.Optional#orElse(Object)} call on an {@link java.util.Optional} that is known to be
-   * non-empty.
+   * Prefer {@link java.util.Objects#requireNonNull(Object)} over more contrived alternatives.
+   *
+   * <p>Since {@link Optional#of(Object)} rejects {@code null}, the {@link Optional#orElse(Object)}
+   * fallback is unreachable and merely obscures a plain non-{@code null} assertion.
    *
    * <p><strong>Warning:</strong> This rewrite is not behaviour-preserving if the {@code fallback}
    * expression has side effects: after the rewrite, such side effects will no longer be executed.
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
@@ -461,6 +461,7 @@ final class OptionalRules {
   // (non-null case) or rewrite to `Optional.ofNullable` (nullable case).
   static final class RequireNonNull<T> {
     @BeforeTemplate
+    @SuppressWarnings("OptionalOfRedundantMethod" /* This violation will be rewritten. */)
     T before(T value, T fallback) {
       return Optional.of(value).orElse(fallback);
     }
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
@@ -447,10 +447,13 @@ final class OptionalRules {
   }
 
   /**
-   * Prefer {@link java.util.Objects#requireNonNull(Object)} over less explicit alternatives.
+   * Prefer {@link java.util.Objects#requireNonNull(Object)} over an {@link
+   * java.util.Optional#orElse(Object)} call on an {@link java.util.Optional} that is known to be
+   * non-empty.
    *
-   * <p><strong>Warning:</strong> this rewrite is not behaviour-preserving if the fallback
-   * expression has side effects; such side effects will no longer be executed after the rewrite.
+   * <p><strong>Warning:</strong> This rewrite is not behaviour-preserving if the {@code fallback}
+   * expression has side effects: after the rewrite, such side effects will no longer be executed.
+   * In practice such code would be extremely fragile, so this is considered acceptable.
    */
   // XXX: Consider introducing a `BugChecker` that also handles the case where `Optional.of` is
   // used with a possibly-null value, offering two suggested fixes: rewrite to `requireNonNull`
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
@@ -1,6 +1,7 @@
 package tech.picnic.errorprone.refasterrules;
 
 import static com.google.errorprone.refaster.ImportPolicy.STATIC_IMPORT_ALWAYS;
+import static java.util.Objects.requireNonNull;
 
 import com.google.common.collect.Streams;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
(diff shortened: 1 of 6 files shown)


Vinu2111:
Thanks for flagging this! I tested this assumption directly: Refaster does not apply rules to a fixed point each rule fires independently against the original input rather than chaining through other rules' outputs. I confirmed this by running the full RefasterRulesTest suite (78 tests, 0 failures) with the existing testOptionalOrElse expectation unchanged. The current output (Optional.of("bar").orElse("baz")) is correct as-is.

rickie:
Ohhh...

It seems like Error Prone actually *does* have a BugPattern for this:

```
[WARNING] /Users/runner/work/error-prone-support/error-prone-support/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java:[465,39] [OptionalOfRedundantMethod] Optional.of() always returns a non-empty Optional. Using 'orElse' method on it is unnecessary and most probably a bug.
    (see https://errorprone.info/bugpattern/OptionalOfRedundantMethod)
  Did you mean 'return Optional.ofNullable(value).orElse(fallback);' or 'return value;'?
```

I don't have time to dive into it now, but it might be the case that this is flagged already...

Vinu2111:
Good catch, I checked the source: OptionalOfRedundantMethod is a built-in Error Prone check (com.google.errorprone), not part of error-prone-support. It's configured as severity ERROR upstream and does build a real SuggestedFix, but that fix only gets applied via the explicit patch mechanism (-XepPatchChecks/-fix), not during a normal build which is why it just failed/warned here rather than auto-fixing.

There's also a difference in intent: its suggested fix (ofNullable(value).orElse(fallback)) assumes value might be null, while our RequireNonNull rule targets the case where value is provably non-null (Optional.of() already throws on null). So requireNonNull(value) is the more precise fix for that case.

Happy to cross-reference it in the Javadoc if useful, or adjust scope if you'd rather consolidate... your call!

rickie:
Hmm, this is an interesting case!

There are indeed three options I see:
- From `Optional.of` -> `Optional.ofNullable`.
- From Optional.of("foo").orElse("bar")` -> "foo". 
- Simply flagging it.

I'm leaning towards keeping + merging this PR as it solves it in a slightly better way IMO. What we have to do, is we need to add a suppression for this check on the Refaster rule that we are introducing. 
We usually use an `@SuppressWarnings` for it with the text: `/* This is a more specific template. */`. You can check the codebase for an example.

> that fix only gets applied via the explicit patch mechanism (-XepPatchChecks/-fix), not during a normal build which is why it just failed/warned here rather than auto-fixing.

There are two modes, the "normal" flagging mode and the patch mechanism. Both checks (the [OptionalOfRedundantMethod](https://errorprone.info/bugpattern/OptionalOfRedundantMethod) and this Refaster rule) can be configured to simply flag issues or automatically patch code. 

---

I'd say let's apply this suppression, link the Error Prone check in the Javadoc, and see if the build is green!

Curious to see what @Stephan202 thinks!

Vinu2111:
Done applied both changes. One note: I checked for precedent on linking to a built-in Error Prone check from the Javadoc itself and found none in this directory (the convention here is // XXX: comments with backtick-quoted names, e.g. AssertThatIsNull → AssertJNullnessAssertion). Using {@link} would also add a Javadoc dependency on an internal Error Prone class, which I felt risky. So I added the cross-reference as an // XXX: comment instead, matching that pattern,  let me know if you'd prefer something more explicit!

rickie (approved):
Thanks @Vinu2111 for the contribution and the effort!

In case that an `@link` introduces a new dependency or import, we sometimes use `@code` instead :).

rickie:
As discussed briefly with @Stephan202 some minor tweaks. Given some holidays he has/had a busy schedule. But we didn't forget this one :D

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -445,6 +446,39 @@ Optional<V> after(Optional<S> optional, Function<W, V> mapper) {
...
+   *
+   * @implNote This rule has overlap with the existing `OptionalOfRedundantMethod`, but this is more
+   *     behavior-preserving.
+   */
+  // XXX: Consider introducing a `BugChecker` that also handles the case where `Optional.of` is
+  // used with a possibly-null value, offering two suggested fixes: rewrite to `requireNonNull`
+  // (non-null case) or rewrite to `Optional.ofNullable` (nullable case).
+  // XXX: This rule partially overlaps with Error Prone's built-in
+  // `OptionalOfRedundantMethod` check, but provides a more precise rewrite.
+  static final class RequireNonNull<T> {
```
Stephan202: Rules are named after their `@AfterTemplate` code, and should also be located in a rule collection class that relates to the `@AfterTemplate` code. In this case we already have a rule that rewrites to `requireNonNull` in `PreconditionsRules`, so we can simply extend _that_ rule, rather than introducing a new one :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -445,6 +446,39 @@ Optional<V> after(Optional<S> optional, Function<W, V> mapper) {
...
+    @BeforeTemplate
+    @SuppressWarnings(
+        "OptionalOfRedundantMethod" /* This is a template with an alternative rewrite. */)
```
Stephan202: Consistency with other code suggests:
```suggestion
    @SuppressWarnings("OptionalOfRedundantMethod" /* This violation will be rewritten. */)
```

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -445,6 +446,39 @@ Optional<V> after(Optional<S> optional, Function<W, V> mapper) {
...
+   * In practice such code would be extremely fragile, so this is considered acceptable.
+   *
+   * @implNote This rule has overlap with the existing `OptionalOfRedundantMethod`, but this is more
+   *     behavior-preserving.
+   */
+  // XXX: Consider introducing a `BugChecker` that also handles the case where `Optional.of` is
+  // used with a possibly-null value, offering two suggested fixes: rewrite to `requireNonNull`
+  // (non-null case) or rewrite to `Optional.ofNullable` (nullable case).
+  // XXX: This rule partially overlaps with Error Prone's built-in
+  // `OptionalOfRedundantMethod` check, but provides a more precise rewrite.
```
Stephan202: I'll rephrase some of this while merging the rules, but content-wise it's good stuff :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -445,6 +446,32 @@ Optional<V> after(Optional<S> optional, Function<W, V> mapper) {
...
+   * expression has side effects: after the rewrite, such side effects will no longer be executed.
+   * In practice such code would be extremely fragile, so this is considered acceptable.
+   */
+  // XXX: Consider introducing a `BugChecker` that also handles the case where `Optional.of` is
+  // used with a possibly-null value, offering two suggested fixes: rewrite to `requireNonNull`
+  // (non-null case) or rewrite to `Optional.ofNullable` (nullable case).
+  static final class RequireNonNull<T> {
+    @BeforeTemplate
+    T before(T value, T fallback) {
+      return Optional.of(value).orElse(fallback);
```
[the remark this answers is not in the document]
Stephan202: The unit tests perform a single pass; they do not run to a fixed point. Instead, we expect users to apply rules to a fixed point, such that rules do not need to be redundant/consider overlap.

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/OptionalRules.java
```diff
@@ -445,6 +446,39 @@ Optional<V> after(Optional<S> optional, Function<W, V> mapper) {
...
+   * <p><strong>Warning:</strong> This rewrite is not behavior-preserving if the {@code fallback}
+   * expression has side effects: after the rewrite, such side effects will no longer be executed.
+   * In practice such code would be extremely fragile, so this is considered acceptable.
+   *
+   * @implNote This rule has overlap with the existing `OptionalOfRedundantMethod`, but this is more
+   *     behavior-preserving.
+   */
+  // XXX: Consider introducing a `BugChecker` that also handles the case where `Optional.of` is
+  // used with a possibly-null value, offering two suggested fixes: rewrite to `requireNonNull`
+  // (non-null case) or rewrite to `Optional.ofNullable` (nullable case).
```
Stephan202: This is kind-of what `OptionalOfRedundantMethod` already does. We could have it also suggest `requireNonNull`, but there'd still be the issue that outside Google, there's no good support for selecting any but the first suggested fix. I propose to omit this comment.

Stephan202 (approved):
Forgot to press approve :)

rickie (approved):

rickie:
Thanks for the effort and patience @Vinu2111 !!

PR #2280, opened 2026-07-15
Support custom `StaticImport` candidates

rickie:
Suggested commit message:
```
Support custom `StaticImport` candidates (#2280)

The candidate types and members recognized by the `StaticImport` check
are hard-coded, so that project-specific APIs cannot benefit from
static import enforcement. Introduce the `StaticImport:CandidateTypes`
and `StaticImport:CandidateMembers` flags, using which additional
candidates can be specified. The `NonStaticImport` check honors the
latter flag, such that it does not undo static imports introduced by
its sibling check.
```

The `StaticImport` and `NonStaticImport` checks decide which type
members belong behind a static import from hard-coded lists, so
project-specific APIs (e.g. a Picnic-internal `AttributeId#parseAttributeId`)
can never benefit from the enforcement. This PR makes those lists
extensible through two Error Prone flags:

- `-XepOpt:StaticImport:CandidateTypes=<type>,...` — additional types
  all of whose members should be statically imported.
- `-XepOpt:StaticImport:CandidateMembers=<type>#<member>,...` —
  additional individual members.

Both flags are purely additive to the built-in lists. `NonStaticImport`
reads the members flag through the same parser, so it never strips a
static import that `StaticImport` was configured to introduce (avoiding
a fix that oscillates between the two checks).

Configuration is validated at construction time: malformed entries,
whitespace, and values that clash with `NonStaticImport`'s exclusion
lists fail fast with a descriptive `IllegalArgumentException` rather
than being silently ignored.

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
@@ -258,6 +258,17 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
         .orElse(Description.NO_MATCH);
   }
 
+  private Optional<String> getCandidateSimpleName(StaticImportInfo importInfo) {
+    String canonicalName = importInfo.canonicalName();
+    return importInfo
+        .simpleName()
+        .toJavaUtil()
+        .filter(
+            name ->
+                candidateTypes.contains(canonicalName)
+                    || candidateMembers.containsEntry(canonicalName, name));
+  }
+
   private static boolean isCandidateContext(VisitorState state) {
     Tree parentTree =
         requireNonNull(state.getPath().getParentPath(), "MemberSelectTree lacks enclosing node")
@@ -277,17 +288,6 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
     Type type = ASTHelpers.getType(tree.getExpression());
     return type != null
         && !NON_STATIC_IMPORT_CANDIDATE_MEMBERS.containsEntry(type.toString(), identifier);
-  }
-
-  private Optional<String> getCandidateSimpleName(StaticImportInfo importInfo) {
-    String canonicalName = importInfo.canonicalName();
-    return importInfo
-        .simpleName()
-        .toJavaUtil()
-        .filter(
-            name ->
-                candidateTypes.contains(canonicalName)
-                    || candidateMembers.containsEntry(canonicalName, name));
   }
 
   private static Optional<Fix> tryStaticImport(
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
@@ -1,18 +1,22 @@
 package tech.picnic.errorprone.bugpatterns;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.errorprone.BugPattern.LinkType.CUSTOM;
 import static com.google.errorprone.BugPattern.SeverityLevel.SUGGESTION;
 import static com.google.errorprone.BugPattern.StandardTags.STYLE;
 import static java.util.Objects.requireNonNull;
 import static tech.picnic.errorprone.bugpatterns.NonStaticImport.NON_STATIC_IMPORT_CANDIDATE_IDENTIFIERS;
 import static tech.picnic.errorprone.bugpatterns.NonStaticImport.NON_STATIC_IMPORT_CANDIDATE_MEMBERS;
+import static tech.picnic.errorprone.bugpatterns.NonStaticImport.NON_STATIC_IMPORT_CANDIDATE_TYPES;
 import static tech.picnic.errorprone.utils.Documentation.BUG_PATTERNS_BASE_URL;
 
 import com.google.auto.service.AutoService;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.CharMatcher;
 import com.google.common.base.Functions;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicates;
+import com.google.common.base.Splitter;
 import com.google.common.base.Verify;
 import com.google.common.collect.Comparators;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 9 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
```diff
@@ -190,8 +209,37 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
...
+   * those defined by {@link #STATIC_IMPORT_CANDIDATE_TYPES}) can be specified.
+   */
+  private static final String CANDIDATE_TYPES_FLAG = "StaticImport:CandidateTypes";
+
+  /**
+   * Flag using which additional type members that should be statically imported (in addition to
+   * those defined by {@link #STATIC_IMPORT_CANDIDATE_MEMBERS}) can be specified, in {@code
+   * <type>#<member>} format.
+   */
+  private static final String CANDIDATE_MEMBERS_FLAG = "StaticImport:CandidateMembers";
```
Stephan202: Instead of the word `Candidate`, perhaps `StaticImport:AdditionalTypes` and `StaticImport:AdditionalMembers` makes more sense to users. Other code references, as well as field names, would have to be renamed.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
```diff
@@ -190,8 +209,37 @@ public final class StaticImport extends BugChecker implements MemberSelectTreeMa
...
+
+  /**
+   * Instantiates a customized {@link StaticImport}.
+   *
+   * @param flags Any provided command line flags.
+   */
+  @Inject
+  StaticImport(ErrorProneFlags flags) {
+    candidateTypes = getCandidateTypes(flags);
+    candidateMembers = getCandidateMembers(flags);
```
Stephan202: `getCandidateMembers` is also called from `NonStaticImport`. As these classes are so linked already, maybe it makes sense to extract a `StaticImportConfig` type that encapsulates all relevant logic. It would have an `@Inject`-annotated constructor with a `ErrorProneFlags` parameter, while the two checks would have a `StaticImportConfig` instance injected, rather than a `ErrorProneFlags` instance as they do now.

For an Error Prone example, see e.g. how [`WellKnownKeep`](https://github.com/google/error-prone/blob/0187d08f98ae23232fa863e76ddf2251acd9e374/core/src/main/java/com/google/errorprone/bugpatterns/WellKnownKeep.java) is defined and used.

Naturally, we'd then also move relevant tests to a `StaticImportConfigTest` class.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
```diff
@@ -253,4 +301,62 @@ private static Optional<Fix> tryStaticImport(
...
+          candidate);
+      members.put(type, member);
+    }
+
+    return members.build();
+  }
+
+  private static boolean isWellFormed(String typeOrMember) {
+    return !typeOrMember.isEmpty() && CharMatcher.whitespace().matchesNoneOf(typeOrMember);
+  }
```
Stephan202: There's overlap between this code and the parsing that happens in `AnnotationAttributeMatcher`. Maybe we unify part of this logic.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
Stephan202: Additionally, perhaps one flag suffices: types can't contain a `#`, so having a single list doesn't introduce ambiguity. See also the flag accepted by `LexicographicalAnnotationAttributeListing`.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
```diff
@@ -231,15 +279,15 @@ private static boolean isCandidate(MemberSelectTree tree) {
         && !NON_STATIC_IMPORT_CANDIDATE_MEMBERS.containsEntry(type.toString(), identifier);
   }
 
-  private static Optional<String> getCandidateSimpleName(StaticImportInfo importInfo) {
+  private Optional<String> getCandidateSimpleName(StaticImportInfo importInfo) {
```
Stephan202: As this method is now non-static, we should move it up a bit.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
```diff
@@ -231,15 +279,15 @@ private static boolean isCandidate(MemberSelectTree tree) {
...
     String canonicalName = importInfo.canonicalName();
     return importInfo
         .simpleName()
         .toJavaUtil()
         .filter(
             name ->
-                STATIC_IMPORT_CANDIDATE_TYPES.contains(canonicalName)
-                    || STATIC_IMPORT_CANDIDATE_MEMBERS.containsEntry(canonicalName, name));
+                candidateTypes.contains(canonicalName)
+                    || candidateMembers.containsEntry(canonicalName, name));
```
Stephan202: Related to the observation about `AnnotationAttributeMatcher` below: if that class were generalized to some `TypeAndMemberMatcher` (name TBD), perhaps it could also be used for this logic.

Stephan202 (commented):
Cool stuff! This looks clean, but I wonder whether there's an opportunity to clean up some more code. Might be easiest to discuss this offline. (Or point an AI at it; didn't try that yet.)

Didn't scrutinize the tests too much yet.

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/StaticImport.java
Stephan202: (Pushed.)

Stephan202:
> (Or point an AI at it; didn't try that yet.)

I asked Claude to have a crack at a subset of the changes. I pushed the result, but _did not yet review_. That'll come later.

mohamedsamehsalah (commented):
Cleaning up pending review notifications. Please re request review when needed.

PR #2287, opened 2026-07-29
Document Refaster rule collection placement convention

rickie:
Suggested commit message:
```
Document Refaster rule collection placement convention (#2287)
```

Follow-up to a review remark on #2275: rules should live in the collection that relates to their `@AfterTemplate` target (the preferred alternative), not the `@BeforeTemplate` pattern. The naming convention already documents that a rule's *name* derives from its `@AfterTemplate`; this extends the same anchor to *collection placement* and adds a pointer to prefer extending an existing rule over introducing a duplicate.

See: https://github.com/PicnicSupermarket/error-prone-support/pull/2275#discussion_r3610694709

🤖

### DIFF
--- a/.agents/skills/refaster-rule/SKILL.md
+++ b/.agents/skills/refaster-rule/SKILL.md
@@ -13,8 +13,10 @@ Use this skill for requests to add or change files in `*/refasterrules/*`.
 Read [`.github/instructions/refaster-rules.instructions.md`][refaster-rules]
 for the full conventions. Then:
 
-1. **Determine the target collection** based on the topic: add to an existing
-   `{Topic}Rules.java` or create a new one.
+1. **Determine the target collection** from the rule's `@AfterTemplate` target
+   (the preferred alternative), not its `@BeforeTemplate` pattern: add to the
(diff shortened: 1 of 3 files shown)


error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/CollectionRules.java
```diff
@@ -675,6 +677,51 @@ S after(SortedSet<T> sortedSet) {
     }
   }
 
+  /** Prefer {@link Collections#sort(List)} over more verbose alternatives. */
+  static final class CollectionsSort<T extends Comparable<? super T>> {
```
mohamedsamehsalah, line 681: Is including this intended?

rickie: Good point, yes, I asked Claude to scan the existing collections to see if there were violations and move them if so. 
This is a result of that :). So it's applying the rule as part of the PR.

mohamedsamehsalah (approved):

PR #2291, opened 2026-08-12
Suggest JDK-native unmodifiable collectors in CollectorMutability

### LINKED ISSUE #954: CollectorMutability: Suggest immutable Guava alternative
### Problem

In https://github.com/checkstyle/checkstyle/pull/14229 we are enabling the `CollectorMutability` check in the Checkstyle repository. The Checkstyle team tries to minimize the dependencies they have on Guava. Therefore, the rewrite to `Immutable{List,Map,Set}.toImmutable{List,Map,Set}()` is not desirable. An alternative would be to use the `Collectors.toUnmodifiable{List,Map,Set}()`. This is a good alternative as this also clearly communicates that a list is used that should _not_ be modified.

- [ ] Improve performance.

```java
.collect(Collectors.toSet());
```

to:
```java
.collect(Collectors.toUnmodifiableSet());
```

### Considerations

Two things:
* As @Stephan202 notes, this should take into account that this rewrite is only valid when JDK > 9.
* [This XXX](https://github.com/PicnicSupermarket/error-prone-support/blob/master/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/util/ThirdPartyLibrary.java#L21-L22) in `ThirdPartyLibrary` should be resolved as well.

### Participation

- [x] I am willing to submit a pull request to implement this improvement.

Vinu2111:
Fixes #954

`CollectorMutability` previously returned `Description.NO_MATCH` entirely when Guava was absent from the classpath, which meant even the Guava-independent mutable fallback fix (`toCollection(...)`) was suppressed. This restructures the checker to always build a `Description`, following the same pattern already used in `FluxImplicitBlock`.

On top of that fix, when the compilation target is JDK 10+, the checker now offers `Collectors.toUnmodifiable{List,Set,Map}()` as an additional suggested fix, a good middle ground for projects (like Checkstyle) that want to avoid a Guava dependency while still clearly communicating (im)mutability.

Fix ordering, when all are applicable:
1. Guava immutable (`Immutable{List,Set,Map}.toImmutable...`)
2. JDK-native unmodifiable (`Collectors.toUnmodifiable...`), JDK 10+ only
3. Explicit mutable fallback (`toCollection(...)` / `HashMap::new`)

The `toMap` case required separate handling from `toList`/`toSet`, since `toUnmodifiableMap` only has 2-arg and 3-arg overloads matching `toMap`'s existing 2-arg/3-arg cases the 4-arg case was already out of scope (returns `NO_MATCH`) before this change.

Tests updated/added:
- `identificationWithoutGuavaOnClasspath` now expects a diagnostic (previously expected none)
- New `identificationWithoutGuavaAndOlderJdkOnClasspath` covers the pre-JDK-10 case, where neither Guava nor the unmodifiable fix should be offered
- `replacementSecondSuggestedFix` now covers the JDK-unmodifiable fix; the previous mutable-fix assertions moved to a new `replacementThirdSuggestedFix`, matching the updated fix ordering

### DIFF
--- a/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CollectorMutability.java
+++ b/error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CollectorMutability.java
@@ -20,6 +20,7 @@ import com.google.errorprone.matchers.Description;
 import com.google.errorprone.matchers.Matcher;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
+import com.sun.tools.javac.code.Source;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -58,8 +59,7 @@ public final class CollectorMutability extends BugChecker implements MethodInvoc
 
   @Override
   public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
-    if (!ThirdPartyLibrary.GUAVA.isIntroductionAllowed(state)
-        || !COLLECTOR_METHOD.matches(tree, state)) {
+    if (!COLLECTOR_METHOD.matches(tree, state)) {
       return Description.NO_MATCH;
     }
 
@@ -67,6 +67,7 @@ public final class CollectorMutability extends BugChecker implements MethodInvoc
       return suggestToCollectionAlternatives(
           tree,
           ImmutableList.class.getCanonicalName() + ".toImmutableList",
+          Collectors.class.getCanonicalName() + ".toUnmodifiableList",
           ArrayList.class.getCanonicalName(),
           state);
     }
@@ -79,6 +80,7 @@ public final class CollectorMutability extends BugChecker implements MethodInvoc
       return suggestToCollectionAlternatives(
           tree,
           ImmutableSet.class.getCanonicalName() + ".toImmutableSet",
+          Collectors.class.getCanonicalName() + ".toUnmodifiableSet",
           HashSet.class.getCanonicalName(),
           state);
     }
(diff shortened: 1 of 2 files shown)


error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CollectorMutabilityTest.java
```diff
@@ -63,13 +64,62 @@ void identification() {
...
+            "    Stream.empty().collect(Collectors.toMap(o -> o, o -> o));",
+            "    // BUG: Diagnostic matches: Z",
+            "    Stream.empty().collect(Collectors.toSet());",
+            "  }",
+            "}")
+        .doTest();
+  }
+
+  @Test
+  void identificationWithoutGuavaAtExactlyJdk10OnClasspath() {
```
rickie:
```suggestion
  void identificationWithoutGuavaAtExactlyJdk10() {
```

Vinu2111:
Hi @rickie, just checking in on this PR whenever you get a chance to look at it. No rush at all, let me know if anything needs changes on my side. Thanks!

rickie:
Hey, thanks for the ping! I have some changes locally that I will push soon 😄

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CollectorMutabilityTest.java
```diff
@@ -63,13 +64,62 @@ void identification() {
...
+    CompilationTestHelper.newInstance(CollectorMutability.class, getClass())
+        .withClasspath()
+        .setArgs("--release", "10")
+        .expectErrorMessage("X", m -> m.contains("toUnmodifiableList"))
```
rickie: Let's also add the other two :)

error-prone-contrib/src/main/java/tech/picnic/errorprone/bugpatterns/CollectorMutability.java
```diff
@@ -112,22 +123,37 @@ private Description suggestToMapAlternatives(MethodInvocationTree tree, VisitorS
...
-                tree, ImmutableMap.class.getCanonicalName() + ".toImmutableMap", state))
-        .addFix(
-            mutableFix
-                .postfixWith(
-                    tree.getArguments().get(argCount - 1),
-                    (argCount == 2 ? ", (a, b) -> { throw new IllegalStateException(); }" : "")
-                        + ", %s::new".formatted(hashMap))
-                .build())
-        .build();
+  private static boolean isJdk10Plus(VisitorState state) {
```
rickie: We can likely make the implementation a bit simpler with something from Error Prone itself!

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CollectorMutabilityTest.java
```diff
@@ -63,13 +64,62 @@ void identification() {
   void identificationWithoutGuavaOnClasspath() {
     CompilationTestHelper.newInstance(CollectorMutability.class, getClass())
         .withClasspath()
+        .expectErrorMessage("X", m -> !m.contains("toImmutableList"))
```
rickie: Let's make them slightly more specific to be 100% sure :)

rickie (commented):
Sorry, I partially reviewed this from a plane, so I couldn't put a comment on some of the things. Added a commit with an initial review. Please take a look at the changes and let me know what you think :) 

It looks great and would say we are close to an approval. Will soon do the second pass :).

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CollectorMutabilityTest.java
Vinu2111: Done, thanks!

error-prone-contrib/src/test/java/tech/picnic/errorprone/bugpatterns/CollectorMutabilityTest.java
Vinu2111: Agreed, thanks for the fix!

Vinu2111:
Thanks for the review and the suggestions commit, @rickie! I've gone through all the changes you pushed, the supportsUnmodifiableCollectors rename, the stricter bidirectional matchers, and the expanded test coverage. Everything looks great.

I've addressed the remaining item: expanded identificationWithoutGuavaOnJdk10 to cover all three collector types (list, map, set) with the same stricter matchers used in the other tests.

Let me know if anything else needs adjusting on the second pass!

rickie (approved):
Hey @Vinu2111 , thanks again for the contribution 🚀 !

Changes look great and very nicely tested ✨

Vinu2111:
Thanks so much, @rickie really appreciate you taking the time to review this so thoroughly, on a plane and all 😄 Enjoyed working through this one, and the polish commits helped tighten it up nicely. Looking forward to seeing it land in 0.31.0!

PR #2335, opened 2026-09-12
Add Agent Skill Validation and Improvements

AmbitiousAnil:
**Summary:**

- Repair stale AGENT.md references to use canonical AGENTS.md.
- Remove cyclic delegation between review-changes and finalize-changes.
- Add local and CI validation for skill metadata, naming, links, scripts, size, triggers, and invocation cycles.

### **Personal note:**
I’ve enjoyed working with this codebase and would be very interested in opportunities to join Picnic’s engineering team. If there are relevant openings ,point me in the right direction to apply and join.

### **About Me**
for more info about my work : https://anil-portfolio-1.netlify.app/


validate-agent-skills.sh
rickie, line 1: Although the idea of automation is generally something we prefer, I'm not sure about this case. This bash script goes a bit far and perhaps this PR without it is already a great improvement. Would it be okay to drop this file (and the validate-agent-skills.yml) and keep the other changes?

.agents/skills/review-changes/SKILL.md
```diff
@@ -25,8 +25,5 @@ checklist. Then:
    naming, test coverage, error handling). Continue passes until diminishing
    returns.
 5. **Commit** any improvements.
-6. **Finalize**: Invoke [`/finalize-changes`][skill-finalize-changes] to run
```
rickie, line 28: For this recursion, in practice this is not really a problem. Also it helps ensure the changes we made and commit are properly tested also :). It seems like it cannot hurt a lot to keep it

rickie (commented):
Thanks for your contribution 🚀 ! 

Happy to hear you like working with our codebase. There is jobs.picnic.app in case you are interested in the current jobs we have!

PR #2350, opened 2026-09-16
Drop references to `BugCheckerRefactoringTestHelper.TestMode`

rickie:
Suggested commit message:
```
Use `doTest()` rather than `doTest(TestMode.TEXT_MATCH)` (#2350)

`BugCheckerRefactoringTestHelper#doTest()` already performs the text
comparison that the explicit `TestMode` argument selects.

While there, drop `BugCheckerRefactoringTestHelper.TestMode` from
`NonStaticImport`'s candidate types: the enum can only be passed to the
deprecated `doTest(TestMode)` overload, and no test exercised the entry.
```


Stephan202 (approved):
```
Drop references to `BugCheckerRefactoringTestHelper.TestMode` (#2350)

`BugCheckerRefactoringTestHelper#toTest()` defaults to
`TestMode.TEXT_MATCH`, so there's no need to reference this mode
explicitly.
```

mohamedsamehsalah (approved):

PR #2351, opened 2026-09-16
Introduce `BugCheckerRefactoringTestHelperDoTest` Refaster rule

rickie:
Suggested commit message:
```
Introduce `BugCheckerRefactoringTestHelperDoTest` Refaster rule (#2351)
```

Complements #2350, which updates the `BugChecker` test conventions accordingly.


rickie:
Hmm, I think we don't need to merge this TBH. It's a bit overkill 😬

error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BugCheckerRules.java
```diff
@@ -59,6 +60,20 @@ BugCheckerRefactoringTestHelper after(
     }
   }
 
+  /** Prefer {@link BugCheckerRefactoringTestHelper#doTest()} over deprecated alternatives. */
+  static final class BugCheckerRefactoringTestHelperDoTest {
+    @BeforeTemplate
+    @SuppressWarnings("deprecation" /* This deprecated API usage will be rewritten. */)
```
Stephan202, line 66: > Hmm, I think we don't need to merge this TBH. It's a bit overkill 😬

Indeed, we generally don't include rules for deprecated APIs; the build anyway won't pass if they're used. Even nicer would be if this method would have an `@InlineMe` annotation, but that's likely also not worth the upstream PR.

Stephan202 (commented):
Propose closing indeed :)

PR #2359, opened 2026-09-17
Don't persist Git auth token on checkout

Stephan202:
Suggested commit message:
```
Don't persist Git auth token on checkout (#2359)
```


.github/workflows/integration-tests.yml
```diff
@@ -39,6 +39,7 @@ jobs:
         with:
           ref: ${{ env.CHECKOUT_REF }}
           sparse-checkout: integration-tests
+          persist-credentials: false
```
mohamedsamehsalah, line 42: Why is this not the default 😅

mohamedsamehsalah (approved):

.github/workflows/integration-tests.yml
Stephan202: Great question; maybe something they'll change in the next major release :shrug:

Stephan202:
I checked: for the `s4u/setup-maven-action` we also use, [`false` is the default](https://github.com/s4u/setup-maven-action/blob/94f595f3319db94afb3c3c3f6d48d040836f85f5/action.yml#L28-L31).

rickie (approved):

PR #2364, opened 2026-09-19
Scratch: queue:max empirical test (do not merge)

Stephan202:
Temporary PR to observe GitHub Actions behavior for `queue: max` combined with a conditional `cancel-in-progress` expression. Will be closed and the branch deleted once verified.


Stephan202:
Empirical test complete: confirmed that `queue: max` combined with a conditional `cancel-in-progress` fails validation when that condition evaluates true, and that making `queue` itself conditional resolves it. Applying the fix in #2363.

PR #2369, opened 2026-09-19
Name every GitHub Actions job

Stephan202:
Suggested commit message:
```
Name every GitHub Actions job (#2369)

This ensures that every job is unambiguously identified in the GitHub
Actions UI, even when workflows reuse the same job ID.

While there, also improve one job ID and two workflow IDs.
```


.github/workflows/build.yml
```diff
@@ -16,6 +16,7 @@ env:
   MAVEN_ARGS: --batch-mode -Dstyle.color=always -T1C
 jobs:
   build:
```
Stephan202, line 18: Here and in `deploy-website.yml`: we had two jobs named `build`, and GitHub is configured to require a job with that name to pass. Now GitHub is therefore waiting for another check to be reported. We'll drop/update that config before merging, but it also shows that things were ambiguous before this change.

rickie (approved):

mohamedsamehsalah (approved):

PR #2373, opened 2026-09-20
Avoid expression interpolation in GitHub Actions `run` arguments

Stephan202:
Suggested commit message:
```
Avoid expression interpolation in GitHub Actions `run` arguments (#2373)

While there:
- Consistently list `env:` before `run:`.
- Better document a PowerShell gotcha.
```


mohamedsamehsalah (approved):

.github/workflows/pitest-analyze-pr.yml
```diff
@@ -43,7 +43,9 @@ jobs:
         # details.
         run: mvn test pitest:mutationCoverage -DargLine.xmx=2048m -Dverification.skip -Dfeatures="+GIT(from[HEAD~1]), +gitci"
       - name: Aggregate Pitest reports
-        run: mvn pitest-git:aggregate -DkilledEmoji=":tada:" -DmutantEmoji=":zombie:" -DtrailingText="Mutation testing report by [Pitest](https://pitest.org/). Review any surviving mutants by inspecting the line comments under [_Files changed_](${{ github.event.pull_request.html_url }}/files)."
+        env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
```
rickie, line 47: So extracting it like this is the prevention?

rickie (approved):

.github/workflows/pitest-analyze-pr.yml
Stephan202: In this case `github.event.pull_request.html_url` isn't even really attacker controlled, so the benefit is effectively nil, but: yes!

The difference is that `${{ expression }}` is replaced with its value during construction of the string that is given to `run` (i.e., GitHub interpolates the value before invoking the step), so if it contains anything malicious, it can influence what `bash` executes. That's not the case for `${ENVIRONMENT_VARIABLE}`: that's just a string that is passed as-is to `bash`, and it's up to `bash` to resolve its value by interpreting it as a reference to an environment variable. And since that value is then embedded as part of a string expression, its contents will simply become part of that string: thus no security concern.

